PR #1, opened 2015-10-12
Fixup some modules initialization

derkling:
These are some fixes to make more robust the initialization of some modules.


devlib/module/cpufreq.py
```diff
@@ -29,7 +29,20 @@ class CpufreqModule(Module):
 
     @staticmethod
     def probe(target):
+
+        # x86 with Intel P-State driver
+        if target.abi == 'x86_64':
```
setrofim: if target.abi in ['x86', 'x86_64']:

?

PR #2, opened 2015-10-12
CPUFreq: add some methods

derkling:
These are a couple of patches which I've found useful especially on x86 machine with many cores.
They idea is to speed up the selection of multiple CPUfreq governors with just one single target-side command.

### DIFF
--- a/devlib/module/cpufreq.py
+++ b/devlib/module/cpufreq.py
@@ -326,3 +326,26 @@ class CpufreqModule(Module):
             self.target.write_value(sysfile, value)
         except ValueError:
             raise ValueError('Frequency must be an integer; got: "{}"'.format(frequency))
+
+    def set_governor_for_cpus(self, cpus, governor, **kwargs):
+        """
+        Set the governor for the specified list of CPUs.
+        See https://www.kernel.org/doc/Documentation/cpu-freq/governors.txt
+
+        :param cpus: The list of CPU for which the governor is to be set.
+        """
+        online_cpus = self.target.list_online_cpus()
+        for cpu in online_cpus:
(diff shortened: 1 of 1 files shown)


devlib/module/cpufreq.py
```diff
@@ -313,3 +313,44 @@ def set_max_frequency(self, cpu, frequency, exact=True):
...
+        Set the governor for the specified list of CPUs.
+        See https://www.kernel.org/doc/Documentation/cpu-freq/governors.txt
+
+        :param cpus: The list of CPU for which the governor is to be set.
+        """
+        online_cpus = self.target.list_online_cpus()
+        for cpu in cpus:
+            if cpu not in online_cpus:
+                continue
+            self.set_governor(cpu, governor, kwargs)
```
setrofim: why not just

    for cpu in online_cpus:
        self.set_governor(cpu, governor, kwargs)
?

devlib/module/cpufreq.py
```diff
@@ -313,3 +313,44 @@ def set_max_frequency(self, cpu, frequency, exact=True):
...
+        Set the frequency for the specified list of CPUs.
+        See https://www.kernel.org/doc/Documentation/cpu-freq/governors.txt
+
+        :param cpus: The list of CPU for which the frequency has to be set.
+        """
+        online_cpus = self.target.list_online_cpus()
+        for cpu in cpus:
+            if cpu not in online_cpus:
+                continue
+            self.set_frequency(cpu, freq, exact)
```
setrofim: why not just

    for cpu in online_cpus:
        self.set_governor(cpu, governor, kwargs)

?

PR #3, opened 2015-11-10
Fixes and additions for master

derkling:
Here is a couple of fixups for the SSH connection problems, due to command being echoed, and a patch which refactors the CGroups module.
This last patch is not incremental, since it changes drastically the interface.
It's commit message should describe quite clearly how to use the new interface.
Validation has been done using an ipython notebook which is part of the (soon to be released) scheduler testing suite.

### DIFF
--- a/devlib/module/cgroups.py
+++ b/devlib/module/cgroups.py
@@ -22,9 +22,7 @@ from devlib.utils.misc import list_to_ranges, isiterable
 from devlib.utils.types import boolean
 
 
-class CgroupController(object):
-
-    kind = 'cpuset'
+class Controller(object):
 
     def __new__(cls, arg):
         if isinstance(arg, cls):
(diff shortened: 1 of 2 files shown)


devlib/module/cgroups.py
```diff
@@ -32,157 +30,230 @@ def __new__(cls, arg):
...
+        if not self.target:
+            raise RuntimeError('CGroup creation failed: {} controller not mounted'\
+                    .format(self.kind))
+        if name not in self._cgroups:
+            self._cgroups[name] = CGroup(self, name, create=False)
+        return self._cgroups[name].existe()
+
+    def list_all(self):
+        self.logger.debug('Listing groups for %s controller', self.kind)
+        output = self.target.execute('find {} -type d'.format(self.mount_point))
```
setrofim: `find` should be invoked via `self.busybox` as it will not be present on some devices (most Android systems)

devlib/module/cgroups.py
```diff
@@ -32,157 +30,230 @@ def __new__(cls, arg):
...
-        task_ids = self.target.read_value(self.tasks_file).split()
-        return map(int, task_ids)
-
-    def add_tasks(self, tasks):
-        for tid in tasks:
-            self.add_task(tid)
+        # Control cgroup path
+        self.directory = controller.mount_point
+        if name != '/':
+            self.directory = self.directory + name
```
setrofim: self.directory = self.target.path.join(self.directory, name)

devlib/module/cgroups.py
```diff
@@ -32,157 +30,230 @@ def __new__(cls, arg):
...
-        super(CpusetController, self).__init__(*args, **kwargs)
-        self.groups = {}
+    def get(self):
+        conf = {}
+
+        logging.debug('Reading %s attributes from:',
+                self.controller.kind)
+        logging.debug('  %s',
+                self.directory)
+        output = self.target.execute('grep \'\' {}/{}.*'.format(
```
setrofim: `grep` should be invoked via `self.target.busybox` as it will not present on some devices (most Android systems).

devlib/module/cgroups.py
derkling: You mean something like:
self.target.execute('{} find {} -type d'.format(self.target.busybox, self.mount_point))?
... going to update.

devlib/module/cgroups.py
derkling: Right, going to update!

PR #5, opened 2015-11-23
module: Add thermal support

punitagrawal:
Add a thermal module that will allow querying for thermal zones and trip
points in the target. The module allow supports enable/disabling thermal
zones as well as change trip temperatures.

Signed-off-by: Punit Agrawal <punit.agrawal@arm.com>

### DIFF
--- a/devlib/module/thermal.py
+++ b/devlib/module/thermal.py
@@ -1,0 +1,104 @@
+#    Copyright 2015 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+
+from devlib.module import Module
+
+class TripPoint(object):
+    def __init__(self, zone, _id):
+        self._id = _id
+        self.zone = zone
+        self.temp_node = 'trip_point_' + _id + '_temp'
+        self.type_node = 'trip_point_' + _id + '_type'
+
+    @property
+    def target(self):
+        return self.zone.target
+
+    def get_temperature(self):
+        """Returns the currently configured temperature of the trip point"""
+        temp_file = self.target.path.join(self.zone.path, self.temp_node)
+        return self.target.read_int(temp_file)
+
+    def set_temperature(self, temperature):
+        temp_file = self.target.path.join(self.zone.path, self.temp_node)
+        self.target.write_value(temp_file, temperature)
+
+    def get_type(self):
+        """Returns the type of trip point"""
+        type_file = self.target.path.join(self.zone.path, self.type_node)
+        return self.target.read_value(type_file)
+
+class ThermalZone(object):
+    def __init__(self, target, root, _id):
+        self.target = target
+        self.name = 'thermal_zone' + _id
+        self.path = target.path.join(root, self.name)
+        self.trip_points = {}
+
+        for entry in self.target.list_directory(self.path):
+            re_match = re.match('^trip_point_([0-9]+)_temp', entry)
+            if re_match is not None:
+                self.add_trip_point(re_match.group(1))
+
+    def add_trip_point(self, _id):
+        self.trip_points[int(_id)] = TripPoint(self, _id)
+
+    def is_enabled(self):
+        """Returns a boolean representing the 'mode' of the thermal zone"""
+        value = self.target.read_value(self.target.path.join(self.path, 'mode'))
+        return value == 'enabled'
+
+    def set_mode(self, enable):
+        value = 'enabled' if enable else 'disabled'
+        self.target.write_value(self.target.path.join(self.path, 'mode'), value)
+
+    def get_temperature(self):
+        """Returns the temperature of the thermal zone"""
+        temp_file = self.target.path.join(self.path, 'temp')
+        return self.target.read_int(temp_file)
+
+class ThermalModule(Module):
+    name = 'thermal'
+    thermal_root = '/sys/class/thermal'
+
+    @staticmethod
+    def probe(target):
+
+        if target.file_exists(ThermalModule.thermal_root):
+            return True
+
+    def __init__(self, target):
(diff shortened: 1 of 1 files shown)


devlib/module/thermal.py
```diff
@@ -0,0 +1,108 @@
...
+    @staticmethod
+    def probe(target):
+
+        if target.file_exists(ThermalModule.thermal_root):
+            return True
+
+    def __init__(self, target):
+        super(ThermalModule, self).__init__(target)
+
+        self.logger = logging.getLogger('Thermal')
```
setrofim: this is unnecessary -- `self.logger` is initialized in the super

devlib/module/thermal.py
```diff
@@ -0,0 +1,108 @@
...
+
+        for entry in target.list_directory(self.thermal_root):
+            re_match = re.match('^(thermal_zone|cooling_device)([0-9]+)', entry)
+
+            if re_match.group(1) == 'thermal_zone':
+                self.thermal_zone(re_match.group(2))
+            elif re_match.group(1) == 'cooling_device':
+                pass
+
+    def thermal_zone(self, _id):
```
setrofim: should be `add_thermal_zone`

devlib/module/thermal.py
```diff
@@ -0,0 +1,108 @@
...
+        self.zones = {}
+        self.cdevs = []
+
+        for entry in target.list_directory(self.thermal_root):
+            re_match = re.match('^(thermal_zone|cooling_device)([0-9]+)', entry)
+
+            if re_match.group(1) == 'thermal_zone':
+                self.thermal_zone(re_match.group(2))
+            elif re_match.group(1) == 'cooling_device':
+                pass
```
setrofim: Why match it and then ignore it? Is this a `TODO`?

devlib/module/thermal.py
```diff
@@ -0,0 +1,108 @@
...
+        value = 'disable'
+        if enable:
+            value = 'enabled'
+
+        self.target.write_value(self.target.path.join(self.path, 'mode', value))
+
+    def get_temperature(self):
+        """Returns the temperature of the thermal zone"""
+        temp_file = self.target.path.join(self.path, 'temp')
+        return self.target.read_value(temp_file)
```
setrofim: `self.target.read_int(temp_file)`?

devlib/module/thermal.py
```diff
@@ -0,0 +1,108 @@
...
+            return True
+        else:
+            return False
+
+    def set_mode(self, enable):
+        value = 'disable'
+        if enable:
+            value = 'enabled'
+
+        self.target.write_value(self.target.path.join(self.path, 'mode', value))
```
setrofim: Closing paren too far to the right. Should be

    self.target.write_value(self.target.path.join(self.path, 'mode'), value)

devlib/module/thermal.py
```diff
@@ -0,0 +1,108 @@
...
+        if value == 'enabled':
+            return True
+        else:
+            return False
+
+    def set_mode(self, enable):
+        value = 'disable'
+        if enable:
+            value = 'enabled'
+
```
setrofim: above could be 

    value = 'enabled' if enable else 'disabled'

devlib/module/thermal.py
```diff
@@ -0,0 +1,108 @@
...
+    def trip_point(self, _id):
+        self.trip_points[int(_id)] = TripPoint(self, _id)
+
+    def is_enabled(self):
+        """Returns a boolean representing the 'mode' of the thermal zone"""
+        value = self.target.read_value(self.target.path.join(self.path, 'mode'))
+        if value == 'enabled':
+            return True
+        else:
+            return False
```
setrofim: Could be just

      return value == 'enabled'

devlib/module/thermal.py
```diff
@@ -0,0 +1,108 @@
...
+        self.name = 'thermal_zone' + _id
+        self.path = target.path.join(root, self.name)
+        self.trip_points = {}
+
+        for entry in self.target.list_directory(self.path):
+            re_match = re.match('^trip_point_([0-9]+)_temp', entry)
+            if re_match is not None:
+                self.trip_point(re_match.group(1))
+
+    def trip_point(self, _id):
```
setrofim: should be `add_trip_point`

devlib/module/thermal.py
```diff
@@ -0,0 +1,108 @@
...
+        self.type_node = 'trip_point_' + _id + '_type'
+
+    def get_temperature(self):
+        """Returns the currently configured temperature of the trip point"""
+        temp_file = self.zone.target.path.join(self.zone.path, self.temp_node)
+        return self.zone.target.read_value(temp_file)
+
+    def set_temperature(self, temperature):
+        temp_file = self.zone.target.path.join(self.zone.path, self.temp_node)
+        self.zone.target_write_value(temp_file, temperature)
```
setrofim: I assume this should be

    self.zone.target.write_value(temp_file, temperature)

`self.zone.target` gets used a lot here; might be cleaner to expose it as `self.target` via a property:

    @property:
    def target(self):
         return self.zone.target

devlib/module/thermal.py
punitagrawal: Dropped the logger initialisation here.

devlib/module/thermal.py
punitagrawal: Changed the name

devlib/module/thermal.py
punitagrawal: Indeed, it's a TODO. I have implemented what I needed - hopefully over time we can cover most thermal related operations in this module.

devlib/module/thermal.py
punitagrawal: I hadn't discovered read_int yet. Now updated to use it.

devlib/module/thermal.py
punitagrawal: Oops! Thanks for spotting - not sure what my test was doing if this is broken.

devlib/module/thermal.py
punitagrawal: Nice! I am showing my lack of python chops here.

devlib/module/thermal.py
punitagrawal: Done.

devlib/module/thermal.py
punitagrawal: Fixed.

I've also made target a property. Perhaps, in the ThermalZone class as well I should refer to the module target. Hmm...

devlib/module/thermal.py
setrofim: Cool; could you add a comment tagged with "TODO" just so it's clear.

PR #8, opened 2015-12-15
android: fix initialization without android

JaviMerino:
In workload automation, utils.android._initialize_without_android_home()
gets android_home from adb's path.  When this code was copied to devlib,
we mistakenly dropped parsing the output of "which" and instead call
os.path.dirname() on "adb", which always returns "" and makes
_initialize_without_android_home() fail.

Make _initialize_without_android_home() parse the output of "which"
again.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -383,13 +383,14 @@ def _initialize_with_android_home(env):
 
 
 def _initialize_without_android_home(env):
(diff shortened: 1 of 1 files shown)


credp:
This looks like it will fix an issue I've also been seeing, fixable for me by defining ANDROID_HOME.

PR #16, opened 2016-01-26
Add ftrace function profiling

derkling:
NOTE: this series depends on the "add_shutils_support" series, actually the first three patches are from that series. Thus, merge it only *after* the shutils series has been accepted.

This series add support for kernel function profiling using FTrace.
The user can specify a set of functions to profile when ftrace is used.
The collection of results from different CPUs is accelerated using the shutils script to aggregate data from all the CPUs available in the system.

### DIFF
--- a/devlib/trace/ftrace.py
+++ b/devlib/trace/ftrace.py
@@ -47,8 +47,7 @@ class FtraceCollector(TraceCollector):
                  events=None,
                  buffer_size=None,
                  buffer_size_step=1000,
-                 buffer_size_file='/sys/kernel/debug/tracing/buffer_size_kb',
-                 marker_file='/sys/kernel/debug/tracing/trace_marker',
+                 tracing_path='/sys/kernel/debug/tracing',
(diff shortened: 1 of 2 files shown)


devlib/trace/ftrace.py
```diff
@@ -69,8 +75,17 @@ def __init__(self, target,
         self.start_time = None
         self.stop_time = None
         self.event_string = _build_trace_events(self.events)
+        self.function_string = _build_trace_functions(self.functions)
         self._reset_needed = True
 
+        # Setup tracing paths
+        self.available_functions_file = os.path.join(self.tracing_path, 'available_filter_functions')
```
setrofim: Should use `self.targe.path` rather than `os.path` for these as they are on-target paths and path handling on target and host may be different.

derkling:
Tested on JUNO with the usage of self.target.path and it seems to work fine.
I've just pushed a refresh of the same branch for merge, if there are not other main comments to address.

PR #17, opened 2016-01-27
FTrace: optional function and events

derkling:
This series adds the support to have a single set of trace events and profile functions to be used with kernel which supports only some of these events. At FTrace initialization time we check which tracepoints/functions are available in the target kernel and enable only those. Not available tracepoints/functions are logged and reported to the user.

The first patch fixes a small bug related to trace marker injection.

### DIFF
--- a/devlib/trace/ftrace.py
+++ b/devlib/trace/ftrace.py
@@ -125,9 +125,9 @@ class FtraceCollector(TraceCollector):
         self.start_time = time.time()
         if self._reset_needed:
             self.reset()
+        self.target.execute('{} start {}'.format(self.target_binary, self.event_string), as_root=True)
         if self.automark:
             self.mark_start()
-        self.target.execute('{} start {}'.format(self.target_binary, self.event_string), as_root=True)
         if 'cpufreq' in self.target.modules:
             self.logger.debug('Trace CPUFreq frequencies')
             self.target.cpufreq.trace_frequencies()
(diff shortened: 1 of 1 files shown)


devlib/trace/ftrace.py
```diff
@@ -103,6 +104,23 @@ def __init__(self, target,
...
+        selected_events = []
+        for event in self.events:
+            # Convert globs supported by FTrace into valid regexp globs
+            _event = event
+            if event[0] != '*':
+                _event = '*' + event
+            event_re = re.compile(_event.replace('*', '.*'))
+            # Select events matching the required ones
+            if len(filter(event_re.match, available_events)) == 0:
+                self.target.logger.warning('Event(s) [{}] not available for tracing'.format(event))
```
setrofim: This should be an exception rather than a warning -- if the client's request cannot be met, the client code should be infromed so it can handle it. (Perhaps add a `strict` flag to make the choice between ``exception``/``warning`` configurable)

derkling:
This is a refresh of the patchset which solves the issue discussed in the previous PR.

derkling:
@setrofim tested on my JUNO and that solution should be more robust than the previous one.
Lemme know if other updates are required.

devlib/trace/ftrace.py
```diff
@@ -111,9 +139,13 @@ def __init__(self, target,
             # Validate required functions to be traced
             available_functions = self.target.execute(
                     'cat {}'.format(self.available_functions_file)).splitlines()
+            selected_functions = []
             for function in self.functions:
                 if function not in available_functions:
-                    raise TargetError('Function [{}] not available for filtering'.format(function))
+                    self.target.logger.warning('Function [{}] not available for profiling'.format(function))
```
setrofim: should raise `TargetError` if `strict` here, as with events.

Issue #43, opened 2016-06-16
LinuxTarget: Circular dependency installing busybox


bjackman:
It's necessary to know the target ABI before installing the busybox binary:
```
self.busybox = self.install(os.path.join(PACKAGE_BIN_DIRECTORY, self.abi, 'busybox'))
```
However we currently require busybox to be installed before we can find the ABI:
```
value = self.execute('{} uname -m'.format(self.busybox)).strip()
```

We should either
 - consider busybox to be a hard dependency (error out in `connect` if `self.get_installed('busybox'` returns None). In this case I guess we'd no longer need to install our own busybox?

-  or find a way to determine the ABI without busybox. My current solution is to try using a native `uname` binary.

setrofim:
Resolved by https://github.com/ARM-software/devlib/commit/5880f6e9efcaf65a35e14a10f482af7510472eb6

PR #44, opened 2016-06-16
LinuxTarget: Allow determining target ABI without busybox

bjackman:
We need to know the target ABI in order to install busybox, therefore we
should attempt to find the ABI when busybox is not yet present. I'm not
aware of a clean way to do this if there's no `uname` command present,
so just raise an exception in that case.

Proposed solution to https://github.com/ARM-software/devlib/issues/43

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -525,7 +525,14 @@ class LinuxTarget(Target):
     @property
     @memoized
     def abi(self):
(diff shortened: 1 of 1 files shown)


setrofim:
Already resolved by https://github.com/ARM-software/devlib/commit/5880f6e9efcaf65a35e14a10f482af7510472eb6

PR #45, opened 2016-06-22
Improve memoize decorator

mdigiorgio:
Using the wrapt module we can improve the memoize decorator. In fact, it allows
to preserve the attributes of the memoized function, such as signature,
docstring, path to the file where the function is implemented, and so on.

Signed-off-by: Michele Di Giorgio <michele.digiorgio@arm.com>

### DIFF
--- a/devlib/utils/misc.py
+++ b/devlib/utils/misc.py
@@ -33,6 +33,7 @@ from operator import itemgetter
 from itertools import groupby
 from functools import partial
 
+import wrapt
 
 # ABI --> architectures list
 ABI_MAP = {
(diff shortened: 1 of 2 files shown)


devlib/utils/misc.py
```diff
@@ -32,6 +32,7 @@
 from operator import itemgetter
 from itertools import groupby
 from functools import partial
+import wrapt
```
setrofim: Please add a space between stdlib imports and third party ones.

Also, this is a new dependency. Please add it to setup.py

mdigiorgio: I pushed a new commit with the requested changes.

PR #49, opened 2016-07-15
android: avoid connect for USB target

derkling:
The "adb connect" command is required only for USB connected devices.
This patch is a small update which avoid the connect for non IP accessed
devices.


ep1cman:
Your commit message/PR description is incorrect. The first line should read "The "adb connect" command is required only for non-USB connected devices."

PR #54, opened 2016-08-02
cpufreq: always pass an integer to list_frequencies()

mdigiorgio:
Being list_frequencies() a memoized method, passing a string as input
argument works the first time, but subsequent calls fail apparently because the
memoized decorator always sees the same argument (probably same reference to the
string object).

This modifies every method that calls list_frequency() using a string to use an
integer instead.

Signed-off-by: Michele Di Giorgio <michele.digiorgio@arm.com>

### DIFF
--- a/devlib/module/cpufreq.py
+++ b/devlib/module/cpufreq.py
@@ -217,7 +217,8 @@ class CpufreqModule(Module):
         """
         if isinstance(cpu, int):
             cpu = 'cpu{}'.format(cpu)
-        available_frequencies = self.list_frequencies(cpu)
+        # Assume cpu is always of the form "cpu<n>"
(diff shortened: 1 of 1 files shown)


mdigiorgio:
Any comments/feedback/issues on this PR?

setrofim:
Sorry for the delay in handling this. I think this PR just works around a bigger issue in one specific case. The real issue seems to be that there is a bug in `memoized` decorator, and should be fixed there rather than inside `list_frequencies`.

I need to dig into this....

setrofim:
The issue is fixed by https://github.com/ARM-software/devlib/commit/6d854fd4dc59936142fff811e575cd034a308c43 instead.

PR #58, opened 2016-09-27
AndroidTarget: Handle ls format change in Android 7.0

credp:
We expect ls to output single column listings. Previous to Linaro Android
16.09 release, this was the default. From the 16.09 release onwards, the
default is multi-column. Unfortunately, pre-16.09 ls does not support the
'-1' option to guarantee single column output.

Test for ls supporting '-1' and use it if possible.

Signed-off-by: Chris Redpath <chris.redpath@arm.com>

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -707,6 +707,7 @@ class AndroidTarget(Target):
     conn_cls = AdbConnection
     path = posixpath
     os = 'android'
+    ls_command = ''
 
     @property
     @memoized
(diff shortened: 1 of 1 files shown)


devlib/target.py
```diff
@@ -832,8 +833,17 @@ def kick_off(self, command, as_root=None):
         else:
             raise ValueError('Background command exited before timeout; got "{}"'.format(output))
 
+    def __setup_list_directory(self):
```
setrofim: I think it may be good to have a short comment here to explain why this is necessary.

credp: ok :)

PR #62, opened 2016-10-10
cpufreq: Improve error when setting all CPUs to unsupported governor

bjackman:
Currently when using set_all_governors to try to set an unsupported
governor, the error is a somewhat cryptic "sh: echo: I/O error". Add a
check that diagnoses this error.

In order to avoid slowing down the "good" path, the check for
unsupported governors is only done if the failure has already occured.

### DIFF
--- a/devlib/module/cpufreq.py
+++ b/devlib/module/cpufreq.py
@@ -375,9 +375,18 @@ class CpufreqModule(Module):
         """
         Set the specified governor for all the (online) CPUs
         """
-        return self.target._execute_util(
+        try:
+            return self.target._execute_util(
                 'cpufreq_set_all_governors {}'.format(governor),
                 as_root=True)
+        except TargetError as e:
+            if "echo: I/O error" in str(e):
+                cpus_unsupported = [c for c in self.target.list_online_cpus()
+                                    if governor not in self.list_governors(c)]
+                raise TargetError("Governor {} unsupported for CPUs {}".format(
+                    governor, cpus_unsupported))
+            else:
+                raise
 
     def get_all_governors(self):
         """


devlib/module/cpufreq.py
```diff
@@ -375,9 +375,18 @@ def set_all_governors(self, governor):
         """
         Set the specified governor for all the (online) CPUs
         """
-        return self.target._execute_util(
+        try:
+            return self.target._execute_util(
                 'cpufreq_set_all_governors {}'.format(governor),
                 as_root=True)
+        except TargetError:
+            cpus_unsupported = [c for c in self.target.list_online_cpus() if
```
setrofim: the `if` keyword should be on the same line as the condition

devlib/module/cpufreq.py
```diff
@@ -375,9 +375,18 @@ def set_all_governors(self, governor):
         """
         Set the specified governor for all the (online) CPUs
         """
-        return self.target._execute_util(
+        try:
+            return self.target._execute_util(
                 'cpufreq_set_all_governors {}'.format(governor),
                 as_root=True)
+        except TargetError:
```
setrofim: The `TargetError`  may not necessarily have been cause by an unsupported governor (e.g. someone may have pulled a cable from the device). It's worth checking the  error output to make sure that the wrong governor is the cause.

devlib/module/cpufreq.py
bjackman: Sure, will amend this.

devlib/module/cpufreq.py
bjackman: True, but there isn't anything useful in the error output that tells you that unsupported governors is the cause. You could check for "sh: echo: I/O error" (which is what gets printed by shutils) but I think that could be caused by other stuff too, if not by someone yanking the cable..

PR #66, opened 2016-10-17
shutils.in: rename manually injected cpu_frequency trace event

mdigiorgio:
The injection of a fake cpu_frequency event in the trace may end up getting
interleaved with trace events generated by the cpufreq governor, for example:

  sh         ... cpu_frequency: state=120000 cpu_id=0
  kschedfreq ... cpu_frequency: state=120000 cpu_id=0
  kschedfreq ... cpu_frequency: state=120000 cpu_id=1
  kschedfreq ... cpu_frequency: state=120000 cpu_id=2
  kschedfreq ... cpu_frequency: state=120000 cpu_id=3
  sh         ... cpu_frequency: state=120000 cpu_id=1
  sh         ... cpu_frequency: state=120000 cpu_id=2
  sh         ... cpu_frequency: state=120000 cpu_id=3

Such a trace may confuse trace parsers during postprocessing and lead to wrong
or no results.

This patch renames the events injected by devlib so that they can be easily
recognized and then used by the trace parser in the best way.

Signed-off-by: Michele Di Giorgio michele.digiorgio@arm.com

### DIFF
--- a/devlib/bin/scripts/shutils.in
+++ b/devlib/bin/scripts/shutils.in
@@ -41,7 +41,7 @@ cpufreq_get_all_governors() {
 cpufreq_trace_all_frequencies() {
 	FREQS=$($CAT /sys/devices/system/cpu/cpu*/cpufreq/scaling_cur_freq)
 	CPU=0; for F in $FREQS; do
-		echo "cpu_frequency:        state=$F cpu_id=$CPU" > /sys/kernel/debug/tracing/trace_marker
+		echo "cpu_frequency_devlib:        state=$F cpu_id=$CPU" > /sys/kernel/debug/tracing/trace_marker
 		CPU=$((CPU + 1))
 	done
 }


setrofim:
Could you describe a bit more the issues you're running into?

IIRC, the whole point was to make these traces look like cpufreq events, so that parsers do not need to have any special knowledge of these events and treat them the same way as any other frequency transitions.

mdigiorgio:
I did this because I want to check whether the frequency for a set of CPUs belonging to a cluster is scaled at cluster level, that is all CPUs report a frequency change. To do that, I look at `N` `cpu_frequency` events, where `N` is the number of CPUs in the cluster and check that the reported frequency is the same.

If you have a trace with interleaved `cpu_frequency` events like the ones I described above it becomes impossible to verify this, or better the check always fails.

I don't think is that difficult to make the parser aware of those events. In fact, what I do is adding only those that are more recent than the ones generated by the `cpufreq` governor.

setrofim:
In this case why not just disable the frequency insertion by devlib?

Please note that there is no "the parser" for this. In making this change, you're forcing *every* existing and future parser that is works with devlib trace and is interested in frequencies be aware of these events (and how would they be?).

derkling:
We need frequency injection, both at the beginning and the end of a trace, to know for sure at post-processing time which was the frequency the experiment was executed. Without injection, it could happen that by configuration (e.g. usage of single OPP governor) of by chance (while the experiment execute) the  OPP is never changed. Again, it can happen that in the middle of an experiment the frequency is changed, but we don't know what was the original frequency.

All that considered, the solution proposed by @mdigiorgio is to differentiate manually injected events from the one generated by CPUFreq. I'm wondering, @mdigiorgio, if you cannot verify frequency domains by just considering CPUFreq generated events. Perhaps we can use the task name to know if the event was injected from userspace and in general disregard these events. In that case the only modification required is at TRAPpy side, with a simple filtering on the dataframe.

mdigiorgio:
@derkling probably we could do that in LISA as well, I'll give it a try later this week.

mdigiorgio:
OK, I addressed this issue directly in LISA... If that gets merged, I will close this PR.

bjackman:
@setrofim 

> Please note that there is no "the parser" for this. In making this change, you're forcing every existing and future parser that is works with devlib trace and is interested in frequencies be aware of these events (and how would they be?).

I agree with this but I conclude from it that these events _should_ be renamed. Other hypothetical parsers should not be tricked into reading fake events, given that there is a possibility for a race condition. IMO parsers (trappy/lisa included) should be aware of the fakeness of the events so that they can account for that possibility. 

 (Also see relevant [comment](https://github.com/ARM-software/lisa/pull/239#issuecomment-262290727) on Lisa PR)

derkling:
I support the comment from @bjackman, we had an internal discussion/review and we convinced ourself that fixing the issue from the LISA side is still subject to a possible race condition. Thus, this patch by @mdigiorgio seems to be the most safe solution.

@setrofim let us know if you need a better explanation of the issue, otherwise can you please merge this patch?

setrofim:
OK, the linked explanation makes sense. Merged.

PR #68, opened 2016-11-01
ftrace: Poke all CPUs before collecting trace

bjackman:
On some systems CPUs sometimes remain idle for several seconds. If a
trace capture begins during one of these long idle periods, that CPU's
idle state is unknown (in practice it is probably in its deepest
available state from cpuidle's perspective but that can't be known for
sure).

The solution to the equivalent problem for cpufreq is to read the
current frequencies from sysfs and inject artificial cpu_frequency
events using trace_marker (see cpu_freq_trace_all_frequencies in
bin/scripts/shutils.in). However we can't read the current idle state
from sysfs.

Instead, wake up each CPU by executing the "true" command on it via
taskset.

### DIFF
--- a/devlib/module/cpuidle.py
+++ b/devlib/module/cpuidle.py
@@ -152,3 +152,9 @@ class Cpuidle(Module):
         for state in self.get_states(cpu):
             state.disable()
 
+    def perturb_cpus(self):
+        """
+        Momentarily wake each CPU. Ensures cpu_idle events in trace file.
+        """
+        output = self.target._execute_util('cpuidle_wake_all_cpus')
+        print(output)
--- a/devlib/bin/scripts/shutils.in
+++ b/devlib/bin/scripts/shutils.in
@@ -47,6 +47,18 @@ cpufreq_trace_all_frequencies() {
 }
 
 ################################################################################
+# CPUIdle Utility Functions
+################################################################################
+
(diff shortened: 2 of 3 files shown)


devlib/module/cpuidle.py
```diff
@@ -152,3 +152,9 @@ def disable_all(self, cpu=0):
         for state in self.get_states(cpu):
             state.disable()
 
+    def trace_states(self):
```
setrofim: The name is not indicative of what the method does. Should be renamed to something like `perturb_cpus()`.

bjackman:
Pushed a new version with the function named `perturb_cpus`.

BTW, this may be a little slow on SSH connections. If that turns out to be true I'll probably do another PR that adds something to `shutils` to do it in a single command. If you'd rather wait until then, that's fine.

setrofim:
Given that this is "on the critical path", i.e. will run once measurement has already started, speed is a concern here. I'll hold off pulling this until the shutils update.

bjackman:
Pushed a new version that uses shutils. Actually running the shutils command takes roughly 25-30ms on both Juno Debian and HiKey Android. Unfortunately the whole kaboodle still takes 400ms on Juno Debian and 250-300ms on HiKey Android.

PR #69, opened 2016-11-02
android: Fixed issue using single quoted command with adb_shell

marcbonnici:
When using 'check_exit_code' and 'as_root' options for adb_shell with
a command containing single quotes, the provided command was escaped
twice. The second escaping has been avoided and some unit tests added
to check functionality.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -334,14 +334,15 @@ def adb_shell(device, command, timeout=None, check_exit_code=False,
     _check_env()
     if as_root:
         command = 'echo \'{}\' | su'.format(escape_single_quotes(command))
-    device_string = ' -s {}'.format(device) if device else ''
+    device_part = ['-s', device] if device else []
+    device_string = ' {} {}'.format(*device_part) if device_part else ''
     full_command = 'adb{} shell "{}"'.format(device_string,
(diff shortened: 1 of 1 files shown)


devlib/utils/android.py
```diff
@@ -312,18 +312,19 @@ def _ping(device):
...
         command = 'echo \'{}\' | su'.format(escape_single_quotes(command))
     device_string = ' -s {}'.format(device) if device else ''
     full_command = 'adb{} shell "{}"'.format(device_string,
                                              escape_double_quotes(command))
     logger.debug(full_command)
     if check_exit_code:
         actual_command = "adb{} shell '({}); echo \"\n$?\"'".format(device_string,
-                                                                    escape_single_quotes(command))
-        raw_output, error = check_output(actual_command, timeout, shell=True)
+                                                                    escape_single_quotes(orig_command if as_root else command))
```
setrofim: Rather than saving `orig_command`, why not  just do

      command if as_root else escape_single_quotes(command))

here?

setrofim: I would also move that logic out into its own line

marcbonnici: Ok, I've added the optional dependency and split the logic but left the saving of the orig_command.

PR #70, opened 2016-11-10
cgroups: ensure CGroups are mounted RW

derkling:
Some systems mount CGroups RO, thus we need to ensure we remount that
FS as root otherwise we cannot create new groups.


bjackman:
Tested (rebased on master) :+1:

PR #71, opened 2016-11-15
target: Note return value in Target::invoke docstring

bjackman:
I can't get sphinx to build the docs to check if this is the right syntax.. it feels too trivial to spend any more time on so feel free to just close this if it looks wrong.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -288,6 +288,7 @@ class Target(object):
                   a ``TimeoutError`` exception will be raised. Set to ``None`` if the
                   invocation should not timeout.
 
+        :returns: output of command.
         """
         command = binary
         if args:


devlib/target.py
```diff
@@ -288,6 +288,7 @@ def invoke(self, binary, args=None, in_directory=None, on_cpus=None,
                   a ``TimeoutError`` exception will be raised. Set to ``None`` if the
                   invocation should not timeout.
 
+        :returns: output of command.
```
derkling: Looks good to me, it's the same syntax we use already for example in:
https://github.com/ARM-software/devlib/blob/master/devlib/utils/serial_port.py#L100

:+1:  for merge

PR #72, opened 2016-11-18
android: Fix whitespace in wildcard match for ADB pull

bjackman:
I have observed that when there are multiple matches for the wildcard, `line` has a space at the end. That means the `pull` command doesn't work.


derkling:
Given the volatility of the github comments, @bjackman: it would be nice if you can copy&paste the same comment above in the commit message of this patch.

bjackman:
Good idea, pushed + amended.

PR #73, opened 2016-11-23
target: Use root in file_exists check

bjackman:
Use case: To check if a kernel supports function profiling we need to
check for the presence of
"/sys/kernel/debug/tracing/function_profiler_enabled", but
"/sys/kernel/debug/"'s permissions are 700.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -362,7 +362,8 @@ class Target(object):
 
     def file_exists(self, filepath):
         command = 'if [ -e \'{}\' ]; then echo 1; else echo 0; fi'
-        return boolean(self.execute(command.format(filepath)).strip())
+        output = self.execute(command.format(filepath), as_root=self.is_rooted)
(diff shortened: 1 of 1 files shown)


derkling:
Seems ok to me, :+1 for merge.

devlib/target.py
```diff
@@ -361,7 +361,8 @@ def ps(self, **kwargs):
 
     def file_exists(self, filepath):
         command = 'if [ -e \'{}\' ]; then echo 1; else echo 0; fi'
-        return boolean(self.execute(command.format(filepath)).strip())
+        output = self.execute(command.format(filepath), as_root=True)
```
setrofim: This change makes `file_exists` unusable on unrooted devices. Should be

    output = self.execute(command.format(filepath), as_root=self.is_rooted)

bjackman: Amended and pushed.

PR #74, opened 2016-11-25
devlib: strip space for adb pull source dir

arnoldlu:
There may be space at the end of the string line.
This will cause adb pull failed.


arnoldlu:
Please help to verify and merge the patch.

PR #75, opened 2016-11-25
devlib: strip space for adb pull source dir

arnoldlu:
There may be space at the end of the string line.
This will cause adb pull failed.


derkling:
That's quite similar/related to #72, perhaps it's better to keep this patch in the same PR?
Maybe you can move it in the previous PR and abandon this one?

arnoldlu:
Thanks @derkling .

Close this, because it fix the same issue as #72 .

PR #79, opened 2016-12-07
Cpuidle additions

bjackman:


devlib/module/cpuidle.py
```diff
@@ -61,13 +61,25 @@ def name(self):
     @property
     @memoized
     def latency(self):
+        """State exit latency in uS"""
         return self.get('latency')
 
     @property
     @memoized
     def power(self):
+        """State power usage in mW"""
```
bjackman: AFAICT this value is actually garbage on ARM. Wasn't sure whether to mention that here...

setrofim: It's pobably worth adding as a ..note::

setrofim: Remove the verb "State" from the description.

devlib/module/cpuidle.py
```diff
@@ -61,13 +61,25 @@ def name(self):
...
         return self.get('power')
 
+    @property
+    @memoized
+    def target_residency(self):
+        """State target residency in uS
```
setrofim: Remove the verb "State" from the description.

bjackman:
Pushed a new version reworded without "state" and with a note about the power property probably being garbage.

PR #83, opened 2017-01-09
Gem5 integration

avanlaer:
#These 4 commits allow you to use gem5 as a target in devlib.
This has only been tested with gem5 simulations running Linux, both independently and using lisa (see pull request #265 in the lisa repository).

I can't seem to figure out how to ask for reviewers so @setrofim @Sticklyman1936 would you mind looking at this?


devlib/platform/arm.py
```diff
@@ -292,3 +294,162 @@ def take_measurement(self):
             result.append(Measurement(value, chan))
         return result
 
+class Gem5SimulationPlatform(Platform):
```
marcbonnici: Gem5 isn't ARM specific so move this to a new module.

devlib/platform/arm.py
```diff
@@ -25,7 +28,6 @@
 from devlib.host import PACKAGE_BIN_DIRECTORY
 from devlib.utils.serial_port import open_serial_connection
 
-
```
marcbonnici: Should be 2 blank lines before the beginning of a class.

devlib/platform/arm.py
```diff
@@ -292,3 +294,162 @@ def take_measurement(self):
...
+
+    def _check_gem5_command(self):
+        """
+        Check if the command to start gem5 makes sense
+        """
+        assert(self.gem5args_binary is not None)
+        assert(self.gem5args_args is not None)
+        self.gem5args_virtio = str(self.gem5args_virtio).format(self.gem5interactdir)
+        assert(self.gem5args_virtio is not None)
+
```
marcbonnici: The `assert` statements should `raise` errors instead.

devlib/platform/arm.py
```diff
@@ -292,3 +294,162 @@ def take_measurement(self):
...
+        # First create the input and output directories for gem5
+        if self.start_gem5_simulation:
+            self.logger.info("Creating temporary directory: {}".format(self.gem5interactdir))
+            if not os.path.exists(self.stats_directory):
+                os.mkdir(self.stats_directory)
+            os.mkdir(self.gem5interactdir)
+            if os.path.exists(self.gem5outdir):
+               self.logger.debug("The output directory {} already existed." \
+                                 " Previous results will be overwritten.".format(self.gem5outdir))
```
marcbonnici: What are these folders used for? 
Should the contents be overridden?

devlib/platform/arm.py
```diff
@@ -292,3 +294,162 @@ def take_measurement(self):
...
+            else:
+                os.mkdir(self.gem5outdir)
+
+            # We need to redirect the standard output and standard error for the
+            # gem5 process to a file so that we can debug when things go wrong.
+            f = os.path.join(self.gem5outdir, 'stdout')
+            self.stdout_file = open(f, 'w')
+            f = os.path.join(self.gem5outdir, 'stderr')
+            self.stderr_file = open(f, 'w')
+            # We need to keep this so we can check which port to use for the telnet
```
marcbonnici: Change these to use `with` statements.

devlib/platform/arm.py
```diff
@@ -292,3 +294,162 @@ def take_measurement(self):
...
+        """
+
+        assert(self.gem5 is not None)
+        while self.gem5_port == -1:
+            # Check that gem5 is running!
+            if self.gem5.poll():
+                raise TargetError("The gem5 process has crashed with error code {}!".format(self.gem5.poll()))
+
+            # Open the stderr file
+            f = open(self.stderr_filename, 'r')
```
marcbonnici: Change this to use a `with` statement.

devlib/target.py
```diff
@@ -70,6 +71,8 @@ def connected_as_root(self):
     @property
     @memoized
     def is_rooted(self):
+        if self.conn.connected_as_root:
+            return True
```
marcbonnici: Does this need to be set here? 
If so, the android implementation needs to be updated to be consistent.

devlib/target.py
```diff
@@ -165,7 +168,6 @@ def __init__(self,
         self._cache = {}
         self._connections = {}
         self.busybox = None
-
```
marcbonnici: Should have a blank line for formatting.

devlib/target.py
```diff
@@ -188,6 +190,10 @@ def connect(self, timeout=None):
         self._update_modules('connected')
         if self.platform.big_core and self.load_default_modules:
             self._install_module(get_module('bl'))
+        if self.busybox is not None:
+            # Busybox is used on this target so pass on to connections
+            # Some connections (i.e. gem5) need to wrap some commands
+            self.conn.set_busybox_wrapper(self.busybox)
```
marcbonnici: Does this need to be set here? 
If so, the android implementation needs to be updated to be consistent.

devlib/target.py
```diff
@@ -197,7 +203,7 @@ def disconnect(self):
     def get_connection(self, timeout=None):
         if self.conn_cls == None:
             raise ValueError('Connection class not specified on Target creation.')
-        return self.conn_cls(timeout=timeout, **self.connection_settings)  # pylint: disable=not-callable
+        return self.conn_cls(timeout=timeout, platform=self.platform, **self.connection_settings)  # pylint: disable=not-callable
```
marcbonnici: Change the `platform` parameter to be a part of `connection_settings`.

devlib/target.py
```diff
@@ -263,14 +272,15 @@ def _execute_util(self, command, timeout=None, check_exit_code=True, as_root=Fal
         command = '{} {}'.format(self.shutils, command)
         return self.conn.execute(command, timeout, check_exit_code, as_root)
 
-    def execute(self, command, timeout=None, check_exit_code=True, as_root=False):
-        return self.conn.execute(command, timeout, check_exit_code, as_root)
+    def execute(self, command, timeout=None, check_exit_code=True,
+                as_root=False, wrap_busybox=False):
```
marcbonnici: This needs to be updated for `AdbConnection` for android as well.

devlib/target.py
```diff
@@ -913,6 +924,7 @@ def ps(self, **kwargs):
             return filtered_result
 
     def capture_screen(self, filepath):
+        # TODO adapt this to gem5 later on
```
marcbonnici: The `todo` shouldn't be here.

devlib/utils/ssh.py
```diff
@@ -175,7 +180,8 @@ def pull(self, source, dest, timeout=30):
         source = '{}@{}:{}'.format(self.username, self.host, source)
         return self._scp(source, dest, timeout)
 
-    def execute(self, command, timeout=None, check_exit_code=True, as_root=False, strip_colors=True):
+    def execute(self, command, timeout=None, check_exit_code=True,
+                as_root=False, strip_colors=True, wrap_busybox=False):
```
marcbonnici: The `wrap_busybox` and `strip_colors` parameter order is different to the `execute` method for gem5connection.

devlib/utils/ssh.py
```diff
@@ -299,6 +312,469 @@ def __init__(self,
...
+                 password_prompt=None,
+                 original_prompt=None,
+                 ):
+        # TODO: need to insert check that host is the actual place
+        # where devlib is now running
+        self.host = host
+        if username is not None or username is not 'root':
+            gem5_logger.info("User is always root in gem5")
+        if password is not None or password is not '':
+            gem5_logger.info("No password needed in gem5")
```
marcbonnici: This should be changed to raise a config error if they are not valid rather than set debugging message.

devlib/utils/ssh.py
```diff
@@ -299,6 +312,469 @@ def __init__(self,
...
+        self.username =  'root'
```
marcbonnici: Extra space after '='.

devlib/utils/ssh.py
```diff
@@ -299,6 +312,469 @@ def __init__(self,
...
+        self.password = None
+        self.port = -1
+        # Long timeouts to account for gem5 being slow
+        self.default_timeout = 3600
```
marcbonnici: Should allow for timeout to be overridden via parameter.

devlib/utils/ssh.py
```diff
@@ -299,6 +312,469 @@ def __init__(self,
...
+
+
+    def set_busybox_wrapper(self, wrapper):
+        self.busybox_wrapper = wrapper
+
+    def connected_as_root(self):
+        # gem5 is always rooted
+        return True
+
+    def push(self, source, dest, timeout=None, wrap_busybox=False):
```
marcbonnici: Does `wrap_busybox` need to be passed to this function?
This needs to be updated in android adb connection as well if required.

devlib/utils/ssh.py
```diff
@@ -299,6 +312,469 @@ def __init__(self,
...
+        self._move_to_temp_dir(source)
+
+        # Back to the gem5 world
+        self._gem5_shell("ls -al /mnt/obb/{}".format(filename))
+        if wrap_busybox:
+            self._gem5_shell("{} cp /mnt/obb/{} {}".format(self.busybox_wrapper, filename, dest))
+        else:
+            self._gem5_shell("cat /mnt/obb/{} > {}".format(filename, dest))
+        self._gem5_shell("sync")
+        self._gem5_shell("ls -al {}".format(dest))
```
marcbonnici: Is the path `/mnt/obb` going to be consistent?

devlib/utils/ssh.py
```diff
@@ -299,6 +312,469 @@ def __init__(self,
...
+        self._gem5_shell("ls -al /mnt/obb/")
+        logger.debug("Push complete.")
+
+    def pull(self, source, dest, timeout=0):
```
marcbonnici: If `push` is going to take  the `wrap_busybox` parameter, `pull` should also take it for consistency.

devlib/utils/ssh.py
```diff
@@ -299,6 +312,469 @@ def __init__(self,
...
+        if len(screen_caps) == 1:
+            # Bail out if we do not have image, and resort to the slower, built
+            # in method.
+            try:
+                import Image
+                gem5_image = os.path.join(self.gem5outdir, screen_caps[0])
+                temp_image = os.path.join(self.gem5outdir, "file.png")
+                im = Image.open(gem5_image)
+                im.save(temp_image, "PNG")
+                shutil.copy(temp_image, filepath)
```
marcbonnici: Is `filepath` defined here?

devlib/utils/ssh.py
```diff
@@ -299,6 +312,469 @@ def __init__(self,
...
+        pass
+
+
+    def _probe_file(self, filepath):
+        """
+        Internal method to check if the target has a certain file
+        """
+        command = 'if [ -e \'{}\' ]; then echo 1; else echo 0; fi'
+        output = self.execute(command.format(filepath), as_root=self.is_rooted())
+        return boolean(output.strip())
```
marcbonnici: Has `boolean` been imported from `devlib.utils.types`?

devlib/utils/ssh.py
```diff
@@ -299,6 +312,469 @@ def __init__(self,
...
+        if as_root and not self.is_rooted:
+            raise TargetError('Attempting to execute "{}" as root on unrooted device.'.format(command))
+        if wrap_busybox:
+            if not self.is_rooted:
+                raise TargetError('Attempting to execute "{}" with busybox. '.format(command) +
+                                  'Busybox can only be deployed to rooted devices.')
+            command = ' '.join([self.busybox_wrapper, command])
+        if background:
+            logger.debug("Attempt to execute in background. Not supported "
+                              "in gem5, hence ignored.")
```
marcbonnici: This should raise an error to alert the user.

devlib/target.py
setrofim: This should not be set here at all if a device has been previously set up, then the `get_installed` call above will not return `None`. If at this point busybox is `None`, then the device has not been set up, and it will be set when `setup()` is called.

devlib/platform/arm.py
```diff
@@ -292,3 +294,162 @@ def take_measurement(self):
...
+            from devlib.utils.ssh import AndroidGem5Connection
+            target.conn_cls = AndroidGem5Connection
+
+    def setup(self, target):
+        """
+        Deploy m5
+        """
+        host_executable = os.path.join(PACKAGE_BIN_DIRECTORY, target.abi, 'm5')
+        m5_path = target.install(host_executable)
+        target.conn.m5_path = m5_path
```
setrofim: As per

http://pythonhosted.org/devlib/target.html#Target.setup

`setup()` may not be invoked on every instantiation of a target. If it isn't, `m5_path` will never be set to the correct values. There needs to be a `target.get_installed('m5')` inside `update_from_target` of the Platform.

devlib/platform/arm.py
```diff
@@ -292,3 +294,162 @@ def take_measurement(self):
...
+        # Start setting up the gem5 parameters/directories
+        # The gem5 subprocess
+        self.gem5 = None
+        self.gem5_port = gem5_telnet_port or -1
+        self.stats_directory = host_output_dir
+        self.gem5outdir = os.path.join(self.stats_directory, "gem5")
+        self.gem5interactdir = '/tmp'
+        self.executable_dir = None
+        self.working_dir = None
+        self.m5_path = 'm5'
```
setrofim: Should be just `None` here (see comments below).

devlib/platform/arm.py
avanlaer: This will change in commit a3b080b.

devlib/platform/arm.py
avanlaer: This will change in commit a3b080b.

devlib/platform/arm.py
avanlaer: This will change in commit a3b080b.

devlib/platform/arm.py
avanlaer: These folders are used for the output of gem5. I made the messages are bit more descriptive in commit a3b080b.

setrofim: The point here is -- will that directory contain the final output that is of interest to the user, or is this a transitional location and "interesting" output would be transferred to a different location?

If it is the latter, then the current implementation is fine. If it is the latter then an error should be raise here rather than removing a directory that potentially has data that the user may want.

devlib/platform/arm.py
avanlaer: This will change in commit a3b080b.

devlib/platform/arm.py
avanlaer: This will change in commit a3b080b.

devlib/platform/arm.py
avanlaer: This will change in commit a3b080b.

devlib/platform/arm.py
avanlaer: This will change in commit a3b080b.

devlib/platform/arm.py
```diff
@@ -292,3 +294,162 @@ def take_measurement(self):
...
+    def init_target_connection(self, target):
+        """
+        Update the type of connection in the target from here
+        """
+        if target.os == 'linux':
+            from devlib.utils.ssh import LinuxGem5Connection
+            target.conn_cls = LinuxGem5Connection
+        else:
+            from devlib.utils.ssh import AndroidGem5Connection
+            target.conn_cls = AndroidGem5Connection
```
marcbonnici: These should both be imported at the beginning of the code and just set appropriately here.

devlib/platform/arm.py
avanlaer: This will change in commit a3b080b.

devlib/platform/arm.py
avanlaer: I don't think that works here as you open these files here and then pass them on to the gem5 subprocess, which is the one that will actually write to these files.

devlib/platform/arm.py
avanlaer: I placed the imports there to prevent the formation of an import loop: the connection needs to be able to access some platform variables so Gem5ConnectionPlatform is imported in Gem5Connection. Is this okay?

setrofim: Could you link to the line you mean? I don't see a `Gem5ConnectionPlatform` import for the connection. From what I understand, an existing connection instance is passed as an argument to the connection's `__init__`.

devlib/platform/arm.py
avanlaer: This directory will contain all the output from gem5 (like stats and the configuration used) so you might be overwriting older results there. I'll change it to raise an error.

devlib/platform/arm.py
avanlaer: In line 37 of utils/ssh.py. I assumed that import was needed as _connect_gem5 will use some variables of the platform.

setrofim:
Right I see. That import is not necessary and should be removed.
`Gem5SimulationPlatform` is not being used anywhere inside
`utils/ssh.py`.

Python, being dynamic language, resolves attribute accesses at runtime.
So as long as the object that gets bound to `platform` parameter at run
time has a `gem5_port` attribute (and the others), the code will work.
(Note that the object doesn't even need to be an instance of
`Gem5SimulationPlatform`.)

Since the `Gem5SimulationPlatform` class itself is not being referenced
in the code, it does not need to be imported.

On 11 Jan 2017, 09:39:54 -0800, Anouk Van Laer wrote:
> avanlaer commented on this pull request.
>
>
>
> > +                        f.close()
> +                        break
> +            else:
> +                time.sleep(1)
> +            f.close()
> +
> +    def init_target_connection(self, target):
> +        """
> +        Update the type of connection in the target from here
> +        """
> +        if target.os == 'linux':
> +            from devlib.utils.ssh import LinuxGem5Connection
> +            target.conn_cls = LinuxGem5Connection
> +        else:
> +            from devlib.utils.ssh import AndroidGem5Connection
> +            target.conn_cls = AndroidGem5Connection
>
> In line 37 of utils/ssh.py. I assumed that import was needed as _connect_gem5 will use some variables of the platform.
>
> --
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/ARM-software/devlib/pull/83
IMPORTANT NOTICE: The contents of this email and any attachments are confidential and may also be privileged. If you are not the intended recipient, please notify the sender immediately and do not disclose the contents to any other person, use it for any purpose, or store or copy the information in any medium. Thank you.

devlib/platform/arm.py
avanlaer: Ok, will remove the import there! Thanks

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
```diff
@@ -299,6 +312,469 @@ def __init__(self,
...
+        # Wait for boot
+        self._wait_for_boot()
+
+        # Mount the virtIO to transfer files in/out gem5 system
+        self._mount_virtio()
+
+        # Ready for interaction now
+        self.ready = True
+
+
```
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: It is only used within this object. If you want I can change it to a variable?

setrofim: The question is -- is it reasonable to expect the path to always be `"/mnt/obb/"`, or would it need to be tweaked based on how a particular gem5 simulation is set up?

avanlaer: That folder is used to interact with the gem5 emulated system and I don't think it depends on the exact system you emulate, so it will stay constant. What do you think, @Sticklyman1936?

Sticklyman1936: OK, so whilst this is likely OK to use, I think that we should change it. Under Linux, it would be fine, but for Android `/mnt/obb` is used as a temporary file system in certain situations. I think that we should change this to something else if possible. How about something along the lines of `/mnt/host`, but we would need to make sure that we are su at the point where we create and mount this node. Any thoughts on the name?

For the record, we would need to make the same change in WA.

devlib/target.py
avanlaer: Fixed in both the AdbConnection & LocalConnection

devlib/target.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed in ADB connection & LocalConnection

devlib/utils/ssh.py
avanlaer: I've changed this to a variable (hostinteractdir, to mirror the gem5interactdir on the host system), for which the default value is /mnt/host

setrofim: Erm I would rethink the variable name -- it makes it sound like it is a location on the host. I would avoid the word "host" for on-target names.

Also please avoid names like `hostintectactdir`, as they are very hard to parse. Use underscore to separate words -- `host_interact_dir` is much more readable.

devlib/target.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fair enough, didn't think about that. Both are changed.

devlib/target.py
avanlaer: Fixed

avanlaer:
I've pushed my changes to my gem5_integration branch. The date on the commits is the original date, but the actual changes incorporate the comments made above.

devlib/utils/android.py
```diff
@@ -216,7 +219,8 @@ def pull(self, source, dest, timeout=None):
         command = "pull '{}' '{}'".format(source, dest)
         return adb_command(self.device, command, timeout=timeout)
 
-    def execute(self, command, timeout=None, check_exit_code=False, as_root=False):
+    def execute(self, command, timeout=None, check_exit_code=False,
+                as_root=False, wrap_busybox=False, strip_colors=True):
```
setrofim: `warp_busybox` parameter should be removed from here, and elsewhere -- this should be be specified at connection level, but on construction of the command to be executed in the one or two places where this is actually needed.

devlib/utils/ssh.py
```diff
@@ -219,6 +224,15 @@ def cancel_running_command(self):
                 return True
         return False
 
+    def connected_as_root(self):
+        pass
+
+    def set_busybox_wrapper(self, wrapper):
+        pass
+
+    def capture_screen(self, filepath): #pylint: disable=unused-argument
```
setrofim: `capure_screen` should not be part of a connection

devlib/utils/ssh.py
```diff
@@ -219,6 +224,15 @@ def cancel_running_command(self):
                 return True
         return False
 
+    def connected_as_root(self):
+        pass
+
+    def set_busybox_wrapper(self, wrapper):
```
setrofim: Connections shouldn't know anything about busybox, because it's nothing to do with the actual connection

devlib/utils/ssh.py
```diff
@@ -167,15 +171,16 @@ def __init__(self,
         timeout = timeout if timeout is not None else self.default_timeout
         self.conn = ssh_get_shell(host, username, password, self.keyfile, port, timeout, False, None)
 
-    def push(self, source, dest, timeout=30):
+    def push(self, source, dest, timeout=30, wrap_busybox=False): #pylint: disable=unused-argument
```
setrofim: remove `wrap_busybox`

devlib/utils/ssh.py
```diff
@@ -167,15 +171,16 @@ def __init__(self,
...
         dest = '{}@{}:{}'.format(self.username, self.host, dest)
         return self._scp(source, dest, timeout)
 
-    def pull(self, source, dest, timeout=30):
+    def pull(self, source, dest, timeout=30, wrap_busybox=False): #pylint: disable=unused-argument
```
setrofim: remove `wrap_busybox`

devlib/utils/ssh.py
```diff
@@ -167,15 +171,16 @@ def __init__(self,
...
         source = '{}@{}:{}'.format(self.username, self.host, source)
         return self._scp(source, dest, timeout)
 
-    def execute(self, command, timeout=None, check_exit_code=True, as_root=False, strip_colors=True):
+    def execute(self, command, timeout=None, check_exit_code=True,
+                as_root=False, wrap_busybox=False, strip_colors=True): #pylint: disable=unused-argument
```
setrofim: remove `wrap_busybox`

devlib/utils/ssh.py
```diff
@@ -299,6 +313,498 @@ def __init__(self,
...
+                 username=None,
+                 password=None,
+                 port=None,
+                 timeout=None,
+                 password_prompt=None,
+                 original_prompt=None,
+                 ):
+        # TODO: need to insert check that host is the actual place
+        # where devlib is now running
+        self.host = host
```
setrofim: Why is this being set here? It does not appear to be used.

devlib/utils/ssh.py
setrofim: It shouldn't be required

devlib/utils/ssh.py
```diff
@@ -299,6 +313,498 @@ def __init__(self,
...
+        written out to the host system using writefile, prior to being moved to
+        the destination on the host.
+        """
+        filename = os.path.basename(source)
+
+        logger.debug("pull_file {} {}".format(source, filename))
+        # We don't check the exit code here because it is non-zero if the source
+        # and destination are the same. The ls below will cause an error if the
+        # file was not where we expected it to be.
+        self._gem5_shell("{} cp {} {}".format(self.busybox_wrapper, source,
```
setrofim: why is busybox needed here?

Also, the command should be `"cp '{}' '{}'"`. Source and/or destination paths may have spaces, parentheses, or other shell characters in their names.

devlib/utils/ssh.py
```diff
@@ -299,6 +313,498 @@ def __init__(self,
...
+        if 'cpu' not in filename:
+            while not os.path.exists(os.path.join(self.gem5_out_dir, filename)):
+                time.sleep(1)
+
+        # Perform the local move
+        shutil.move(os.path.join(self.gem5_out_dir, filename), dest)
+        logger.debug("Pull complete.")
+
+    def execute(self, command, timeout=1000, check_exit_code=True,
+                as_root=False, wrap_busybox=False, strip_colors=True):
```
setrofim: `wrap_busybox`

devlib/platform/arm.py
```diff
@@ -13,9 +13,9 @@
 # limitations under the License.
```
setrofim: Changes in this file are nothing to do with a Gem5Platform, and their purpose is not clear. Is intended to be part of the commit?

devlib/platform/gem5.py
```diff
@@ -0,0 +1,205 @@
...
+    def __init__(self, name, host_output_dir, gem5_bin, gem5_args, gem5_virtio,
+                 gem5_telnet_port=None):
+
+        # First call the parent class
+        super(Gem5SimulationPlatform, self).__init__(name=name)
+
+        # Start setting up the gem5 parameters/directories
+        # The gem5 subprocess
+        self.gem5 = None
+        self.gem5_port = gem5_telnet_port or -1
```
setrofim: Why `-1`? The usual value for "unspecified" is `None`.

devlib/utils/ssh.py
```diff
@@ -299,6 +313,498 @@ def __init__(self,
...
+        # codes from these command because some of them WILL fail.
+        self._gem5_shell('stty columns 1024', check_exit_code=False)
+        self._gem5_shell('{} stty columns 1024'.format(self.busybox_wrapper), check_exit_code=False)
+        self._gem5_shell('stty cols 1024', check_exit_code=False)
+        self._gem5_shell('{} stty cols 1024'.format(self.busybox_wrapper), check_exit_code=False)
+        self._gem5_shell('reset', check_exit_code=False)
+
+    def _gem5_util(self, command):
+        """ Execute a gem5 utility command using the m5 binary on the device """
+        self._gem5_shell('{} {}'.format(self.m5_path, command))
```
setrofim: There is a missing check here. It is not guaranteed that `self.m5_path` is set by this point. This needs to be verified and a `TargetError` raised if that is not the case.

devlib/utils/ssh.py
```diff
@@ -299,6 +313,498 @@ def __init__(self,
...
+                logger.info('Ignoring the given timeout --> gem5 needs longer timeouts')
+        self.ready_timeout = self.default_timeout * 3
+        # Intermediate directory to push things to gem5 using VirtIO
+        self.gem5_interact_dir = None
+        # Counterpart in gem5
+        self.gem5_input_dir = '/mnt/host/'
+        # Directory to store things from gem5 on the host
+        self.gem5_out_dir = None
+        self.busybox_wrapper = ''
+        self.m5_path = None
```
setrofim: This should be added to the parameter list. Even though, in the current usage,  this not being set on init and is updated later, `Gem5Connection` is a stand-alone class that may be used in contexts other than that of a Target/Platform, and there needs to be a clear way to configure it.

devlib/platform/gem5.py
```diff
@@ -0,0 +1,205 @@
...
+        assert self.gem5 is not None
+        while self.gem5_port == -1:
+            # Check that gem5 is running!
+            if self.gem5.poll():
+                raise TargetError("The gem5 process has crashed with error code {}!".format(self.gem5.poll()))
+
+            # Open the stderr file
+            with open(self.stderr_filename, 'r') as f:
+                for line in f:
+                    m = re.search(r"Listening\ for\ system\ connection\ on\ port\ (?P<port>\d+)", line)
```
setrofim: Erm, why are the spaces escaped here?

devlib/platform/gem5.py
```diff
@@ -195,6 +195,16 @@ def update_from_target(self, target):
         m5_path = target.get_installed('m5')
         target.conn.m5_path = m5_path
 
+        # Overwrite the following methods (monkey-patching)
+        self.logger.debug("Overwriting the 'capture_screen' method in target")
```
setrofim: These changes should be part of the previous commit.

devlib/target.py
```diff
@@ -203,6 +205,9 @@ def setup(self, executables=None):
         self.execute('mkdir -p {}'.format(self.working_directory))
         self.execute('mkdir -p {}'.format(self.executables_directory))
         self.busybox = self.install(os.path.join(PACKAGE_BIN_DIRECTORY, self.abi, 'busybox'))
+        # Busybox is used on this target so pass on to connections
+        # Some connections (i.e. gem5) need to wrap some commands
+        self.conn.set_busybox_wrapper(self.busybox)
```
setrofim: busybox...

devlib/target.py
```diff
@@ -263,14 +271,15 @@ def _execute_util(self, command, timeout=None, check_exit_code=True, as_root=Fal
...
+    def execute(self, command, timeout=None, check_exit_code=True,
+                as_root=False, wrap_busybox=False):
+        return self.conn.execute(command, timeout, check_exit_code, as_root, wrap_busybox)
 
     def background(self, command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, as_root=False):
         return self.conn.background(command, stdout, stderr, as_root)
 
     def invoke(self, binary, args=None, in_directory=None, on_cpus=None,
-               as_root=False, timeout=30):
+               as_root=False, timeout=30, wrap_busybox=False):
```
setrofim: No busybox

devlib/target.py
```diff
@@ -582,6 +592,22 @@ def _install_module(self, mod, **params):
     def _resolve_paths(self):
         raise NotImplementedError()
 
+    # Some connections (i.e. gem5) have an internal, more efficient method of
+    # capturing the screen so try that first
+    def _overwritten_capture_screen(self, filepath):
```
setrofim: This should be implemented in the platform  and monkey patched in to the target.

devlib/target.py
```diff
@@ -582,6 +592,22 @@ def _install_module(self, mod, **params):
...
+        connection_screencapped = self.conn.capture_screen(filepath)
+        if connection_screencapped == False:
+            # The connection was not able to capture the screen so use the target
+            # implementation
+            self.logger.debug('{} was not able to screen cap, using the original target implementation'.format(self.conn.__class__.__name__))
+            self.target_impl_capture_screen(filepath)
+
+    def _overwritten_reset(self):
```
setrofim: ditto for these two

devlib/host.py
```diff
@@ -34,15 +34,19 @@ def __init__(self, keep_password=True, unrooted=False, password=None):
         self.unrooted = unrooted
         self.password = password
 
-    def push(self, source, dest, timeout=None, as_root=False):  # pylint: disable=unused-argument
+    def is_rooted(self):
+        pass
```
avanlaer: Fixed

devlib/utils/android.py
```diff
@@ -195,13 +195,16 @@ def __init__(self, device=None, timeout=None):
         AdbConnection.active_connections[self.device] += 1
         self._setup_ls()
 
-    def push(self, source, dest, timeout=None):
+    def is_rooted(self):
+        pass
```
avanlaer: In the end, I removed the need for the connections to be able to say whether or not they are rooted. It was called in the target (is_rooted property) and as gem5 is always rooted, I added this 'shortcut', which is actually not needed.

devlib/target.py
```diff
@@ -153,6 +155,7 @@ def __init__(self,
                  ):
         self.connection_settings = connection_settings or {}
         self.platform = platform or Platform()
+        self.connection_settings['platform'] = self.platform
```
marcbonnici: The `platform` argument needs to be added in all the other connections (e.g. `ADBConnection`).
Also there should be a check here to ensure that platform is not already set in `connection_settings`.

devlib/utils/ssh.py
```diff
@@ -299,6 +313,498 @@ def __init__(self,
...
+        self.port = -1
+        # Long timeouts to account for gem5 being slow
+        # Can be overriden if the given timeout is longer
+        self.default_timeout = 3600
+        if timeout is not None:
+            if timeout > self.default_timeout:
+                logger.info('Overwriting the default timeout of gem5 ({})'
+                                 ' to {}'.format(self.default_timeout, timeout))
+                self.default_timeout = timeout
+            else :
```
marcbonnici: Shouldn't be a space after `else`

devlib/utils/ssh.py
```diff
@@ -299,6 +313,498 @@ def __init__(self,
...
+        logger.debug("filename: {}".format(filename))
+
+        # We need to copy the file to copy to the temporary directory
+        self._move_to_temp_dir(source)
+
+        # Dest in gem5 world is a file rather than directory
+        if os.path.basename(dest) != filename:
+            dest = os.path.join(dest, filename)
+        # Back to the gem5 world
+        self._gem5_shell("ls -al {}{}".format(self.gem5_input_dir,filename))
```
marcbonnici: Should have a space between `,` and `filename`.

devlib/utils/ssh.py
```diff
@@ -299,6 +313,498 @@ def __init__(self,
...
+        the destination on the host.
+        """
+        filename = os.path.basename(source)
+
+        logger.debug("pull_file {} {}".format(source, filename))
+        # We don't check the exit code here because it is non-zero if the source
+        # and destination are the same. The ls below will cause an error if the
+        # file was not where we expected it to be.
+        self._gem5_shell("{} cp {} {}".format(self.busybox_wrapper, source,
+                                              filename),check_exit_code=False)
```
marcbonnici: Space between `,` and `check_exit_code`.

devlib/utils/android.py
avanlaer: Removed these, it was a muddled up combination of workload automation code and my understanding of it.

devlib/utils/ssh.py
```diff
@@ -34,14 +35,14 @@
...
-
+from devlib.utils.types import boolean
 
 ssh = None
 scp = None
 sshpass = None
 
 logger = logging.getLogger('ssh')
-
+gem5_logger = logging.getLogger('gem5-connection')
```
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: I took this code from workload-automation as this was the way it was done in gem5 devices there and so I assumed this was needed in gem5, @Sticklyman1936?
This is the reason for busybox being passed to the gem5 connection

devlib/platform/arm.py
avanlaer: This was because in the first version the Gem5SimulationPlatform was placed in this file

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/target.py
avanlaer: Fixed

devlib/target.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Changed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: But then the init method here should not contain the platform either?
So would something like the following be okay:

- if you pass a platform during initialisation of the connection, the connection connects to it automatically (_connect_gem5) and those attributes are set by that method
- or you pass no platform and the values for these attributes (eg gem5_out_dir) are provided by the init arguments and another connection method is added, that allows you to connect to a gem5 simulation (not a platform)?

And then I'll add checks to check if an attribute is actually set before you use it?

devlib/platform/gem5.py
avanlaer: Fixed

devlib/platform/gem5.py
avanlaer: Fixed

devlib/target.py
avanlaer: Fixed

devlib/target.py
avanlaer: Fixed

devlib/target.py
setrofim: platform will not be passed as an argument into instantiation of any connection, not just gem5. So all existing connection implementations must be updated to (optionally) expect it.

avanlaer: Yes, but isn't that the same for other arguments as well? Like for example, device in AdbConnection? (Sorry, not trying to be annoying, just want to learn)

setrofim: Correct. However, they're are already expecting the other arguments. this is a new argument being added. So it needs to be added to all connection implementations.

avanlaer: Fixed now :)

devlib/platform/arm.py
```diff
@@ -13,9 +13,9 @@
...
 #
 from __future__ import division
+import csv
```
bjackman: Just looking over these commits and noticed there's some residual diff-noise in this file.

devlib/target.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
```diff
@@ -34,14 +35,14 @@
 
 from devlib.exception import HostError, TargetError, TimeoutError
 from devlib.utils.misc import which, strip_bash_colors, escape_single_quotes, check_output
-
+from devlib.utils.types import boolean
```
avanlaer: Fixed

avanlaer:
I've updated the code and pushed the commits again.

devlib/utils/ssh.py
```diff
@@ -299,6 +306,454 @@ def __init__(self,
...
+                                   "differs from the one given {}!"
+                                   .format(host_system, host))
+        if username is not None and username != 'root':
+            raise ValueError('User should be root in gem5!')
+        if password is not None and password != '':
+            raise ValueError('No password needed in gem5!')
+        self.username = 'root'
+        self.is_rooted = True
+        self.password = None
+        self.port = -1
```
setrofim: Why `-1` rather than `None`?

devlib/utils/ssh.py
```diff
@@ -299,6 +306,454 @@ def __init__(self,
...
+    def push(self, source, dest, timeout=None):
+        """
+        Push a file to the gem5 device using VirtIO
+
+        The file to push to the device is copied to the temporary directory on
+        the host, before being copied within the simulation to the destination.
+        Checks, in the form of 'ls' with error code checking, are performed to
+        ensure that the file is copied to the destination.
+        """
+        filename = os.path.basename(source)
```
setrofim: There should be a call to `self._check_ready()` here

devlib/utils/ssh.py
```diff
@@ -299,6 +306,454 @@ def __init__(self,
...
+    def pull(self, source, dest, timeout=0): #pylint: disable=unused-argument
+        """
+        Pull a file from the gem5 device using m5 writefile
+
+        The file is copied to the local directory within the guest as the m5
+        writefile command assumes that the file is local. The file is then
+        written out to the host system using writefile, prior to being moved to
+        the destination on the host.
+        """
+        filename = os.path.basename(source)
```
setrofim: There should be a call to `self._check_ready()` here

devlib/utils/ssh.py
```diff
@@ -299,6 +306,454 @@ def __init__(self,
...
+        self._check_ready()
+        output = self._gem5_shell(command, as_root=as_root)
+        if strip_colors:
+            output = strip_bash_colors(output)
+        return output
+
+    def background(self, command, stdout=subprocess.PIPE,
+                   stderr=subprocess.PIPE, as_root=False):
+        raise TargetError("Attempting to execute {} in background."
+                          " This is not supported in gem5".format(command))
```
setrofim: Why? I understand that multiple parallel connections to a gem5 simulation are not supported. However it should be possible for the user to run something in the background on gem5, while doing something else on the host.

setrofim: Actually, now that I think about it. There should be a locking mechanism for a gem5 connection, and this error should be raised on failing to lock.

devlib/utils/ssh.py
```diff
@@ -299,6 +306,454 @@ def __init__(self,
...
+
+        gem5_logger.debug("gem5_shell command: {}".format(command))
+
+        # Send the actual command
+        self.conn.send("{}\n".format(command))
+
+        # Wait for the response. We just sit here and wait for the prompt to
+        # appear, as gem5 might take a long time to provide the output. This
+        # avoids timeout issues.
+        command_index = -1
```
setrofim: Using `None` would be more pythonic

devlib/target.py
```diff
@@ -153,6 +153,15 @@ def __init__(self,
                  ):
         self.connection_settings = connection_settings or {}
         self.platform = platform or Platform()
+        if 'platform' in self.connection_settings:
+            if connection_settings['platform'] is not self.platform:
```
setrofim: There is an edge case where this fails. If the user instantiates like so:

    target = LinuxTarget(connection_settings={'platform': Platform()})

This will result in an error, because if a platform is not passed to the Target by the user, a default one will be automatically created (line 155); so if one was passed in the connection settings, there would still be two target objects.

line 155 should be updated to

    self.platform = connection_settings.get('platform', Platform())

so that the default platform is only created if one has not been passed to the connection settings.

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: I think it has to be -1 because that is what the find() function returns if it cannot find the substring in the string (on line 631)

devlib/utils/ssh.py
avanlaer: Fixed

devlib/target.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

devlib/utils/ssh.py
avanlaer: Fixed

avanlaer:
I've updated the commits and added a new commit with documentation

devlib/platform/gem5.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        self.logger("This functionality is not yet implemented")
+        raise TargetError()
+
+    def _intercept_telnet_port(self):
+        """
+        Intercept the telnet port of a running gem5 simulation
+        """
+
+        assert self.gem5 is not None
```
marcbonnici: Assert statement left in.

devlib/target.py
```diff
@@ -152,7 +152,20 @@ def __init__(self,
                  conn_cls=None,
                  ):
         self.connection_settings = connection_settings or {}
-        self.platform = platform or Platform()
+        # Set self.platform either to platform given in connection_settings,
```
marcbonnici: Is part of this comment missing?

doc/platform.rst
```diff
@@ -0,0 +1,95 @@
...
+
+.. method:: Platform.update_from_target([target])
+
+    This method provides specific setup procedures for a gem5 simulation. First of all, the m5
+    binary will be installed on the guest (if it is not present). Secondly, three methods
+    in the :class:`Target` will be monkey-patched:
+
+            - **reboot**: this is not supported in gem5
+            - **reset**: this is not supported in gem5
+            - **capture_screen**: gem5 might already have screencaps so the
```
marcbonnici: `capture_screen` doesn't seem to display properly when I build the documentation.

devlib/target.py
```diff
@@ -152,7 +152,20 @@ def __init__(self,
...
+        if platform is None:
+            self.platform = self.connection_settings('platform', Platform())
```
marcbonnici: This should be `self.platform = self.connection_settings.get('platform', Platform())`

avanlaer:
I've updated the commits based upon Marc's comments

marcbonnici:
Would you be able to rebase your commits on top of master so that it can be merged?

avanlaer:
I have rebased the commits on the master branch.

PR #84, opened 2017-01-12
Some fixes for running commands into cgroups

bjackman:


devlib/module/cgroups.py
```diff
@@ -393,7 +393,10 @@ def run_into(self, cgroup, cmdline):
         :returns: Output of command.
         """
         cmd = self.run_into_cmd(cgroup, cmdline)
-        return self.target.execute(cmd)
+        raw_output = self.target.execute(cmd)
+
+        # First line of output comes from shutils; strip it out.
+        return '\n'.join(raw_output.splitlines()[1:])
```
setrofim: This unnecessarily splits up the entire (potentially long) output of the command and then joins it back up. Why not just

    return raw_output.split('\n', 1)[1]

bjackman: Good idea, will amend.

bjackman:
Pushed a version with your suggestion.

Issue #85, opened 2017-01-23
ADB now seems to return with an error if the target command failed


bjackman:
With my current ADB if I run `exit 1` I find that my ADB client process (on my host machine) returns 1. I'm 99% sure this didn't use to be the behaviour, and sure enough `adb_shell` in `devlib/utils/android.py` assumes that the ADB command returns 0, explicitly checking the exit code on the target to return a `TargetError` when necessary.

So this means I'm getting `CalledProcessError`s instead of `TargetError`s.

This is the version I had in my Android SDK install:
```
$ adb version
Android Debug Bridge version 1.0.36
Revision 302830efc153-android
```
I updated the SDK and get the same result with
```
$ adb version
Android Debug Bridge version 1.0.36
Revision 0e9850346394-android
```
I also tried a version of ADB from the Android sources with the same result:
```
$ out/host/linux-x86/bin/adb version
Android Debug Bridge version 1.0.36
Revision dfca6cdfd32b-android
```

Can anyone confirm that this behaviour has changed or that I am being a dunce? I couldn't find any of the commit hashes above in the ADB repo in the Android source tree (it seems to be maintained by an evil robot) so I haven't read the code.

Edit: Umm, [a comment here seems to suggest that the behaviour has indeed changed](https://code.google.com/p/android/issues/detail?id=3254).

Edit2: I'm using a Google Pixel target (Android 7.1), just realised that probably makes a difference.

bjackman:
Had some time to look into this properly and the ADB behaviour has really changed. However I misunderstood the devlib code, the ADB change isn't the source of my error.

Will raise a separate Issue or submit a PR once I've wrapped my head around the `adb_shell`/`check_output` stuff again.

bjackman:
I actually mis-misunderstood :confounded: and this is indeed the source of my problem. Submitting PR shortly.

PR #86, opened 2017-01-25
android: Handle variation in ADB behaviour

bjackman:
On older combinations of ADB/Android versions, the adb host command always
exits with 0 if it was able to run the command on the target, even if the
command failed (https://code.google.com/p/android/issues/detail?id=3254).

When we need the target's exit code (check_exit_code=True), we currently work
around this behaviour by echoing the target's exit code after the command and
parsing it out of the output.

The ADB behaviour is now "fixed" on newer versions with newer Androids (It's not
clear which versions these are and it appears that different builds of ADB with
the same version number differ in this respect). For those version combinations
adb_shell will currently raise a CalledProcessError when check_exit_code=False
and the target command fails.

So lets now use the "echo $?" trick whether or not we need the exit code.

Fixes https://github.com/ARM-software/devlib/issues/85

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -335,29 +335,33 @@ def adb_shell(device, command, timeout=None, check_exit_code=False,
     if as_root:
         command = 'echo \'{}\' | su'.format(escape_single_quotes(command))
     device_part = ['-s', device] if device else []
-    device_string = ' {} {}'.format(*device_part) if device_part else ''
-    full_command = 'adb{} shell "{}"'.format(device_string,
-                                             escape_double_quotes(command))
-    logger.debug(full_command)
-    if check_exit_code:
-        adb_shell_command = '({}); echo \"\n$?\"'.format(command)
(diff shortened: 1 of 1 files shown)


bjackman:
As per out-of-band discussion, pushed a new version where the command logged doesn't include the "echo $?" bit, as it's easier to scan the logs to see what's going on that way.

Note that I've slightly changed the format of the exception message. The old code uses `full_command` which is the ADB command minus the "$?" bit. The code I've just pushed uses `command` which is just the command the target saw. 

Not sure if it's better to raise an exception with the target command or the full ADB command, but I don't think in that context we should use something "in between" (i.e. with the "echo $?" removed) - IMO it should really be something that the user can copy-paste into a terminal with confidence that they're doing the same thing that devlib did.

setrofim:
I think for the exception, it makes sense to report just the command that executed on the Target -- because that is what caused the exception that resulted in the error code, not adb itself.

PR #89, opened 2017-02-16
Android support updates

derkling:
Adding few APIs which could be useful to have in devlib

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1084,6 +1084,31 @@ class AndroidTarget(Target):
             message = 'Could not find mount point for executables directory {}'
             raise TargetError(message.format(self.executables_directory))
 
+    _charging_enabled_path = '/sys/class/power_supply/battery/charging_enabled'
+
+    @property
+    def charging_enabled(self):
+        """
+        Whether drawing power to charge the battery is enabled
+
+        Not all devices have the ability to enable/disable battery charging
+        (e.g. because they don't have a battery). In that case,
+        ``charging_enabled`` is None.
+        """
+        if not self.file_exists(self._charging_enabled_path):
+            return None
+        return self.read_bool(self._charging_enabled_path)
+
+    @charging_enabled.setter
+    def charging_enabled(self, enabled):
+        """
+        Enable/disable drawing power to charge the battery
+
+        Not all devices have this facility. In that case, do nothing.
+        """
+        if not self.file_exists(self._charging_enabled_path):
+            return
+        self.write_value(self._charging_enabled_path, int(bool(enabled)))
 
 FstabEntry = namedtuple('FstabEntry', ['device', 'mount_point', 'fs_type', 'options', 'dump_freq', 'pass_num'])
 PsEntry = namedtuple('PsEntry', 'user pid ppid vsize rss wchan pc state name')
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -62,10 +62,11 @@ class Target(object):
         return self.conn is not None
 
     @property
-    @memoized
     def connected_as_root(self):
-        result = self.execute('id')
-        return 'uid=0(' in result
+        if self._connected_as_root is None:
+            result = self.execute('id')
+            self._connected_as_root = 'uid=0(' in result
+        return self._connected_as_root
 
     @property
     @memoized
@@ -151,6 +152,7 @@ class Target(object):
                  shell_prompt=DEFAULT_SHELL_PROMPT,
                  conn_cls=None,
                  ):
+        self._connected_as_root = None
         self.connection_settings = connection_settings or {}
         # Set self.platform: either it's given directly (by platform argument)
         # or it's given in the connection_settings argument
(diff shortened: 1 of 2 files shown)


devlib/target.py
```diff
@@ -1041,6 +1041,12 @@ def adb_reboot_bootloader(self):
     def adb_root(self):
         adb_command(self.adb_name, 'root', timeout=30)
 
+    def usb_charge(self, enable=True, sysfs_path=None):
```
setrofim: Why is the path being passed as argument? Given that it is in sysfs, I would expect it to be standardized.

Even if it can vary, it is not something I would expect the calling code to be aware of. If there are several common possibilities (e.g. varying between kernel versions etc), they should all be tested in turn, starting with the most likely. If the path can vary widely between different devices, it should be added as a configurable option in the `Platform` (which encapsulates device-specific knowledge).

derkling: Nor really sure how much standard is that path... but AFAIK is the only one we found so far.
Should we keep it hardcoded in the method, for the time being, and refactor it better once we should find other possible options?

setrofim: Yes, I think hard-coding it for now is the best course of action. Given that there is no evidence that the path may vary at this time, it is better to keep things simple.

bjackman: There's an alternative version here: https://github.com/bjackman/devlib/commit/96d0e610441410d40f223f31836f24df4e8ff74d

devlib/target.py
```diff
@@ -1035,6 +1035,18 @@ def dump_logcat(self, filepath, filter=None, append=False, timeout=30):  # pylin
     def clear_logcat(self):
         adb_command(self.adb_name, 'logcat -c', timeout=30)
 
+    def adb_reboot_bootloader(self):
+        adb_command(self.adb_name, 'reboot-bootloader', timeout=30)
```
bjackman: Might be useful to have `timeout` as a parameter to `adb_reboot_bootloader`

devlib/target.py
derkling: I rather prefer the (just discovered) patch by @bjackman ... going to cherry-pick that and refresh this PR ;-)

devlib/target.py
```diff
@@ -1035,6 +1035,18 @@ def dump_logcat(self, filepath, filter=None, append=False, timeout=30):  # pylin
...
+
+    def adb_root(self):
+        adb_command(self.adb_name, 'root', timeout=30)
```
bjackman: Should probably update `self.connected_as_root` in this case.

devlib/target.py
setrofim: Sounds good. Though the setter should be part of the property, see: http://stackoverflow.com/questions/6618002/python-property-versus-getters-and-setters

devlib/target.py
derkling: I wanted to keep the API consistent with what we already have, `clear_logcat` does not expose such a timeout... but probably it makes more sense for the reboot command to expose a timeout.

Ok, let's have it as a parameter.

setrofim: Erm, do we expect this to vary that much between devices? I would have thought booting into bootloader should never exceed 30 seconds.

devlib/target.py
derkling: That seems interesting... however, I quite don't get the potential side-effect on:
https://github.com/ARM-software/devlib/blob/master/devlib/target.py#L739

@setrofim so you think is it safe to update `self.connected_as_root` when this command succeeds?

devlib/target.py
derkling: That's quite like... but if we keep:

```
def adb_reboot_bootloader(self, timeout=30):
```

that should five more flexibility in case that's required.

Or do you prefer to avoid exposing potential "platform dependencies"?

devlib/target.py
setrofim: That _resolve_paths only happens for `LinuxTarget`; `AndroidTarget` equivalent is here:
https://github.com/ARM-software/devlib/blob/master/devlib/target.py#L1050

So yes, should be safe (and also necessary -- as the rooted commands would still be passed via 'su' otherwise -- great catch by @bjackman)

devlib/target.py
setrofim: Actually, now that I think about it, exposing it may be the right thing to do. My thinking was -- given that this is bootloader, rather than a full system boot, it's hard to imagine any system where booting would take longer. However, this time needs to account for the system shutdown before the reboot as well. While still unlikely to be over 30 seconds, it's less unreasonable to expect that there is a platform out there where that might be the case. So yes, adding it as a parameter with a default makes sense.

devlib/target.py
derkling: Ok, I've got something working... let's refresh the PR and see if what I'm proposing is sound... otherwise I still have my :parasol_on_ground: to shield from :tomato: s

devlib/target.py
```diff
@@ -62,10 +62,11 @@ def is_connected(self):
         return self.conn is not None
 
     @property
-    @memoized
```
derkling: I guess this has to go if we want to switch the property values using the adb_root(enable) call.
Unless there is some API I'm not aware of to update the cached values of a memoized...

Moreover, the setattr cannot be used in this property because the way we set it depends on Linux/AndroidTarget implementation... unless I can defined different connected_as_root.setattr for each derived class. :-?

setrofim: There should not be a setter -- as far as the user code is concerned, this property is read-only. Internally, the `self._connected_as_root` value should be updated directly, rather than via the property.

PR #95, opened 2017-03-01
android: Improve error when _setup_ls fails

bjackman:
If the ADB command fails (e.g. if you provide the wrong device ID), adb_command
raises a CalledProcessError. CalledProcessError doesn't print the output of the
failed command, so you get a useless error message.

This is relevent here in particular as _setup_ls is the first thing to run an
ADB command if a non-IP device ID is provided.

Catch the CalledProcessError and instead raise a HostError with the command
output in the exception message.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -177,7 +177,12 @@ class AdbConnection(object):
     # versions of the ls tool in Android pre-v7.
     def _setup_ls(self):
         command = "shell '(ls -1); echo \"\n$?\"'"
-        output = adb_command(self.device, command, timeout=self.timeout)
+        try:
+            output = adb_command(self.device, command, timeout=self.timeout)
+        except subprocess.CalledProcessError as e:
+            raise HostError(
+                'Failed to set up ls command on Android device. Output:\n'
+                + e.output)
         lines = output.splitlines()
         retval = lines[-1].strip()
         if int(retval) == 0:


devlib/utils/android.py
```diff
@@ -177,7 +177,12 @@ def newline_separator(self):
     # versions of the ls tool in Android pre-v7.
     def _setup_ls(self):
         command = "shell '(ls -1); echo \"\n$?\"'"
-        output = adb_command(self.device, command, timeout=self.timeout)
+        try:
+            output = adb_command(self.device, command, timeout=self.timeout)
+        except subprocess.CalledProcessError as e:
+            raise HostError(
```
setrofim: Should be `TargetError` rather than `HostError`. The former is for errors that occur on target or with communication with the target; the latter is for issues with the host itself (missing dependencies, non-existing on-host paths etc).

bjackman: OK. Did think about this but was unsure. `adb_get_device` raises a `HostError` if it can't communicate with a device. I guess that's because the implication is that there *is no* device in that case?

setrofim: Hm, good point. On second though, you're right. Given that this is the first thing that happens for non-IP devices, the assumptions are essentially the same as in `adb_get_device` case. Hm may be there should be a "ConnectionError" or something in both cases. Anyway, for now, consistency with `adb_get_device` makes sense. I withdraw my previous comment.

PR #100, opened 2017-03-17
devlib: add monsoon instrument

joelagnel:
Signed-off-by: Joel Fernandes <joelaf@google.com>

### DIFF
--- a/devlib/instrument/monsoon.py
+++ b/devlib/instrument/monsoon.py
@@ -1,0 +1,170 @@
+import devlib
+import json
+import os
+import os.path
+import psutil
+import time
+import logging
+from subprocess import Popen, STDOUT
+from threading import Timer
+import pandas as pd
+import StringIO
+from collections import namedtuple
+from subprocess import Popen, PIPE, STDOUT
+from devlib.instrument import Instrument, CONTINUOUS, MeasurementsCsv
+from devlib.exception import HostError
+from devlib.utils.misc import which
+
+class MonsoonInstrument(Instrument):
+    MONSOON_BIN = "./tools/scripts/monsoon.par"
+
+    def log_error(self, errmsg, exception=0):
+	print "ERROR: " + errmsg
+	self.logger.error(errmsg)
+	if exception == 1:
+		raise(Exception(errmsg))
+
+    def log_info(self, msg):
+	print "INFO: " + msg
+	self.logger.info(msg)
+
+    def __init__(self, target, device_entry=None):
+	super(MonsoonInstrument, self).__init__(target)
+
+	if not os.path.exists(MonsoonInstrument.MONSOON_BIN):
+		self.log_error("monsoon.par not found", exception=1)
+
+	if device_entry != None and not os.path.exists(device_entry):
+		self.log_error(str(device_entry) + " doesn't exist", exception=1)
+
+	self.device_entry = device_entry
+	self.process = None
+
+	self.log_info('Found monsoon.par')
+
+    def reset(self):
+        self.log_info('Initiailzing MonsoonInstrument')
+	# Waiting for monsoon to get ready
+	self.run_monsoon(['--status'], timeout=3)
+	self.log_info("MonsoonInstrument ready for action")
+
+    def kill_monsoon(self, p):
+	try:
+		p.kill()
+	except OSError:
+		pass # ignore
+
+    # Args should be a list of arguments ex, ['--status']
+    # In !stream mode, we continously kill monsoon till it behaves (that is it doesn't block)
+    # This is for the initial --status command and other commands where we're not streaming
+    # but just running a simple monsoon command and returning.
+    # For the streaming case (where we are confident monsoon is working after passing it --status),
+    # we kill monsoon after timeout but this isn't an error, just termination of data collection.
+    def run_monsoon(self, args=None, stream=0, async_mode=0, timeout=5, report=0):
+	if report == 0 and args == None:
+		raise(Exception("Can't run monsoon without args"))
+
+	if report == 0:
+		popen_args = [MonsoonInstrument.MONSOON_BIN]
+		if self.device_entry:
+			popen_args += ['--device', self.device_entry]
+		popen_args += args
+
+	# Give up after 20 trials for !stream case (every attempt is timeout seconds)
+	n_trys_max = 20
+	n_trys = n_trys_max
+
+	while n_trys > 0:
+		# When reporting, don't try to run monsoon
+		if report == 0:
+			msg = "Running monsoon command: " + str(popen_args)
+			self.log_info(msg)
+			p = Popen(popen_args, stdout=PIPE, stderr=PIPE)
+
+		if async_mode == 1:
+			self.process = p
+			return "ASYNC"
+
+		if report == 1:
+			stream = 1
+			p = self.process
+			if p == None:
+				self.log_error("Monsoon not running") 
+			p.kill()
+		else:
+			# In sync mode (async_mode = 0), we kill monsoon after sometime
+			t = Timer(timeout, self.kill_monsoon, [p])
+			t.start()
+
+		# Get return code for task after either it quit or it was killed
+		waitret = p.wait()
+		if waitret != 0:
+			if report == 0:
+				t.cancel()
+
+			# If we received an error and we're !stream, do a retry
+			if stream == 0:
+				n_trys = n_trys - 1
+				errmsg="timed out waiting or error for monsoon, try resetting it"
+				self.log_error(errmsg)
+
+			errmsg = p.stderr.read()
+			stderrmsg = ""
+			for l in errmsg.split(os.linesep):
+				if l == "":
+					continue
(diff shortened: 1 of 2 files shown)


setrofim:
This does not implement the `Instrument` API as described here:

http://pythonhosted.org/devlib/instrumentation.html

- The `mode` must be set appropriately
- The measurements this instrument is capable of collecting must be advertised via channels
-  Depending on the `mode`, `take_measurment` and/or all of `start/stop/get_data` must be implemented as described in the docs.

Also please could you provide more information on what "monsoon" is. How does one get a hold of it, why would someone want to use it, etc?

devlib/instrument/monsoon.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def log_info(self, msg):
+	print "INFO: " + msg
+	self.logger.info(msg)
+
+    def __init__(self, target, device_entry=None):
+	super(MonsoonInstrument, self).__init__(target)
+
+	if not os.path.exists(MonsoonInstrument.MONSOON_BIN):
+		self.log_error("monsoon.par not found", exception=1)
```
derkling: Here, if possible, we should report some tip on how to get this binary.
This is what we did in LISA for the ACME cape:
   https://github.com/ARM-software/lisa/blob/master/libs/utils/energy.py#L344
@setrofim: can this work for devlib as well?

devlib/instrument/monsoon.py
```diff
@@ -0,0 +1,170 @@
...
+	print "ERROR: " + errmsg
+	self.logger.error(errmsg)
+	if exception == 1:
+		raise(Exception(errmsg))
+
+    def log_info(self, msg):
+	print "INFO: " + msg
+	self.logger.info(msg)
+
+    def __init__(self, target, device_entry=None):
```
derkling: What about `s/device_entry/device/`?
This `_entry` suffix seems not necessary...

devlib/instrument/monsoon.py
```diff
@@ -0,0 +1,170 @@
...
+	if not os.path.exists(MonsoonInstrument.MONSOON_BIN):
+		self.log_error("monsoon.par not found", exception=1)
+
+	if device_entry != None and not os.path.exists(device_entry):
+		self.log_error(str(device_entry) + " doesn't exist", exception=1)
+
+	self.device_entry = device_entry
+	self.process = None
+
+	self.log_info('Found monsoon.par')
```
derkling: Do not report "it's working" type of comments, let's log only useful information and/or warnings/errors.

In devlib modules are "as much quite as possible". ;-)

devlib/instrument/monsoon.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def reset(self):
+        self.log_info('Initiailzing MonsoonInstrument')
```
derkling: Some as above, not useful information to be reported here.

devlib/instrument/monsoon.py
```diff
@@ -0,0 +1,170 @@
...
+	# Waiting for monsoon to get ready
+	self.run_monsoon(['--status'], timeout=3)
+	self.log_info("MonsoonInstrument ready for action")
+
+    def kill_monsoon(self, p):
```
derkling: For the Monsoon device we can probably support (at least initially) only the CONTINUOUS interface, thus you need to implement only 4 public methods: `reset`, `start`, `stop` and `get_data`.

Here is an example:
https://github.com/ARM-software/devlib/blob/master/devlib/instrument/energy_probe.py#L35

devlib/instrument/monsoon.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def set_current(self, current):
+	return self.run_monsoon(['--current', str(current)])
+
+    def set_start_current(self, current):
+	return self.run_monsoon(['--startcurrent', str(current)])
+
+    # Returns pandas dataframe for monsoon output (time and current in mA)
+    def get_samples_sync(self, duration):
+	txt = self.run_monsoon(['--hz', '5', '--timestamp', '--samples', '2000000'], stream=1, timeout=duration)
```
derkling: These params should probably be exported somehow. Perhaps, since we have a single channel, we can specify them in the `__init__`?

In that case you can use them to configure the single InstrumentChannel which is going to be registered, as part of the instrument specific `attrs`:
https://github.com/ARM-software/devlib/blob/master/devlib/instrument/__init__.py#L142

derkling:
@setrofim Monsoon is this *single channel* energy meter:
   https://www.msoon.com/LabEquipment/PowerMonitor/
which is quite used within Google for all energy measurements related to the Android platform.

It is thus quite important to have support for this device within devlib and LISA.

bjackman:
I have a simplified version of this functionality here: https://github.com/bjackman/devlib/commit/ea5ab35e21757fa7e3bc4bcad60e0cea0a967c52

It seems to be working pretty well, not ready for merge though because:

- I have only seen nonsense data so far (I don't think it's due to issues with my code or Monsoon, but let's be sure first!)

- The Monsoon meter is only officially supported using Windows. Google have a Python tool ("monsoon.par") to communicate with it from Unix, and this code depends on it. Hopefully we can distribute it in Devlib but we need to figure out if there are issues with licensing, Google IP, or Monsoon IP.

setrofim:
> Hopefully we can distribute it in Devlib but we need to figure out if there are issues with licensing, Google IP, or Monsoon IP.

The path to the tool can be passed as an argument on `__init__` , rather than be hard-coded in a class attribute in the mean time. This means the tool doesn't have to be included with devlib, and it would be up to the user to obtain it and point devlib to it.

This would make the instrument more awkward to use, so including the binary and setting the path in a class attribute would be preferable, but the above approach can be used to unblock the merging of the instrument in the mean time. 

If/when the IP issues are resolved, and the tool can be included in devlib, the attribute can remain, and simply be made optional, defaulting to the internal version. That way backward compatibility with existing setups will be maintained.

bjackman:
Some notes on this PR following from out-of-band chat with @joelagnel:

- As @setrofim mentioned, need to get this compliant with the Instrument API, the branch I linked [above](https://github.com/ARM-software/devlib/pull/100#issuecomment-294791285) aims to demo the bare minimum for that.

- This returns a pandas DataFrame. However devlib doesn't currently depend on pandas. Instead, we should return CSV/`MeasurementsCsv` and then convert that to a `DataFrame` in a LISA wrapper. The current LISA energy meter classes don't do this but it's easy to add (I have a private branch that I should push at some point).

- Will defer to @setrofim on this but I don't think the voltage & current stuff belongs in this class (which is for measuring), maybe it should be a `Module`, separate from the `Instrument`? Then I think the desired voltage and current could be parameters of the `Platform`? Not too sure but it's not a big deal.

- I haven't yet been convinced of the need for loop in `run_monsoon`, when testing my version I never actually came across Monsoon flakiness when using it via devlib. 

- @joelagnel pointed out that Google's monsoon script is already [public](https://android.googlesource.com/platform/cts/+/master/tools/utils/monsoon.py) :grin: It has dependencies so I think the best thing is to just require users to install it themselves, like we do for caiman in energy_probe.py - @setrofim do you agree?

setrofim:
> Will defer to @setrofim on this but I don't think set_voltage & set_current belongs in this class (which is for measuring), maybe it should be a Module, separate from the Instrument? Then I think the desired voltage and current could be parameters of the Platform? Not too sure but it's not a big deal.

Form what I understood from the code, these set voltage/current in the monsoon device, not on the target. If that is the case, then the instrument is the right place for this, as they relate to monsoon device, rather than the target (even if they need to be adjusted to different values for different targets).

>@joelagnel pointed out that Google's monsoon script is already public :grin: It has dependencies so I think the best cost/benefit is to just require users to install it themselves, like we do for caiman in energy_probe.py - @setrofim do you agree?

Agreed. It would be good to link to the the script in the documentation for this instrument.

bjackman:
> Form what I understood from the code, these set voltage/current in the monsoon device, not on the target. If that is the case, then the instrument is the right place for this, as they relate to monsoon device, rather than the target (even if they need to be adjusted to different values for different targets).

Actually Monsoon is a power supply as well as a measurement tool, so these really do set the voltage and (max?) current on the device. It's designed for testing mobile device prototypes by acting as a "fake battery".

setrofim:
> Actually Monsoon is a power supply as well as a measurement tool, so these really do set the voltage and (max?) current on the device. It's designed for testing mobile device prototypes by acting as a "fake battery".

Right, so these are the configurations of the Monsoon power supply, not of the target. So the `set_voltage/set_current` belong in the instrument. You could argue that the values for the voltage/current are properties of a particular device (what it expects to be supplied with), and so belong in the Platform, wherefrom they would be take by the instrument. However, given that this currently the only place where is this need, and that it's not something can actually be read from the target, I would suggest keeping them as the instrument parameters for now; If, in the future, we add support for some other power supply-type device to devlib that required the same configuration, then can move the setting into Platform.

bjackman:
OK, sounds good.

setrofim:
Superseded by https://github.com/ARM-software/devlib/pull/119

PR #107, opened 2017-04-10
cpufreq: Expose error when writing invalid governor tunable

bjackman:
If we get an TargetError when trying to set a governor tunable we currently fall
back to an older sysfs layout under the assumption that the file we tried to
write doesn't exist. However it may be that the file exists, but we tried to
write an invalid value (or something else went wrong). In this case we fall back
to the old file location, fail, and produce a nonsense error about the old path
not existing.

Instead, when we get a TargetError, check if the file exists, and fall back to
the old location only if it does not.

### DIFF
--- a/devlib/module/cpufreq.py
+++ b/devlib/module/cpufreq.py
@@ -152,10 +152,14 @@ class CpufreqModule(Module):
         valid_tunables = self.list_governor_tunables(cpu)
         for tunable, value in kwargs.iteritems():
             if tunable in valid_tunables:
+                path = '/sys/devices/system/cpu/{}/cpufreq/{}/{}'.format(cpu, governor, tunable)
                 try:
-                    path = '/sys/devices/system/cpu/{}/cpufreq/{}/{}'.format(cpu, governor, tunable)
                     self.target.write_value(path, value)
-                except TargetError:  # May be an older kernel
+                except TargetError:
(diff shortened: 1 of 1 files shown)


devlib/module/cpufreq.py
```diff
@@ -152,10 +152,14 @@ def set_governor_tunables(self, cpu, governor=None, **kwargs):
         valid_tunables = self.list_governor_tunables(cpu)
         for tunable, value in kwargs.iteritems():
             if tunable in valid_tunables:
+                path = '/sys/devices/system/cpu/{}/cpufreq/{}/{}'.format(cpu, governor, tunable)
                 try:
-                    path = '/sys/devices/system/cpu/{}/cpufreq/{}/{}'.format(cpu, governor, tunable)
                     self.target.write_value(path, value)
-                except TargetError:  # May be an older kernel
+                except TargetError:
+                    if te.target.file_exists(path):
```
setrofim: Should be `self.target.file_exists`?

bjackman: Eh yeah. Thought I tested this (I saw the error I wanted to see) so may have misunderstood the whole thing, will have to rewrite.

bjackman:
OK I didn't misunderstand anything, I must have just failed to test the code. Fixed now.

PR #111, opened 2017-04-20
shutils: Speed up cpuidle_wake_all_cpus

bjackman:
This can currently take several hundred milliseconds on slow targets with many
CPUs. To speed it up, use '&' to spawn the perturbations in paralell.


bjackman:
NB this

PR #114, opened 2017-04-25
ssh: Combine exit code command with main command

bjackman:
Reducing the number of commands executed speeds up operation on
SSH-conected targets.

### DIFF
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -185,16 +185,22 @@ class SshConnection(object):
                 as_root=False, strip_colors=True): #pylint: disable=unused-argument
         try:
             with self.lock:
-                output = self._execute_and_wait_for_prompt(command, timeout, as_root, strip_colors)
+                _command = '({}); __devlib_ec=$?; echo; echo $__devlib_ec'.format(command)
+                raw_output = self._execute_and_wait_for_prompt(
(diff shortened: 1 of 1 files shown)


bjackman:
I've only tested this against ARM Juno running Debian...

setrofim:
> Does splitting on '\n' always work? I think it does

I think it should too. One edge case that comes to mind are commands that do not return any output (not event a new line). Have you tried this with that case?

> Does the target always have `printf`? I think it's required by POSIX so should be fine (or maybe we should use busybox?)

Use busybox. This is typically the safe choice for command for which it has applets. Even if a command is present on a target, its behavior may vary (e.g. BSD vs GNU version of binaries); always using the busybox applet ensures consistent results.

> Is this actually faster? I tested it by using the hotplug API, it reduced the time to hotplug a CPU from around 2.2s to 1.8s. I think parsing a printf is always going to be faster than sending a line over SSH.

Yes, this will definitely be faster, as you're saving on a network round trip. Depending on the connection speed, this may add up to a significant saving over a course of a long run.

bjackman:
Cool, I'll make it use busybox and double check with a command that produces 0 output. Thanks for review as always.

bjackman:
Putting this on hold until I can figure out how to make it work for all inputs. Currently doesn't work properly for `target.execute("printf '\\r'")`

PR #116, opened 2017-04-26
Add sample_rate_hz attribute to CONTINUOUS instruments

bjackman:
Don't have a DAQ so couldn't test that. Also couldn't get the Juno readenergy.c instrument working on my board. Decided that this can't really break anything, though.

### DIFF
--- a/devlib/instrument/daq.py
+++ b/devlib/instrument/daq.py
@@ -27,7 +27,7 @@ class DaqInstrument(Instrument):
                  device_id='Dev1',
                  v_range=2.5,
                  dv_range=0.2,
-                 sampling_rate=10000,
+                 sample_rate_hz=10000,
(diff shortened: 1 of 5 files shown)


devlib/instrument/daq.py
```diff
@@ -55,6 +55,7 @@ def __init__(self, target, resistor_values,  # pylint: disable=R0914
                                                  resistor_values=resistor_values,
                                                  channel_map=channel_map,
                                                  labels=labels)
+        self.sample_rate_hz = sampling_rate
```
setrofim: Probably worth renaming `sampling_rate` argument to `sample_rate_hz` for consistency.

bjackman: OK, I actually thought about this but then thought I should not break backwards compatibility. Don't know how many people use DAQs.

setrofim: We're still in version 0, so backward compatibility is not a hard guarantee yet. It's better to fix these things now, while we have the chance.

PR #117, opened 2017-04-28
android: Raise better error when trying to push non-existent file

bjackman:


devlib/utils/android.py
```diff
@@ -204,7 +205,12 @@ def push(self, source, dest, timeout=None):
         if timeout is None:
             timeout = self.timeout
         command = "push '{}' '{}'".format(source, dest)
-        return adb_command(self.device, command, timeout=timeout)
+        try:
+            return adb_command(self.device, command, timeout=timeout)
+        except CalledProcessError:
```
bjackman: Actually, `os.path.exists` is probably really fast, maybe we can just check before trying the ADB command - what do you think?

bjackman: Actually yeah, pushed a version that just checks every time.

bjackman:
Initially pushed a version that only called `os.path.exists` if we got a `CalledProcessError` (to avoid slowing down the successful case), but then realised the time taken by `os.path.exists` is almost certainly negligible compared to an `adb push`, so decided we can just check every time. Easier to read.

PR #119, opened 2017-05-03
instrument: Add support for Monsoon Power Monitor

bjackman:
This adds support for the Monsoon Power Monitor:

https://www.msoon.com/LabEquipment/PowerMonitor/

This device officialy supports only MS Windows so we use a tool from AOSP to
communicate with it from Unix targets:

https://android.googlesource.com/platform/cts/+/master/tools/utils/monsoon.py

The user is required to install this tool and its dependencies, then pass the
path to it as the monsoon_bin parameter.

### DIFF
--- a/devlib/instrument/monsoon.py
+++ b/devlib/instrument/monsoon.py
@@ -1,0 +1,132 @@
+import csv
+import os
+import signal
+from subprocess import Popen, PIPE
+from tempfile import NamedTemporaryFile
+from devlib.instrument import Instrument, CONTINUOUS, MeasurementsCsv
+from devlib.exception import HostError
+from devlib.host import PACKAGE_BIN_DIRECTORY
+from devlib.utils.misc import which
+
+INSTALL_INSTRUCTIONS="""
+MonsoonInstrument requires the monsoon.py tool, available from AOSP:
+
+https://android.googlesource.com/platform/cts/+/master/tools/utils/monsoon.py
+
+Download this script and put it in your $PATH (or pass it as the monsoon_bin
+parameter to MonsoonInstrument). `pip install gflags pyserial` to install the
+dependencies.
+"""
+
+class MonsoonInstrument(Instrument):
+    """Instrument for Monsoon Solutions power monitor
+
+    To use this instrument, you need to install the monsoon.py script available
+    from the Android Open Source Project. As of May 2017 this is under the CTS
+    repository:
+
+    https://android.googlesource.com/platform/cts/+/master/tools/utils/monsoon.py
+
+    Collects power measurements only, from a selection of two channels, the USB
+    passthrough channel and the main output channel.
+
+    :param target: Ignored
+    :param monsoon_bin: Path to monsoon.py executable. If not provided,
+                        ``$PATH`` is searched.
+    :param tty_device: TTY device to use to communicate with the Power
+                       Monitor. If not provided, a sane default is used.
+    """
+
+    mode = CONTINUOUS
+
+    def __init__(self, target, monsoon_bin=None, tty_device=None):
+        super(MonsoonInstrument, self).__init__(target)
+        self.monsoon_bin = monsoon_bin or which('monsoon.py')
+        if not self.monsoon_bin:
+            raise HostError(INSTALL_INSTRUCTIONS)
+
+        self.tty_device = tty_device
+
+        self.process = None
+        self.output = None
+
+        self.sample_rate_hz = 500
+        self.add_channel('output', 'power')
+        self.add_channel('USB', 'power')
+
+    def reset(self, sites=None, kinds=None, channels=None):
+        super(MonsoonInstrument, self).reset(sites, kinds)
+
+    def start(self):
+        if self.process:
+            self.process.kill()
+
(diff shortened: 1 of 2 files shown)


bjackman:
Not able to request reviewers but CC: @joelagnel and thanks for all the help on this.

devlib/instrument/monsoon.py
```diff
@@ -0,0 +1,116 @@
...
+    passthrough channel and the main output channel.
+
+    :param target: Ignored
+    :param monsoon_bin: Path to monsoon.py executable
+    :param device_entry: TTY device to use to communicate with the Power
+                         Monitor. If not provided, a sane default is used.
+    """
+    mode = CONTINUOUS
+
+    def __init__(self, target, monsoon_bin, device_entry=None):
```
setrofim: Perhaps `tty_device` instead of `device_entry`?

bjackman: Used `device_entry` because that's what `energy_probe.py` uses but agree `tty_device` is better, will rename.

joelagnel:
Seems you removed the monsoon_bin search? I'd like for the search to be automatic based on the old approach of scanning the devlib scripts directory and warning the user if it isn't found there with a message of where it needs to be added.
Also, do we want to supersede https://github.com/ARM-software/lisa/pull/335 too based on your version of it? Can you submit that PR as well? Thanks

setrofim:
> Seems you removed the monsoon_bin search? I'd like for the search to be automatic based on the old approach of scanning the devlib scripts directory and warning the user if it isn't found there with a message of where it needs to be added.

As mentioned in the previous PR, there are potential licensing issues with including the monsoon script directy in devlib. Besides which, asset tracking is something that should be kept out of the library to the extent that is possible. It is the responsibility of the application using devlib to provide them. devlib itself should impose as little structure as possible on the host. The application (e.g. LISA) can automatically download the script and place it where ever it makes sense on the host, passing the location to devlib on instantiation.

joelagnel:
Inclusion into devlib wasn't enforced (the previous discussion was to copy monsoon.py or .par manually to the scripts directory).
That's Ok with me (LISA downloading monsoon into a directory) but I strongly believe that this should be automated somewhere... Brendan, do you think we can have LISA download it from AOSP? thanks

setrofim:
> Inclusion into devlib wasn't enforced (the previous discussion was the user to copy monsoon.py or .par manually to the scripts directory).

This is messy, may be impractical, and is inviting deployment-specific issues. devlib may be installed via PyPI and the user should not have to know or care where it's source is. If installed by a system administrator the user may not even have permissions to modify the package directory structure. Conversely modify the clone source tree just increases the possibility of something being accidentally committed and pushed. In either case modifying Python packages in a local environment is prone to result in issues that are non-reproducible outside the environment (e.g. say the scripts directory moves, for whatever reason after an upgrade). A package directory structure should only contain files belonging to the package. So the assets must either be included as part of the source (and this should really only be limited to the "essentials"), or provided to devlib by the application that is using it. 

> but I strongly believe that this should be automated somewhere

Asset discover is, at the moment, outside the scope of devlib. Right now, the application would be the right place to do it. That said, I am open to a discussion regarding  expanding devlib  scope to provide an asset management mechanism. We have a resource discovery mechanism in WA, and I'm sure LISA has its own way of dealing with this as well. We could discuss aligning on a common approach for handing assets; though I'm just as happy to leave that in the remit of the application.

bjackman:
I don't think it's worth having devlib download monsoon.py, and unfortunately LISA doesn't have a framework for doing so. To be honest I think the one-time cost of manually downloading monsoon.py (having been given a link by LISA) is pretty bearable, especially since the number of users it affects is pretty slim.

bjackman:
> Also, do we want to supersede ARM-software/lisa#335 too based on your version of it? Can you submit that PR as well? 

Will do, it depends on https://github.com/ARM-software/lisa/pull/370 so I'll submit the PR as soon as that's merged. The branch is here if you want to take a look: https://github.com/bjackman/lisa/blob/monsoon/libs/utils/energy.py

joelagnel:
@bjackman Ok. Does this mean that the user of LISA has to provide the monsoon.py path in the emeter conf? I think that's a bit unacceptable.
I agree with you that it doesn't make sense to download from AOSP, but  in the case where monsoon_bin is None, can LISA or devlib atleast scan PATH for monsoon.py and then die if its not?
I think it complicates the conf if monsoon.py path has to be passed to the emeter everytime.

About slim number of monsoon users, there is a growing interest in LISA here and many people use monsoon so I would expect quite a few folks to be using this (specially since @derkling added emeter support for android workload.py). In this case, I wouldn't want them to have to pass monsoon.py path everytime. Cheers

bjackman:
> in the case where monsoon_bin is None, can LISA or devlib atleast scan PATH for monsoon.py and then die if its not?

Yeah, that's definitely a good point, I'll make it do that. Do you think that would be good enough? That's actually what the ARM Energy Probe (energy_probe.py) does and I've always found that to be pretty usable (at least from devlib's perspective...)

bjackman:
OK, pushed a new version that checks the $PATH. Also added installation instructions which I forgot to do before.

PR #124, opened 2017-05-17
Add iter_domains method to cpufreq module

bjackman:


devlib/module/cpufreq.py
```diff
@@ -414,10 +414,12 @@ def trace_frequencies(self):
...
+
+        Returned as a tuple of CPU IDs
         """
         if isinstance(cpu, int):
             cpu = 'cpu{}'.format(cpu)
 
         sysfile = '/sys/devices/system/cpu/{}/cpufreq/affected_cpus'.format(cpu)
 
-        return [int(c) for c in self.target.read_value(sysfile).split()]
+        return tuple(int(c) for c in self.target.read_value(sysfile).split())
```
setrofim: I don't think this should return a tuple. A tuple implies a fixed size and a fixed meaning for the value at each index. This is not the case here.

bjackman: Fair enough. I thought this was necessary for the following commit (because of the use of `set`) but I just realised that was a brainfart. Will just drop this commit.

bjackman: So do you also think `iter_domains` should also yield lists rather than tuples?

setrofim: yes, by the same logic.

bjackman:
Updated to drop the `tuple` stuff.

PR #127, opened 2017-05-22
target: add dry run for command execution

valschneider:
It sometimes can be useful to enable dry-run before calling
high-level functions in order to check what individual commands
are being called (or do whatever else with them).

This patch adds dry-run with an on/off switch:
once start_dryrun() is called, every subsequent command passed to
execute() will be accumulated in a list. stop_dryrun() returns that
list and disables dry-run.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -145,6 +145,10 @@ class Target(object):
         else:
             return None
 
+    @property
+    def dryrun(self):
+        return self._dryrun
+
     def __init__(self,
                  connection_settings=None,
                  platform=None,
@@ -186,6 +190,8 @@ class Target(object):
         self._cache = {}
         self._connections = {}
         self.busybox = None
+        self._dryrun_enabled = False
+        self._dryrun = []
 
         if load_default_modules:
             module_lists = [self.default_modules]
@@ -311,8 +317,20 @@ class Target(object):
 
     # execution
 
+    def start_dryrun(self):
+        if not self._dryrun_enabled:
+            self._dryrun_enabled = True
+            self._dryrun = []
+
+    def stop_dryrun(self):
+        if self._dryrun_enabled:
+            self._dryrun_enabled = False
+
     def execute(self, command, timeout=None, check_exit_code=True, as_root=False):
-        return self.conn.execute(command, timeout, check_exit_code, as_root)
+        if self._dryrun_enabled:
+            self._dryrun.append(command)
+        else:
+            return self.conn.execute(command, timeout, check_exit_code, as_root)
 
     def background(self, command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, as_root=False):
         return self.conn.background(command, stdout, stderr, as_root)
--- a/doc/target.rst
+++ b/doc/target.rst
@@ -145,6 +145,11 @@ Target
              thread. This will always be set to the connection for the current
              thread.
 
+.. attribute:: Target.dryrun
+
+   The accumulated dry-run commands. If dry-run mode is not active,
+   this will return the last accumulation list.
+
 .. method:: Target.connect([timeout])
    
    Establish a connection to the target. It is usually not necessary to call
@@ -217,9 +222,19 @@ Target
    :param timeout: timeout (in seconds) for the transfer; if the transfer does
        not  complete within this period, an exception will be raised.
 
+.. method:: Target.start_dryrun()
+
+   Starts dry-run mode, making further calls to Target.execute() accumulate
+   commands in a list rather than executing them.
+
+.. method:: Target.stop_dryrun()
+
+   Stops the dry-run mode.
+
 .. method:: Target.execute(command [, timeout [, check_exit_code [, as_root]]])
 
    Execute the specified command on the target device and return its output.
+   Command is not executed if dry-run mode is enabled.
 
    :param command: The command to be executed.
    :param timeout: Timeout (in seconds) for the execution of the command. If


setrofim:
There doesn't appear to be an API for accessing or doing anything with the accumulated dryrun data.

valschneider:
Right, I probably should have linked that here: https://github.com/ARM-software/lisa/pull/403/commits/1e421dc4481ae8c3ab765533341a156710a4fa0e
This LISA PR enables the dry-run mode, uses a few functions that end up calling target.execute(), then collects those commands and gathers them in a script that is pushed to the target. 

If the dry-run mode really doesn't belong here, I can introduce it LISA-side, it just so happens that it is much easier to add it in execute() than to all the functions that are going to call execute().

setrofim:
Right OK, I see. It seems non-obvious that `stop_dryrun` also returns the events. I would add a separate function/property for that. Also, please could you update the documentation for Target with the description of the dhryrun API.

setrofim:
Right OK, I see. It seems non-obvious that `stop_dryrun` also returns the events. I would add a separate function/property for that. Also, please could you update the documentation for Target with the description of the dhryrun API.

valschneider:
Added a Target.dryrun getter property, and updated the doc.

devlib/target.py
```diff
@@ -145,6 +145,10 @@ def conn(self):
         else:
             return None
 
+    @property
+    def dryrun(self):
```
setrofim: +1

bjackman:
Oh, this is merged already? Can we still tweak the design? I think it's really important to get this right.

setrofim:
I'm not sure either a context manger or separate interface are desirable here.

> It's obvious when skimming the code that the commands aren't actually being executed (contextmgr gives you that too)

The whole point of a "dry run" is that you execute exactly the same code that would run otherwise, but without the side effects. So it *shouldn't* be obvious when skimming the code -- it's not the code that determines the dry run status. Dry run is something you want to be able to switch on and  off for one bit of code. E.g. inside your entry point, if you want to do it for the entire execution, or around a particular area during debug. In either case, you want to minimize the impact on the code that is actually being dry run.

With a context manager, there is no way to turn dry-run behavior without modifying code; with a separate object it is even worse, as you would end up having to re-write code. If anything, I would suggest going the other way -- rather than having start/stop methods in the `Target`, turn them into stand-alone functions that goggle a global flag. That way the dry-run functionality can be uniformly configured for other components (e.g host-side operations) with minimal support required in the code. (note: doing this now is probably going a step to far, and the methods in Target are fine at the moment).

bjackman:
Interesting. I think it would be great to have dry-run mode be totally transparent, but at the moment there are fundamental differences between calling stuff in dry-run and regular mode:

1. In regular mode, you get a return value and possibly exceptions from `execute` that influence flow control. In dry-run mode, `execute` never raises an exception and always returns `None`.
2. The other parameters than `command` are ignored in dry-run mode
3. The order of actual operations on the target does not follow the order of `Target` method calls in the code, unless they are all `.execute`. 

2 is probably relatively easy to solve. 3 can be solved by only by tweaking _lots_ of code in the target classes (after making lots of decisions about the semantics of each method & memoization wrt. dry-run mode). Solving 1 essentially means constructing a DSL that generates `sh` code! 

Therefore I don't think transparency is a tractable goal, and code that uses dry-run mode needs to be aware of the fact that it's using dry run mode.

valschneider:
> Alternative idea: when I've been thinking about designing an API like this in the past, I've considered that the dry-run target should be a separate object, that you might use something like this

I also considered this, but went with the current quick and dirty tweak simply because it seemed to be good enough in the short run.

In my use-case, a context manager would fit perfectly and would look neater. However as @setrofim mentioned, this causes a lot of issues if you want to trigger the dry-run at some point, then exit the scope of your method but still want the dry-run to be enabled. I think that for the current use-case that is mostly accumulating data, a global toggle should be enough.

setrofim:
> To me these both suggest that the client code must be aware of the dry-run, and someone reading the code should clearly be able to see this. Either the dry-run is 100% transparent (unfeasible) or client code is aware.
> 
> Furthermore IMO there should be a clear delineation between what can be called, with which arguments, in dry-run contexts and what can't. This delineation could move as the feature is extended in the future, but this should never result in a change in behaviour for the same piece of code, except for going from "I got an error saying I can't call write_value within a dry-run" to "my code worked".

I don't necessarily disagree  with the above, but that is what leads me to

> Perhaps individual devlib calls are too low a level at which to implement dry-run execution for a script?

My point here is that, if the client *must* be aware of dry-run mode, is there any benefit in making `Target` aware of it as well, or can it be handled entirely in the client? So that in, in dry-run mode, the client doesn't even make calls to the target but does something else instead (e.g. logs the target method it would have called, along with params).

bjackman:
Ah, gotcha, yeah that's a good point at least as far as the current use case goes - it doesn't really make use of anything in devlib but rather is a trick for avoid having to rewrite client stuff that currently calls `target.execute`.

For more complex use cases I think it could be extended into a useful feature but I don't think this API is a good starting point for that extension.

@valschneider, this discussion has made me think that this functionality is better implemented purely in LISA, even if that is more painful. If we have to we can even replace the `target` in the relevant LISA code with a dummy object that has only an `execute` method (let's avoid that if at all possible). Do you agree?

setrofim:
OK, so it looks like I jumped the gun in merging this pull request. Are you happy for me to revert it?

valschneider:
@bjackman Right, following that discussion it seems that there's too many bits in devlib for this dry-run implementation to be considered adequate. What I needed for now is a command accumulator more than a proper dry-run, which is something else. I'll have a go at it in LISA.
@setrofim  Revert away !

PR #134, opened 2017-06-20
instrument: Add power monitoring support on Gem5 platforms

qperret:
This adds an instrument to enable power monitoring in Gem5 by parsing simulation statistics.

### DIFF
--- a/devlib/instrument/gem5power.py
+++ b/devlib/instrument/gem5power.py
@@ -1,0 +1,74 @@
+#    Copyright 2017 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import division
+import csv
+import re
+
+from devlib.platform.gem5 import Gem5SimulationPlatform
+from devlib.instrument import Instrument, CONTINUOUS, MeasurementsCsv
+from devlib.exception import TargetError, HostError
+
+
+class Gem5PowerInstrument(Instrument):
+    '''
+    Instrument enabling power monitoring in gem5
+    '''
+
+    mode = CONTINUOUS
+    roi_label = 'power_instrument'
+    
+    def __init__(self, target, power_sites):
+        '''
+        Parameter power_sites is a list of gem5 identifiers for power values. 
+        One example of such a field:
+            system.cluster0.cores0.power_model.static_power
+        '''
+        if not isinstance(target.platform, Gem5SimulationPlatform):
+            raise TargetError('Gem5PowerInstrument requires a gem5 platform')
+        if not target.has('gem5stats'):
+            raise TargetError('Gem5StatsModule is not loaded')
+        super(Gem5PowerInstrument, self).__init__(target)
+
+        # power_sites is assumed to be a list later
+        if isinstance(power_sites, list):
+            self.power_sites = power_sites
+        else:
+            self.power_sites = [power_sites]
+        self.add_channel('sim_seconds', 'time')
+        for field in self.power_sites:
+            self.add_channel(field, 'power')
+        self.target.gem5stats.book_roi(self.roi_label)
+        self.sample_period_ns = 10000000
+        self.target.gem5stats.start_periodic_dump(0, self.sample_period_ns)
+
+    def start(self):
+        self.target.gem5stats.roi_start(self.roi_label)
+
+    def stop(self):
+        self.target.gem5stats.roi_end(self.roi_label)
+       
+    def get_data(self, outfile):
+        active_sites = [c.site for c in self.active_channels]
+        with open(outfile, 'wb') as wfh:
+            writer = csv.writer(wfh)
+            writer.writerow([c.label for c in self.active_channels]) # headers
+            for rec, rois in self.target.gem5stats.match_iter(active_sites, [self.roi_label]):
+                writer.writerow([float(rec[s]) for s in active_sites])
+        return MeasurementsCsv(outfile, self.active_channels)
+    
+    def reset(self, sites=None, kinds=None, channels=None):
+        super(Gem5PowerInstrument, self).reset(sites, kinds, channels)
+        self.target.gem5stats.reset_origin()
+
--- a/devlib/__init__.py
+++ b/devlib/__init__.py
@@ -17,6 +17,7 @@ from devlib.instrument.frames import GfxInfoFramesInstrument
 from devlib.instrument.hwmon import HwmonInstrument
 from devlib.instrument.monsoon import MonsoonInstrument
 from devlib.instrument.netstats import NetstatsInstrument
+from devlib.instrument.gem5power import Gem5PowerInstrument
 
 from devlib.trace.ftrace import FtraceCollector
 
--- a/devlib/module/gem5stats.py
+++ b/devlib/module/gem5stats.py
@@ -28,6 +28,7 @@ class Gem5ROI:
         self.target = target
         self.number = number
         self.running = False
+        self.field = 'ROI::{}'.format(number)
 
     def start(self):
         if self.running:
@@ -42,7 +43,7 @@ class Gem5ROI:
         self.target.execute('m5 roiend {}'.format(self.number))
         self.running = False
         return True
-    
+
 class Gem5StatsModule(Module):
     '''
     Module controlling Region of Interest (ROIs) markers, satistics dump 
@@ -56,7 +57,7 @@ class Gem5StatsModule(Module):
 
     @staticmethod
     def probe(target):
-       return isinstance(target.platform, Gem5SimulationPlatform)
+        return isinstance(target.platform, Gem5SimulationPlatform)
 
     def __init__(self, target):
         super(Gem5StatsModule, self).__init__(target)
@@ -112,26 +113,41 @@ class Gem5StatsModule(Module):
         Keys must match fields in gem5's statistics log file. Key example:
             system.cluster0.cores0.power_model.static_power
         '''
+        records = defaultdict(lambda : defaultdict(list))
+        for record, active_rois in self.match_iter(keys, rois_labels):
+            for key in record:
+                for roi_label in active_rois:
+                    records[key][roi_label].append(record[key])
+        return records
+
+    def match_iter(self, keys, rois_labels):
+        '''
+        Yields for each dump since origin a pair containing:
+        1. a dict storing the values corresponding to each of the specified keys
+        2. the list of currently active ROIs among those passed as parameters.
+
+        Keys must match fields in gem5's statistics log file. Key example:
+            system.cluster0.cores0.power_model.static_power
+        '''
         for label in rois_labels:
             if label not in self.rois:
                 raise KeyError('Impossible to match ROI label {}'.format(label))
             if self.rois[label].running:
                 self.logger.warning('Trying to match records in statistics file'
                         ' while ROI {} is running'.format(label))
+        
+        def roi_active(roi_label, dump):
+            roi = self.rois[roi_label]
+            return (roi.field in dump) and (int(dump[roi.field]) == 1)
 
-        records = {}
-        for key in keys:
-            records[key] = defaultdict(list)
         with open(self._stats_file_path, 'r') as stats_file:
             stats_file.seek(self._current_origin)
             for dump in iter_statistics_dump(stats_file):
-                for label in rois_labels:
-                    # Save records only when ROIs are ON
-                    roi_field = 'ROI::{}'.format(self.rois[label].number)
-                    if (roi_field in dump) and (int(dump[roi_field]) == 1):
-                        for key in keys:
-                            records[key][label].append(dump[key])
-        return records
+                active_rois = [l for l in rois_labels if roi_active(l, dump)]
+                if active_rois:
+                    record = {k: dump[k] for k in keys}
+                    yield (record, active_rois)
+
 
     def reset_origin(self):
         '''


devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+        self.add_channel('Duration', 'time', 'sim_seconds')
+        # Track power only for now
+        for f in channel_map:
+            self.add_channel(f, 'power', channel_map[f])
+        
+        self._stats_file_path = stats_file_path
+        self._roi_state = [False] * 8
+        self.dumps = []
+
+    def start(self, roi=0):
```
setrofim: `roi` should be specified for the instrument on `__init__`. The API methods *must* match the base signatures.

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+        aggregates all the data of stats.txt corresponding to a specific ROI
+        number. Thus, several time intervals marked with the same ROI number 
+        will be aggregated as one if stats.txt has not been cleaned.
+        '''
+        super(Gem5StatsInstrument, self).reset(sites, kinds, channels)
+        for i in range(0,8):
+            if self._roi_state[i]:
+                self.target.execute('m5 roiend {}'.format(i))
+                self._roi_state[i] = False
+        self.target.execute('m5 dumpresetstats')
```
Sticklyman1936: Again not needed.

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+# limitations under the License.
+#
+from __future__ import division
+import csv
+import re
+
+from devlib.instrument import Instrument, CONTINUOUS, MeasurementsCsv
+from devlib.exception import TargetError, HostError
+
+class Gem5StatsInstrument(Instrument):
```
setrofim: As this instrument is specifically for measuring power, it should be given a more specific name to reflect that, e.g. `Gem5PowerInstrument`.

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+        if not self._roi_state[roi]:
+            raise TargetError('ROI {} not running'.format(roi))
+        self._roi_state[roi] = False
+        self.target.execute('m5 dumpresetstats')
+        self.target.execute('m5 roiend {}'.format(roi))
+
+    def get_data(self, outfile, roi=0):
+        if roi > 7 or roi < 0:
+            raise RuntimeError('Incorrect ROI number')
+        self.target.execute('m5 dumpresetstats')
```
Sticklyman1936: As I understand it, one should call `stop` before calling `get_data`. Therefore, the data for the ROI should already be dumped. This one should also not be needed.

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+        with open(self._stats_file_path, 'r') as f:
+            self._parse_stats(f)
```
Sticklyman1936: The way that this is implemented, you could potentially dump for ROIs where were not stopped, yet. Is this something we want? Would it not make more sense check that the ROI is stopped (you already track this in `self._roi_state`)?

Have you considered the memory requirements for this? I worry that loading the data for 10,000 stats dumps into memory will kill many lower spec'd systems.

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+    power analysis on a specific time window which spans over the execution of
+    a workload for example. Up to 8 ROIs can be used, with or without 
+    overlapping.
+    '''
+
+    mode = CONTINUOUS
+    dump_head='---------- Begin Simulation Statistics ----------'
+    dump_tail='---------- End Simulation Statistics   ----------'
+    
+    def __init__(self, target, channel_map, stats_file_path):
```
setrofim: Please document the use of these parameters as it is non-obvious.

`channel_map` -- this should probably be renamed. The calling code is not necessarily aware of what channels are. Normally, this information would be extracted from the target; if something is needed from the user in order to define the channels, please rename this to indicate what those things are (e.g. `power_rails`, `regions_of_interest`, or whatever).

`stats_file_path` -- should this not be acquired from the target rather than passed separately?

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+        super(Gem5StatsInstrument, self).__init__(target)
+
```
setrofim: Since this instrument works *only* with gem5, there should be a check here to ensure that the target's platform is indeed a gem5 platform and raise a `ValueError` here if it isn't.

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+        Beware that reset() does **NOT** erase "stats.txt". Doing so while 
+        Gem5 is running might be harmful. A proper Gem5 reboot is preferable. 
+        Consequently, a ROI should probably not be used more than once
+        without cleaning "stats.txt" first (by restarting Gem5). get_data() 
+        aggregates all the data of stats.txt corresponding to a specific ROI
+        number. Thus, several time intervals marked with the same ROI number 
+        will be aggregated as one if stats.txt has not been cleaned.
+        '''
+        super(Gem5StatsInstrument, self).reset(sites, kinds, channels)
+        for i in range(0,8):
```
setrofim: `for i in xrange(8):`

devlib/instrument/gem5stats.py
Sticklyman1936: From what I understand, this is not the best way to go about using this. I would have expected to have these two bits of functionality split. Specifically:

* One instrument which sets up an starts periodic stats dumping, and does the processing of the dumps.
    * This could potentially still filter on the ROI, but it should not be a requirement.
* One instrument which marks the starts and ends of regions of interest.

By combining these two, I think you are limiting the functionality, as well as complicating the interface.

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+            if self._roi_state[i]:
+                self.target.execute('m5 roiend {}'.format(i))
+                self._roi_state[i] = False
+        self.target.execute('m5 dumpresetstats')
+
+    # Config tree management
+    def _parse_stats(self,f):
+        # Builds a dict-based tree for each simulation stats. 
+        regex = re.compile("^(?P<key>[^- ]\S*) +(?P<value>[^#]+).+$")
```
setrofim: pre-compile at top level?

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+                continue
+            if self.dump_tail in l:
+                i -= 1
+                continue
+            res = regex.match(l)
+            if res:
+                k = res.group("key")
+                v = res.group("value").split()
+                _insert_val( self.dumps[-1], k , v )
+        if i!=0:
```
setrofim: `if i:`

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+                self._roi_state[i] = False
+        self.target.execute('m5 dumpresetstats')
+
+    # Config tree management
+    def _parse_stats(self,f):
+        # Builds a dict-based tree for each simulation stats. 
+        regex = re.compile("^(?P<key>[^- ]\S*) +(?P<value>[^#]+).+$")
+        self.dumps = [] 
+        i = 0
+        for l in f:
```
setrofim: Please  use more meaningful names for variables -- favor clarity over brevity.

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+            raise TargetError('ROI {} not running'.format(roi))
+        self._roi_state[roi] = False
+        self.target.execute('m5 dumpresetstats')
+        self.target.execute('m5 roiend {}'.format(roi))
+
+    def get_data(self, outfile, roi=0):
+        if roi > 7 or roi < 0:
+            raise RuntimeError('Incorrect ROI number')
+        self.target.execute('m5 dumpresetstats')
+        with open(self._stats_file_path, 'r') as f:
```
Sticklyman1936: Note that it is possible to gzip compress the stats files directly with gem5 (typically, this is the norm for larger runs - saves around 90% of the space). In that case, you need to do a `import gzip; gzip.open(...)`.

HOWEVER, it is likely that you cannot read the gzip compressed file whilst gem5 still has the file open! This is something we might need to solve in the future.

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+        self.dumps = []
+
+    def start(self, roi=0):
+        if roi > 7 or roi < 0:
+            raise IndexError('Incorrect ROI number')
+        if self._roi_state[roi]:
+            raise TargetError('ROI {} already running'.format(roi))
+        self._roi_state[roi] = True
+        self.target.execute('m5 roistart {}'.format(roi))
+        self.target.execute('m5 dumpresetstats')
```
setrofim: OK, this probably should not be done here. Marking of regions of interest should probably be implemented as a module rather than an Instrument. Instrument APIs imply a very specific usage pattern, and Instruments that collect measurements of the same type are intended as drop-in replacements for each other. The methods would typically be invoke in reset-start-stop-get_data sequence.

devlib/instrument/gem5stats.py
Sticklyman1936: If this is purely for power, then it should be named as such. Ideally, I would suggest making it generic.

qperret: Ok you're right. As mentionned by @Sticklyman1936, it could probably be made generic. I'll look into that.

devlib/instrument/gem5stats.py
qperret: Ok I'll add documentation for that. They might change a little bit anyway as I'll try to make it more generic.

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+    '''
+    Instrument enabling power monitoring in Gem5
+
+    This instrument parses the stats file produced by Gem5 and extracts power
+    values. More information on the "stats.txt" file structure can be found 
+    here:
+    
+    http://pages.cs.wisc.edu/~david/courses/cs752/Spring2015/gem5-tutorial/part1/gem5_stats.html
+    
+    This instrument also implements Gem5's ROIs (Regions of Interest). A ROI is
```
qperret: Ok, thanks. That's good to know.

devlib/instrument/gem5stats.py
qperret: You're right, I have dumpresetstats calls around that were there just to be sure not to miss things but some of them are not needed.

devlib/instrument/gem5stats.py
qperret: Hmm I see your point but I'm really not sure if marking and reading from ROIs should be split in two parts. ROIs seem to be perfectly appropriate to implement start() and stop() methods of this instrument in my opinion. 
What could be possible is to completely hide the notion of ROI to users (which would meet @setrofim's request to have API methods to match base signatures btw) and to assign a ROI to each instrument instance in order to differentiate their start/stop call from the unique stats.txt file. This could be done using a simple class counter. Does it sound reasonable ?

devlib/instrument/gem5stats.py
qperret: That's actually a good question. I understand your concern but after all, is it absolutely required to stop the ROI first ? What if we want to have an intermediate value to see how things are going while the simulation is running ? Maybe we should just raise a warning ?

Indeed, memory is a concern given the typical size of stats.txt files. As an improvement I can filter out the important dumps based on ROI values and ignore everything else.

devlib/instrument/gem5stats.py
qperret: Agreed. Please have a look at my answer to @Sticklyman1936's comment about this.

devlib/instrument/gem5stats.py
setrofim: It seems there are two separate bits of functionality here:

- Marking of the regions of interest
- Collection of power measurements (possibly filtered by one or more regions of interest)

It doesn't seem to be a reason for the two to be linked (a region of interested may be marked for something other than power, right?), and there may be a need to keep them separate. E.g. of more than one region needs to be marked , this does not go with the reset-start-stop-get_data usage model -- e.g. a start might need to be called multiple times (for different regions) before a first stop, which would not happen for an instrument.

This is why I've suggested implementing the marking functionality as a module -- it can be gem5-specific, and does not dictate a particular usage model.

qperret: Assuming that the instrument will be made more general, meaning not focusing only on power fields but also applicable on any field of the stats file, you will need to use ROIs to define start and stop of measurements anyway. You could set up this instrument to track the number of L3 cache misses while some task is running using the same start/stop/get_data usage model.

I don't think _what_ you want to measure and _when_ you want to measure it is dissociated in any of devlib's instrument so far. You always say what you're focusing on in `__init__` and when you focus on it with start/stop. That's the same thing here, ROIs are just a mean of implementing start/stop cleanly.

devlib/instrument/gem5stats.py
```diff
@@ -0,0 +1,152 @@
...
+            for s in (self.dumps):
+                if int(s['ROI::{}'.format(roi)])==0: # ignore out-of-roi dumps
+                    continue
+                row = [float(_get_val(s, self.channels[c].name)) for c in cur_chans]
+                writer.writerow(row)
+        return MeasurementsCsv(outfile, self.active_channels)
+    
+    def reset(self, sites=None, kinds=None, channels=None):
+        '''
+        Beware that reset() does **NOT** erase "stats.txt". Doing so while
```
setrofim: This breaks the API. The API dictates that reset must reset the instrument for a fresh collection session. The calling code may call reset-start-stop-get_data to do something on the target and then immediately perform another reset-start-stop-get_data session, and the stats collected for that must not be influenced by the previous session.

Perhaps this should not be manipulating the raw stats file directly. Access to the stats can be mediated via something like a `get_stats` method of a gem5-specific module, which would track where in the stats file it was last time it was invoked and offset accordingly on each call.

qperret: Yes, that's a good idea. This could be done by saving the number of dumps since the beginning of the simulation every time reset is called and fast-forwarding to it on each get_data call. Does is sound reasonable ?

devlib/instrument/gem5stats.py
qperret: Just to elaborate a little bit more on this, I understand your point in regards to separating the instrument's features that are exposed to users and managing gem5 features (ROIs) internally. I agree that an elaborated wrapper on gem5's commands is usefull. What I am really trying to emphasize is that maybe this instrument is that wrapper already and that we don't need an additional module to do it. Do you see where we could benefit from this gem5 module in other parts of devlib ?

qperret:
I wrote a module to manage gem5's statistics file as requested and submitted it in a separate PR #142 since it aims at solving a different issue. I will keep this PR focused on its original purpose: a power instrument.

qperret:
I updated the instrument to use the gem5 statistics module merged in PR #142.

devlib/instrument/gem5power.py
```diff
@@ -0,0 +1,79 @@
...
+    '''
+    Instrument enabling power monitoring in gem5
+    '''
+
+    mode = CONTINUOUS
+    roi_label = 'power_instrument'
+    
+    def __init__(self, target, power_fields):
+        '''
+        Parameter power_fields is a list of gem5 identifiers for power values.
```
Sticklyman1936: Maybe `power_fields` is a bit misleading. How about `power_sites` instead?

devlib/instrument/gem5power.py
```diff
@@ -0,0 +1,79 @@
...
+    
+    def __init__(self, target, power_fields):
+        '''
+        Parameter power_fields is a list of gem5 identifiers for power values. 
+        One example of such a field:
+            system.cluster0.cores0.power_model.static_power
+        '''
+        if not isinstance(target.platform, Gem5SimulationPlatform):
+            raise TargetError('Gem5PowerInstrument requires a gem5 platform')
+        if not hasattr(target, 'gem5stats'):
```
Sticklyman1936: Can we change this to `target.has('gem5stats')` ?

devlib/instrument/gem5power.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def start(self):
+        self.target.gem5stats.roi_start(self.roi_label)
+
+    def stop(self):
+        self.target.gem5stats.roi_end(self.roi_label)
+       
+    def get_data(self, outfile):
+        active_sites = [c.site for c in self.active_channels]
+        records = self.target.gem5stats.match(active_sites, [self.roi_label])
```
Sticklyman1936: It would be good if this could use an iterator. Else you still need the memory for all of the match stats dumps.

Ideally, cover both cases by adding a `match_iter` method to the gem5stats module.

qperret: It would be clearly more elegant but I'm not sure if it's really needed. In practice you will keep in memory the (probably small) subset of fields matched over the dumps covered by the ROI. So, assuming that we track 20 power channels during 3000 dumps (30 seconds of target-time (which is long with gem5) with dumps at 100Hz) and that each value is a double-precision floating point number, we end-up using 20*3000*8 < 0.5MiB of memory. Problems will occur only if we start tracking *many* things in the stats file

devlib/instrument/gem5power.py
qperret: Done

devlib/instrument/gem5power.py
qperret: Done

devlib/instrument/gem5power.py
qperret: I tried to update as requested. Let me know what you think

Issue #136, opened 2017-06-22
Freezing userspace task not fully reliable


qperret:
Freezing userspace tasks is currently achieved by setting the state of the appropriate cgroup to `FROZEN`. This is done by calling `target.write_value` which (by default) raises an exception if the read value does not match the written one. If freezing the tasks does not happen promptly, the read value is `FREEZING` instead of `FROZEN` and the following exception is raised:
`TargetError: Could not set the value of /data/local/tmp/devlib-target/cgroups/devlib_cgh3/DEVLIB_FREEZER/freezer.state to "FROZEN" (read "FREEZING")`

The correct behaviour is to probably to re-read in a loop the value after a short delay if `FREEZING` is read until `FROZEN` is reached or to raise an exception in all other cases.

setrofim:
I've submitted a pull request with a proposed fix: https://github.com/ARM-software/devlib/pull/137

Please could you review, and let me know if this fixes the issue in your setup.

qperret:
This should be fixed by #159.

PR #137, opened 2017-06-23
cgroup: wait to verify value in set()

setrofim:
This is a fix for https://github.com/ARM-software/devlib/issues/136

The standard behavior for write_value() is to read it back immediately
afterwards and verify that it has been written. When writing some cgroup
parameter values (e.g. "FROZEN"), it may take a while for them to take
effect. If they are read back immediately afterwards, the result won't
match the expected value, producing an error.

While it is possible to disable validation on write, it is still
desirable to make sure that the parameters are set properly. This commit
changes the behavior to continue reading the value up to a timeout
period afterwards, giving it time to take effect, and only raise an
error if the timeout is reached.

### DIFF
--- a/devlib/module/cgroups.py
+++ b/devlib/module/cgroups.py
@@ -14,6 +14,7 @@
 #
 # pylint: disable=attribute-defined-outside-init
 import logging
+import os
 from collections import namedtuple
 
 from devlib.module import Module
@@ -207,13 +208,15 @@ class Controller(object):
             tasks[cg] = self.tasks_count(cg)
         return tasks
 
+
 class CGroup(object):
 
-    def __init__(self, controller, name, create=True):
+    def __init__(self, controller, name, create=True, set_timeout=10):
         self.logger = logging.getLogger('cgroups.' + controller.kind)
         self.target = controller.target
         self.controller = controller
         self.name = name
+        self.set_timeout = set_timeout
 
         # Control cgroup path
         self.directory = controller.mount_point
@@ -273,7 +276,7 @@ class CGroup(object):
 
             # Set the attribute value
             try:
-                self.target.write_value(path, attrs[idx])
+                self._do_set(path, attrs[idx])
             except TargetError:
                 # Check if the error is due to a non-existing attribute
                 attrs = self.get()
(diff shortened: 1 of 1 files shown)


qperret:
Thanks for the fix. 

Please `import os.path`. Apart from that it looks good to me. It's probably a shame to wait before raising an exception if the read value if clearly wrong but I don't really see how to do that in a generic way. This is probably the best we can do without managing all transition states of cgroups independently.

setrofim:
> Please `import os.path`.

Done.

>  It's probably a shame to wait before raising an exception if the read value if clearly wrong but I don't really see how to do that in a generic way. 

Well this is only in the exceptional case, and only 10s of target time. The alternative would be to add an awareness of in-progress/final value pairs, such as FREEZING/FROZEN, but that would periodically break when things change.

qperret:
> The alternative would be to add an awareness of in-progress/final value pairs, such as FREEZING/FROZEN, but that would periodically break when things change.

I agree, it's probably not worth it.

Design-wise, would it make sense to move the timeout feature to `target.write_value` ? It could be used together with `verify` to wait before raising an exception. I would suggest a default timeout of -1 meaning raise exception with no delay.

devlib/module/cgroups.py
```diff
@@ -207,13 +208,15 @@ def tasks_per_group(self):
             tasks[cg] = self.tasks_count(cg)
         return tasks
 
+
 class CGroup(object):
 
-    def __init__(self, controller, name, create=True):
+    def __init__(self, controller, name, create=True, set_timeout=10):
```
derkling: The timeout is really required only for the `freezer` controller... it seems a bit overfilling to have general timeout defined here for each and every controller.

What about moving it to the `set(self, **attrs)` method below or, maybe even better, fix the only really broken client, which is the `freeze` method defined at the bottom of this module?
Maybe we can implement freeze/thaw operations as a dedicated shutil call, which will properly include the required timeout for this operation only, to be used by the freeze() method.

derkling:
I would better consider the shutil alternative implementation to better support freeze/thaw operations.

setrofim:
> I would better consider the shutil alternative implementation to better support freeze/thaw operations.

Yeah, I think that might be a better idea. Closing this pull.

PR #140, opened 2017-06-28
Let the caller decide whether using sudo is needed or not

AnthonyARM:
Checking the username is not the right way of checking whether or not 'sudo' is needed, on some platforms the user is called 'root' but is not actually privileged (And therefore requires the use of 'su' or 'sudo').

If you really want to check whether or not sudo is needed, then you should do like WA and check `id == 0` rather than the username.


setrofim:
I'm uncomfortable merging this as-is. In my experience, user called "root"  not having id of 0 is a much rarer condition that a target not having sudo installed. So simply removing this check would cause more breaks than it would fix. Please either update with an alternative check, or raise this as an issue instead, so that it may be resolved properly at a later time.

AnthonyARM:
Found another work around with my platform, abandoning this pull request

PR #141, opened 2017-06-28
Check if any big/LITTLE core is online before dereferencing a potentially empty list

AnthonyARM:


setrofim:
These should have else condition raising something like `ValueError('All bigs/littles appear to be offline')` to prevent silent failures, especially for the `set_` methods (where there may be no feedback otherwise that the target is not configured in the way the caller expects).

PR #142, opened 2017-07-05
module: Control and parse gem5's statistics log file

qperret:
Gem5's statistics log file contains plenty of interesting information
that are not exposed so far. This module enables control and parsing of
the statistics file by:
 - configuring periodic dumps of statistics;
 - marking Regions of Interest (ROIs);
 - and extracting values of specific fields during the ROIs.

### DIFF
--- a/devlib/module/gem5stats.py
+++ b/devlib/module/gem5stats.py
@@ -1,0 +1,148 @@
+#    Copyright 2017 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import os.path
+from collections import defaultdict
+
+import devlib
+from devlib.module import Module
+from devlib.platform import Platform
+from devlib.platform.gem5 import Gem5SimulationPlatform
+from devlib.utils.gem5 import iter_statistics_dump, GEM5STATS_ROI_NUMBER, GEM5STATS_DUMP_TAIL
+
+
+class Gem5ROI:
+    def __init__(self, number, target):
+        self.target = target
+        self.number = number
+        self.running = False
+
+    def start(self):
+        if self.running:
+            return False
+        self.target.execute('m5 roistart {}'.format(self.number))
+        self.running = True
+        return True
+    
+    def stop(self):
+        if not self.running:
+            return False
+        self.target.execute('m5 roiend {}'.format(self.number))
+        self.running = False
+        return True
+    
+class Gem5StatsModule(Module):
+    '''
+    Module controlling Region of Interest (ROIs) markers, satistics dump 
+    frequency and parsing statistics log file when using gem5 platforms.
+
+    ROIs are identified by user-defined labels and need to be booked prior to
+    use. The translation of labels into gem5 ROI numbers will be performed
+    internally in order to avoid conflicts between multiple clients.
+    '''
+    name = 'gem5stats'
+
+    @staticmethod
+    def probe(target):
+       return isinstance(target.platform, Gem5SimulationPlatform)
+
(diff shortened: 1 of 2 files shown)


devlib/module/gem5stats.py
```diff
@@ -0,0 +1,132 @@
...
+            stats_file.seek(self._current_origin)
+            for dump in self._iter_dump(stats_file):
+                for roi in xrange(GEM5STATS_ROI_NUMBER):
+                    # Save records only when ROIs are ON
+                    if int(dump['ROI::{}'.format(roi)])==1:
+                        for key in keys:
+                            records[key][roi].append(dump[key])
+        return records
+
+    def _iter_dump(self, stats_file):
```
setrofim: This does not need to be a method, as it does not rely on any of the module's internal state. Making this into a stand-alone function would make it easier to reuse. I would also move this out into `devlib/utils/gem5.py`

devlib/module/gem5stats.py
```diff
@@ -0,0 +1,132 @@
...
+
+        records = {}
+        for key in keys:
+            records[key] = defaultdict(list)
+        with open(self._stats_file_path, 'r') as stats_file:
+            stats_file.seek(self._current_origin)
+            for dump in self._iter_dump(stats_file):
+                for roi in xrange(GEM5STATS_ROI_NUMBER):
+                    # Save records only when ROIs are ON
+                    if int(dump['ROI::{}'.format(roi)])==1:
```
setrofim: Spaces around `==`

setrofim: Is `'ROI::{}'.format(roi)` guaranteed to be in the dump, even if ROI are not used? If not, this should be made more robust.

devlib/module/gem5stats.py
```diff
@@ -0,0 +1,132 @@
...
+    def probe(target):
+       return isinstance(target.platform, devlib.platform.gem5.Gem5SimulationPlatform)
+
+    def __init__(self, target):
+        super(Gem5StatsModule, self).__init__(target)
+        self._roi_state = [False] * GEM5STATS_ROI_NUMBER
+        self._current_origin = 0
+        self._stats_file_path = os.path.join(target.platform.gem5_out_dir,
+                                            'stats.txt') 
+        self.logger = logging.getLogger('Gem5Stats')
```
setrofim: This is unnecessary as the logger is created by the super.

devlib/module/gem5stats.py
```diff
@@ -0,0 +1,132 @@
...
+
+import re
+import logging
+import os.path
+from collections import defaultdict
+
+import devlib
+from devlib.module import Module
+from devlib.platform import Platform
+
```
setrofim: Two lines between includes and globals

devlib/module/gem5stats.py
```diff
@@ -0,0 +1,132 @@
...
+GEM5STATS_FIELD_REGEX = re.compile("^(?P<key>[^- ]\S*) +(?P<value>[^#]+).+$")
+GEM5STATS_DUMP_HEAD = '---------- Begin Simulation Statistics ----------'
+GEM5STATS_DUMP_TAIL = '---------- End Simulation Statistics   ----------'
+GEM5STATS_ROI_NUMBER = 8
+
```
setrofim: Two lines between globals and class declarations

devlib/module/gem5stats.py
```diff
@@ -0,0 +1,132 @@
...
+class Gem5StatsModule(Module):
+    '''
+    Module controlling Region of Interest markers, satistics dump frequency and
+    parsing statistics log file when using gem5 platforms.
+    '''
+    name = 'gem5stats'
+
+    @staticmethod
+    def probe(target):
+       return isinstance(target.platform, devlib.platform.gem5.Gem5SimulationPlatform)
```
setrofim: Import `Gem5SimulationPlatform` at the top and just use the short name here.

devlib/module/gem5stats.py
qperret: Good question ... The ROI fields have always been present during my runs, they were just equal to 0 all the time if no ROI was used. Maybe @Sticklyman1936 could confirm there are no special cases/configs where the ROI fields are not dumped ?

qperret:
I moved the dump iterator into utils.gem5 as requested and I added a ROI booking system. The idea is that several clients (different instruments for example) might use this module simultaneously and they should not share ROIs without knowing it. To solve the issue I'm proposing to hide the ROI numbers from clients which will use labels instead.

PR #143, opened 2017-07-06
Add option to run target.invoke() in the background

valschneider:
I find it useful to launch several tasks and pin each of them to a specific cpu. The implementation is trivial, maybe a separate invoke_bg method with the `timeout` parameter would be preferred ?


setrofim:
A separate method is preferred, as the blocking behavior and  return value differ. Please do not abbreviate to "bg", as that is not sufficiently clear. Something like `invoke_in_background` or `background_invoke`

valschneider:
Moved the changes in a separate `background_invoke()` method.

PR #144, opened 2017-07-10
utils/misc: update MIDR table

setrofim:
Update the MIDR CPU part number table with values from
https://github.com/torvalds/linux/blob/master/arch/arm64/include/asm/cputype.h

### DIFF
--- a/devlib/utils/misc.py
+++ b/devlib/utils/misc.py
@@ -79,8 +79,17 @@ CPU_PART_MAP = {
         0xd08: {None: 'A72'},
         0xd09: {None: 'A73'},
     },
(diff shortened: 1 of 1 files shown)


devlib/utils/misc.py
```diff
@@ -79,9 +79,18 @@
         0xd08: {None: 'A72'},
         0xd09: {None: 'A73'},
     },
+    0x42: {  # Boardcom
```
marcbonnici: Should this be `Broadcom?`

setrofim: Yes. Yes, it should. Corrected.

PR #149, opened 2017-07-19
[RFC] module/gem5stats: parse zipped statistics file

qperret:
This is an attempt to parse gem5's statistics file with on-line compression. The PR should probably **not** be merged in the current state, I submitted it to collect ideas and feedback. 

I think the overall state of the generic zipped text file reader and the connection with the existing gem5 infrastructure added with this PR are probably ok but there is still another gem5-specific issue that might be worth tackling. Tests on my machines revealed that, when using gem5 with on-line compression of stats log, dumps often need some time to be written into the output .gz file. Consequently it happens that calls to `gem5stats.match()` right after the end of a ROI can miss some of the latest dumps which have not been written in the statistics file yet. I tried to mitigate this issue by "syncing" the host OS between dump reads but it does not seem to be enough -- that's very likely an internal gem5 issue. Adding a small delay at the begining of `gem5stats.match()` could probably solve the issue but is looks quiet hacky.

Any ideas or suggestions to fix that more cleanly are very welcome :)

### DIFF
--- a/devlib/platform/gem5.py
+++ b/devlib/platform/gem5.py
@@ -55,6 +55,7 @@ class Gem5SimulationPlatform(Platform):
         self.stdout_file = None
         self.stderr_file = None
         self.stderr_filename = None
+        self.stats_filename = None
         if self.gem5_port is None:
             # Allows devlib to pick up already running simulations
             self.start_gem5_simulation = True
@@ -126,6 +127,13 @@ class Gem5SimulationPlatform(Platform):
             # We need to keep this so we can check which port to use for the
             # telnet connection.
             self.stderr_filename = f
+
+            # Keep track of the statistics log file
+            match = re.search('--stats-file=(?P<file>\S*)', self.gem5args_args)
+            if match:
+                self.stats_filename = match.group('file') 
+            else:
+                self.stats_filename = 'stats.txt'
 
             # Start gem5 simulation
             self.logger.info("Starting the gem5 simulator")
--- a/devlib/utils/zip.py
+++ b/devlib/utils/zip.py
@@ -1,0 +1,175 @@
+#    Copyright 2017 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import zlib
+from io import FileIO, UnsupportedOperation, TextIOBase
+try:
+    from cStringIO import StringIO
+except:
+    from StringIO import StringIO
+
+
+class BufferedZippedTextReader(TextIOBase):
+
+    def __init__(self, file_path, window_size=16384, wbits=47):
+        '''
+        parameters:
+            :file_path: path to the archive.
+            :window_size: size of buffer holding compressed data (bytes).
+            :wbits: configuration parameter for zlib decompressor.
+        '''
+        self.file_path = file_path
+        self.in_stream = FileIO(self.file_path, 'r')
+        self.window_size = window_size
+        self.wbits = wbits
+        self._reset()
+        self.mode = 'r'
+
+    def tell(self):
+        '''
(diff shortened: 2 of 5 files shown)


devlib/platform/gem5.py
```diff
@@ -127,6 +128,13 @@ def _start_interaction_gem5(self):
             # telnet connection.
             self.stderr_filename = f
 
+            # Keep track of the statistics log file
+            ret = re.search('--stats-file=(?P<file>\S*)', self.gem5args_args)
```
setrofim: Rename `ret` --> `match`.

setrofim: Is there a short-form version of the option (e.g. `-s` or something) that can be used to specify the stats file?

qperret: > Is there a short-form version of the option (e.g. -s or something) that can be used to specify the stats file?

I don't think so. Or at least it's not documented in the list of options in `--help`

devlib/utils/zip.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def __enter__(self):
+        self._reset()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
+
+    def __iter__(self):
+        return self
```
qperret: I think the parent `IOBase` implementation already does that

devlib/utils/zip.py
```diff
@@ -0,0 +1,174 @@
...
+            target_pos = self.tell() + offset
+        else:
+            raise ValueError('Invalid whence: {}'.format(whence))
+        if target_pos < self.tell():
+            self._reset()
+        while self.tell() != target_pos and self.tell() != -1:
+            self.char_iterator.next()
+        return self.tell()
+
+    def _iter_buff(self):
```
qperret: Done

devlib/utils/zip.py
```diff
@@ -0,0 +1,174 @@
...
+import zlib
+from io import FileIO, UnsupportedOperation, TextIOBase
+try:
+    from cStringIO import StringIO
+except:
+    from StringIO import StringIO
+
+
+class BufferedZippedTextReader(TextIOBase):
+    def __init__(self, file_path, window_size=16384, wbits=47):
```
qperret: Done

Sticklyman1936:
I am happy with this in general, but don't think that the sync here will help. I think that the buffering is happening during the gem5 output, and therefore forcing the OS to sync the buffers won't have any real effect. It seems like gem5 should already flush the buffers though:

```
void
Text::end()
{
    ccprintf(*stream, "\n---------- End Simulation Statistics   ----------\n");
    stream->flush();
}
```

`stream` can be either a standard C++ ostream, or a gzstream. However, in either case, flush should cause it to write out anything outstanding. Therefore, I am slightly stuck on what to suggest here. Maybe try and read the stream in a loop until the "End Simulation Statistics" line is encountered? Obviously, this would be a terrible idea if it is indeed buffered...

qperret:
Ok interesting to see that steam are flushed on dumps ... I agree that the host sync is probably useless here. Actually I never had that sort of problem without enabling the compression so the issue is probably not the OS' buffers that should be under less pressure with compressed, and hence less, data to write.

The question is, are the m5 calls "blocking" in some way ? I mean, assuming that `m5 dumpstats` is called on the gem5 target, is it required that the dump has completed on the host side to complete the command on the target ?

setrofim:
@qperret  @Sticklyman1936 @avanlaer what is the current status of this? Are more changes necessary or are we happy for this to be merged?

qperret:
Well, this can still be considered as somehow broken ... However, I'm not sure if the problem is really devlib-related. I think the zip-reader implemented in this PR is OK, it's just that there is a buffer somewhere (maybe in gem5?) that needs to be flushed _before_ devlib actually starts to read. We can probably workaround with delays (host-side) here and there but that's ugly. Otherwise, I don't know what we can change in the current code to make it work better. Maybe it's just OK to document the current behaviour and to use it as-is.

setrofim:
>  Maybe it's just OK to document the current behaviour and to use it as-is.

OK, I would be fine with this, if @Sticklyman1936  and @avanlaer  are.

qperret:
Ok cool. And actually since this is going to be used transparently from the user perspective, I think I'll also log a warning message when the zip reader is in use with something like: "Reading for a zipped stats file can involve buffering issues causing latest dumps to be ignored".

qperret:
I don't think this is used by anyone, and I don't really have time to update this PR so I close it now.
I'll keep the branch on my devlib repo for reference if someone wants to re-spin a PR something based on it.

PR #151, opened 2017-07-25
Add support for arbitrary ADB servers

AnthonyARM:
Allow the user to specify where the ADB server is located: in some cases the Android devices might be connected to a different machine. (See -H option: https://developer.android.com/studio/command-line/adb.html )

However, usually the ADB server runs on the same machine as the ADB client therefore the default value for the server is 'localhost'


setrofim:
Looks good, but please could you update commit message with a bit more detail on what this is doing and why it is useful.

PR #156, opened 2017-07-28
gem5: make explicit power modelling parameters

qperret:
It is currently assumed in the gem5 power instrument that the platform
has been appropriately configured to bind gem5 with power models. This
makes this configuration explicit in Gem5SimulationPlatform and extends 
the power instrument setup with a check on that.

### DIFF
--- a/devlib/platform/gem5.py
+++ b/devlib/platform/gem5.py
@@ -37,7 +37,8 @@ class Gem5SimulationPlatform(Platform):
                  big_core=None,
                  model=None,
                  modules=None,
-                 gem5_telnet_port=None):
+                 gem5_telnet_port=None,
+                 power_models_dir=None):
 
         # First call the parent class
         super(Gem5SimulationPlatform, self).__init__(name, core_names, core_clusters,
@@ -52,6 +53,7 @@ class Gem5SimulationPlatform(Platform):
         self.gem5_interact_dir = '/tmp' # Host directory
         self.executable_dir = None # Device directory
         self.working_dir = None # Device directory
+        self.power_models_dir = power_models_dir # Location of the power models
         self.stdout_file = None
         self.stderr_file = None
         self.stderr_filename = None
@@ -126,6 +128,11 @@ class Gem5SimulationPlatform(Platform):
             # We need to keep this so we can check which port to use for the
             # telnet connection.
             self.stderr_filename = f
+
+            # Enable power modelling if required
+            if self.power_models_dir:
+                os.environ['POWER_MODEL_ROOT'] = self.power_models_dir
+                self.gem5args_args += ' --power-model'
 
             # Start gem5 simulation
             self.logger.info("Starting the gem5 simulator")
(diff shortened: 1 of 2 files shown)


qperret:
@setrofim, @Sticklyman1936 , @ionela-voinescu 
Any chance one of you guys has a feedback on that ?

Sticklyman1936:
This looks good to me, but I would like @avanlaer to take a quick look.

devlib/platform/gem5.py
```diff
@@ -127,6 +129,11 @@ def _start_interaction_gem5(self):
             # telnet connection.
             self.stderr_filename = f
 
+            # Enable power modelling if required
+            if self.power_models_dir:
```
avanlaer: You might want to put a comment in that exact argument needed to 'activate' the gem5 power models will depend on the user's implementation of it and so the user might need to change this in order for it to work. Some gem5 configuration scripts do no use "--power-model" to enable power models, for example.

avanlaer:
This looks good to me as well, only one slight remark.

devlib/platform/gem5.py
setrofim: If this is the case, then perhaps that should be made configurable in the platform as well?

qperret: Sorry, I am not 100% sure to have understood that correctly. If power modelling is activated using completely different approaches depending on the platform scripts (I didn't know that), we probably don't want this PR at all since it will work for only one platform. Is that correct ? Are we always using `POWER_MODEL_ROOT` ? Is it just the `--power-model` argument that's an issue ?

avanlaer: You are right - the 'activation' of the power models depends on the platform scripts. Both the environment variable (POWER_MODEL_ROOT) and --power-model are specific to one platform script. There is no default method for activating power models in gem5 unfortunately.

setrofim: What other activation methods/scripts currently exist? Is there a reason there cannot be a standard interface for these?

Maybe we can still have some lowest-common-denominator configuration, e.g.  the model could take `power_model_args` and `power_model_env_vars` parameters.

qperret: > Maybe we can still have some lowest-common-denominator configuration, e.g. the model could take power_model_args and power_model_env_vars parameters.

Yes, we could have at least that. However, if something else is needed to activate power models for a specific platform, `power_model_args` and `power_model_env_vars` will not be enough and it will be _implicitly_ assumed that the user as configured the external bits correctly anyway. 

Actually, by exposing power-related parameters, we will give the impression that devlib is setting everything up and running for power while it will actually setup things only partially for some platforms, and that can be misleading ...
Maybe this is just too platform-dependent and we should leave users to manage that using `gem5args_args` in `Gem5SimulationPlatform`.

avanlaer: The gem5 configuration scripts are written by the user, hence why they are so diverse and non-standardised. I think Quentin is right and if these parameters are exposed here, the user might wrongly assume devlib sets up certain power related setting for gem5.

setrofim: OK, it seems there is a consensus that this is too specific to a particular script to be generally useful, and there isn't much more that makes sense beyond the existing `gem5args`. I'm closing this pull.

setrofim:
Closing, as power management configuration is too tied to user-provided gem5 configuration scripts for there to be a useful generic configuration interface in devlib.

PR #157, opened 2017-08-07
Adds additional Android methods

marcbonnici:


setrofim:
Looks good, however,

* The APIs should be documented
* There should be a corresponding getter associated with each setter.

derkling:
As a generic comment, we have similar command in LISA which we was planning to port to devlib.
Can you cross check the implementations to ensure we are using the best API for each of the operations you are proposing?

Our calls are implemented in:
https://github.com/ARM-software/lisa/blob/master/libs/utils/android/system.py
https://github.com/ARM-software/lisa/blob/master/libs/utils/android/screen.py

For example, the screen unlock call implemented in:
https://github.com/ARM-software/lisa/blob/master/libs/utils/android/screen.py#L152
seems more simple than the swipe to unlock you have. I'm wondering if sending the keycodes we are using is just enough in general, considering we tested on pixel phones without a password or unlock sequence of course.

marcbonnici:
From initial testing of both methods I have not noticed any differences between the 2 so far. For my implementations I based them off of the Android Settings provider as detailed here: https://developer.android.com/reference/android/provider/Settings.System.html
However I was not aware of the API that the LISA implementations use, do you have a link to more information about it? 

For the screen unlock, although the LISA implementation is simpler, it appears to only work on devices with stock lock screens. I've tested this with a Samsung and OnePlus device, both of which have non standard lock screens and the LISA version to fails to unlock.

derkling:
Well done, in that case it would be nice if, once this has been merged, you can try to update the LISA API to use these new calls whenever possible. This will reduce code duplication and improve consistency of LISA wrappers on top of what devlib provides.... in certain cases it could be likely possible to replace completely LISA's APIs with corresponding devlib's ones.

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+
+    :class:`AndroidTarget` is a subclass of :class:`Target` with additional features specific to a device running Android.
+
+    :param conn_cls: This is the type of connection that will be used to communicate
+        with the device. The majority of Android devices will use an ADB connection.
+
+    :param package_data_directory: This is the location of the data stored for
+        installed Android packages on the device.
+
+.. method:: Target.set_rotation(rotation)
```
setrofim: Should be `AndroidTarget` rather than `Target` here (similar below).

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+
+   Sets the screen orientation of the device as an integer value
+   representing the screen orientations. ``Natural : 0``, ``Rotated Left: 1``,
+   ``Inverted : 2`` and ``Rotated Right : 3``.
```
setrofim: Should be more explicit about what `param` values should/can be. E.g. can i specify the text form just the number? Also only  use monospace for literals and identifiers.

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+
+.. method:: Target.get_rotation(rotation)
+
+   Returns an integer value representing the orientation of the devices screen.
+   ``0 : Natural``, ``1 : Rotated Left``,
```
setrofim: Why a new line here?

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+   otherwise.
+
+.. method:: Target.homescreen()
+
+   Returns the device to its home screen.
+
+.. method:: Target.swipe_to_unlock(direction=None)
+
+   Performs a swipe input on the device to try and unlock the device.
+   A direction of `horizontal` or `vertical` can be supplied to specify in
```
setrofim: Double garve accents for monspaces, also since these are literal strings, rather than identifiers, they should be quoted.

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+   Returns the devices current screen brightness as a percentage.
+
+.. method:: Target.get_auto_brightness()
+
+   Returns ``True`` if the targets auto brightness is currently enabled and
+   ``False`` otherwise.
+
+.. method:: Target.set_auto_brightness(auto_brightness)
+
+   Specify whether the devices auto brightness is currently enabled.
```
setrofim: Should be more explicit about the the values for `auto_brightness`

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+   Returns ``True`` if the target is currently in airplane mode and ``False``
+   otherwise.
+
+.. method:: Target.get_brightness()
+
+   Returns the devices current screen brightness as an integer between 0 and 255.
+
+.. method:: Target.set_brightness(value)
+
+   Sets the devices screen brightness as an integer between 0 and 255 to the
```
setrofim: Literals should be monospaced.

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+   .. note:: Requires the device to be rooted if the device is running Android 7+.
+
+.. method:: Target.get_airplane_mode()
+
+   Returns ``True`` if the target is currently in airplane mode and ``False``
+   otherwise.
+
+.. method:: Target.get_brightness()
+
+   Returns the devices current screen brightness as an integer between 0 and 255.
```
setrofim: Literals should be monospaced.

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+   ``0 : Natural``, ``1 : Rotated Left``,
+   ``2 : Inverted`` and ``3 : Rotated Right``.
+
+.. method:: Target.set_natural_rotation()
+
+   Sets the screen orientation of the device to its natural orientation.
+
+.. method:: Target.set_left_rotation()
+
+   Sets the screen orientation of the device to its 90 deg orientation.
```
setrofim: Do not abbreviate "degrees" here (similar below).

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+
+
+Android Target
+---------------
+
+.. class:: AndroidTarget(connection_settings=None, platform=None, working_directory=None, executables_directory=None, connect=True, modules=None, load_default_modules=True, shell_prompt=DEFAULT_SHELL_PROMPT, conn_cls=AdbConnection, package_data_directory="/data/data")
+
+    :class:`AndroidTarget` is a subclass of :class:`Target` with additional features specific to a device running Android.
+
+    :param conn_cls: This is the type of connection that will be used to communicate
```
setrofim: Why document this here? It's the same as in `Target`.

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+.. method:: Target.get_brightness()
+
+   Returns the devices current screen brightness as an integer between 0 and 255.
+
+.. method:: Target.set_brightness(value)
+
+   Sets the devices screen brightness as an integer between 0 and 255 to the
+   specified value.
+
+.. method:: Target.get_brightness()
```
setrofim: Re-arrange the entries so that the the getters immediately follow/proceed (be consistent with existing docs, if applicable) the corresponding setters; rather than listing all setters followed by all getters.

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+   Specify whether the devices auto-rotation should be enabled.
+
+.. method:: Target.get_auto_rotation()
+
+   Returns ``True`` if the targets auto rotation is currently enabled and
+   ``False`` otherwise.
+
+.. method:: Target.set_airplane_mode(mode)
+
+   Specify whether the device should be in airplane mode.
```
setrofim: Be more explicit about the expected values for ``mode``

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+   Sets the screen orientation of the device to its inverted (180 deg)
+   orientation.
+
+.. method:: Target.set_right_rotation()
+
+   Sets the screen orientation of the device to its 270 deg orientation.
+
+.. method:: Target.set_auto_rotation(autorotate)
+
+   Specify whether the devices auto-rotation should be enabled.
```
setrofim: Be more explicit abou the expected values for `autorotate`

doc/target.rst
setrofim: I would also add `"diagonal"` as a valid value here and set it as the default (obviously, this relates to the actual function as well as the docs).

doc/target.rst
```diff
@@ -450,3 +450,112 @@ Target
...
+   Returns the devices current screen brightness as an integer between 0 and 255.
+
+.. method:: Target.set_brightness(value)
+
+   Sets the devices screen brightness as an integer between 0 and 255 to the
+   specified value.
+
+.. method:: Target.get_brightness()
+
+   Returns the devices current screen brightness as a percentage.
```
setrofim: Be more explicit here, e.g. 

> Returns an integer between `0` and `100`, representing screen brightness as a percentage.

doc/target.rst
marcbonnici: Added to ``Target`` documentation instead.

doc/target.rst
marcbonnici: Changed.

doc/target.rst
marcbonnici: Fixed

marcbonnici:
Fixed remaining comments.

PR #159, opened 2017-08-14
module/cgroups: robustify task freezer

qperret:
The `set()` method of the CGroup class used to freeze tasks relies on target's `write_value()`. Sometimes, the freezing procedure takes some time and the call to `write_value()` in `set()` fails by reading `FREEZING` while it expected `FROZEN` (reported in issue #136). To avoid this issue, this PR allows `write_value()` to retry the check several times before raising an exception.

### DIFF
--- a/devlib/module/cgroups.py
+++ b/devlib/module/cgroups.py
@@ -466,11 +466,11 @@ class CgroupsModule(Module):
         if freezer is None:
             raise RuntimeError('freezer cgroup controller not present')
         freezer_cg = freezer.cgroup('/DEVLIB_FREEZER')
-        thawed_cg = freezer.cgroup('/')
+        cmd = 'cgroups_freezer_set_state {{}} {}'.format(freezer_cg.directory)
 
         if thaw:
             # Restart froozen tasks
-            freezer_cg.set(state='THAWED')
+            freezer.target._execute_util(cmd.format('THAWED'), as_root=True)
             # Remove all tasks from freezer
             freezer.move_all_tasks_to('/')
             return
@@ -482,7 +482,7 @@ class CgroupsModule(Module):
         tasks = freezer.tasks('/')
 
         # Freeze all tasks
-        freezer_cg.set(state='FROZEN')
+        freezer.target._execute_util(cmd.format('FROZEN'), as_root=True)
 
         return tasks
 
--- a/devlib/bin/scripts/shutils.in
+++ b/devlib/bin/scripts/shutils.in
@@ -177,6 +177,23 @@ cgroups_tasks_in() {
 	exit 0
 }
 
+cgroups_freezer_set_state() {
+    STATE=${1}
+    SYSFS_ENTRY=${2}/freezer.state
+
+    # Set the state of the freezer
+    echo $STATE > $SYSFS_ENTRY
+    
+    # And check it applied cleanly
+    for i in `seq 1 10`; do
+        [ $($CAT $SYSFS_ENTRY) = $STATE ] && exit 0
+        sleep 1
+    done
+
+    # We have an issue
+    echo "ERROR: Freezer stalled while changing state to \"$STATE\"." >&2
+    exit 1
+}
 
 ################################################################################
 # Main Function Dispatcher
@@ -212,6 +229,9 @@ cgroups_tasks_move)
 	;;
 cgroups_tasks_in)
 	cgroups_tasks_in $*
+	;;
+cgroups_freezer_set_state)
+	cgroups_freezer_set_state $*
 	;;
 ftrace_get_function_stats)
     ftrace_get_function_stats


devlib/module/cgroups.py
```diff
@@ -297,6 +292,14 @@ def add_tasks(self, tasks):
     def add_proc(self, pid):
         self.target.write_value(self.procs_file, pid, verify=False)
 
+    def _attr_path(self, entry):
```
setrofim: Rename to `build_attr_path` and get rid of the subsequent comment.

- Method names should generally be verbs
- Since this is invoked outside the class implementation, it's not an internal method, and should not start with a `_`.

qperret: Done

setrofim:
:+1:

qperret:
Any thoughts @derkling ? It's not exactly what you suggested in #137 put I think it does the job.

derkling:
I still find this proposal a bit overkilling. I'm not really sure there are so many sysfs files which behave like the freezer controller... and if that's the case, extending the devlib's sysfs write method just to cover a single use-case it's adding maintenance burden in an API with a feature which is seldom used.

IOW, I still think that a dedicated shutil call is a better solution for the specific issue you are targeting with this task.

qperret:
I see your point but I think it doesn't hurt to have better support for things such as posted-writes in `write_value()` anyway. It seems like a relatively common semantics and it's maybe just missing in the API. That said, I agree that I don't see any other _short-term_ usage than the freezer controller. 

How do you envision a shutil version ? Something with two freezer-specific calls, one to freeze and one to thaw ? Or a more general cgroup sysfs setter kind of thing ?

setrofim:
My knowledge of the kernel source is no where near as extensive as yours, so I was happy to defer to your judgment on the best way of implementing this. However, as there appears to be some disagreement regarding this, I would like to clarify

> It seems like a relatively common semantics and it's maybe just missing in the API. That said, I agree that I don't see any other short-term usage than the freezer controller.

What are you basing this on? Could you give some examples of instances in the sys/proc/debugfs where this might come in useful (even if only in the long-term)? I agree that this *seems* like it might be a relatively common thing, but without evidence to support this feeling, I would be reluctant to add complexity to a common API.

I'm not actually opposing this change, per se; it will have no impact on existing code, and seems reasonable enough. However, I agree with @derkling  in being hesitant to handle what might turn out to be a one-off special case inside a core API, rather than the API dealing with that particular case.

setrofim:
> The thing is, I generally prefer fixes targeting the root cause of an issue (not having any form of support for things such as posted-writes here) over too specific things. 

I agree with attitude of dealing with root causes rather than slapping on cosmetic fixes. However, I question whether this is indeed the root cause there.  Are immediate-return/delayed-update writes accepted possible semantics for sysfs nodes, rather than an idiosyncrasy of a particular kernel module module?

> If (and I understand the problem is this if) we need that in 6 months for something different and this has been hidden into a shutil call, it will probably be re-implemented then because we (or at least I) will forget about that shutil call in 2 weeks.

The flip side of this is that pro-actively handling particularities of specif drivers/modules inside core APIs, as if they were systemic issues, will result in in said APIs growing increasingly bloated and unmanageable. It is true that this change is not particularly large. The next one won't be the either; and the one after that. But the sum total will result in an unwieldy API and a difficult to maintain implementation.

Hence my request for evidence that the problem being solved here is a systemic issue rather than an idiosyncrasy of a particular kernel module.

qperret:
There is no rule (that I know of) preventing you from implementing some sort of transient states or delayed writes in your sysfs if you want to, and that's the whole issue I suppose. I don't know if any other module implement that. I don't see any other at least. But given the number of them and the absence of rule, I wouldn't be surprised to find one, eventually, one day, maybe.

However, I do understand your concern regarding the bloating APIs, especially since _you_ are in charge of maintaining them :). As I said, I'm OK with the shutil version if you think it's preferable. How do you see it: a freezer specific pair of calls to freeze/thaw in `CgroupModule.freeze()` or a more generic cgroup sysfs setter to replace `Cgroup.set()` ?

setrofim:
> How do you see it: a freezer specific pair of calls to freeze/thaw in CgroupModule.freeze() or a more generic cgroup sysfs setter to replace Cgroup.set() ?

Since the freezer is the only know instance where this is necessary, I would suggest the former.

qperret:
I added a dedicated `cgroups_freezer_set_state` shutil call which is now used in `CgroupModule.freeze()`. It's ultra simple but I think it does the job. Let me know if this is what you expected.

setrofim:
:+1:

qperret:
Cool :smiley: ! @derkling: Are you OK with this implementation then ?

derkling:
I :heart: it better now, thanks @qperret and :+1: from my side as well! ;-)

PR #160, opened 2017-08-16
module/gem5stats: enable asynchronous origins in gem5 stats file

qperret:
Currently, reset_origin() in the gem5stats module enables to virtually
truncate the existing dumps from the gem5 statistics file so that any
subsequent match() calls will apply only on new dumps. However, the
current implementation resets the origin of the stats file unilaterally,
without other clients knowing about it, hence leading to data being
lost.

This commits removes the reset_origin() method and provides a different
API to handle virtual truncatures in the stats file. Namely, the match()
method now takes a base_dump parameter letting clients specifiying from
which dump they want match() to apply. This feature could also be usefull
if someone wanted to add off-line processing features for statistics
files.

### DIFF
--- a/devlib/instrument/gem5power.py
+++ b/devlib/instrument/gem5power.py
@@ -52,6 +52,7 @@ class Gem5PowerInstrument(Instrument):
         self.target.gem5stats.book_roi(self.roi_label)
         self.sample_period_ns = 10000000
         self.target.gem5stats.start_periodic_dump(0, self.sample_period_ns)
+        self._base_stats_dump = 0
 
     def start(self):
         self.target.gem5stats.roi_start(self.roi_label)
@@ -64,11 +65,12 @@ class Gem5PowerInstrument(Instrument):
         with open(outfile, 'wb') as wfh:
             writer = csv.writer(wfh)
             writer.writerow([c.label for c in self.active_channels]) # headers
-            for rec, rois in self.target.gem5stats.match_iter(active_sites, [self.roi_label]):
+            for rec, rois in self.target.gem5stats.match_iter(active_sites, 
+                    [self.roi_label], self._base_stats_dump):
                 writer.writerow([float(rec[s]) for s in active_sites])
         return MeasurementsCsv(outfile, self.active_channels)
     
     def reset(self, sites=None, kinds=None, channels=None):
         super(Gem5PowerInstrument, self).reset(sites, kinds, channels)
-        self.target.gem5stats.reset_origin()
+        self._base_stats_dump = self.target.gem5stats.next_dump_no()
 
--- a/devlib/module/gem5stats.py
+++ b/devlib/module/gem5stats.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
 import logging
 import os.path
 from collections import defaultdict
@@ -66,6 +67,7 @@ class Gem5StatsModule(Module):
         self._stats_file_path = os.path.join(target.platform.gem5_out_dir,
                                             'stats.txt')
         self.rois = {}
+        self._dump_pos_cache = {0: 0}
 
     def book_roi(self, label):
         if label in self.rois:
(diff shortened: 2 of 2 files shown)


devlib/instrument/gem5power.py
```diff
@@ -64,11 +65,12 @@ def get_data(self, outfile):
         with open(outfile, 'wb') as wfh:
             writer = csv.writer(wfh)
             writer.writerow([c.label for c in self.active_channels]) # headers
-            for rec, rois in self.target.gem5stats.match_iter(active_sites, [self.roi_label]):
+            for rec, rois in self.target.gem5stats.match_iter(active_sites, \
```
setrofim: `\` unnecessary

devlib/module/gem5stats.py
```diff
@@ -103,10 +105,10 @@ def start_periodic_dump(self, delay_ns=0, period_ns=10000000):
             raise ValueError(msg.format(delay_ns, period_ns))
         self.target.execute('m5 dumpresetstats {} {}'.format(delay_ns, period_ns))
     
-    def match(self, keys, rois_labels):
+    def match(self, keys, rois_labels, base_dump=0):
         '''
         Tries to match the list of keys passed as parameter over the statistics
-        dumps covered by selected ROIs since origin. Returns a dict indexed by 
+        dumps covered by selected ROIs since base_dump. Returns a dict indexed by
```
setrofim: Add double grave accents around `base_dump` to mark it as an identifier (similar in the docstring below).

devlib/module/gem5stats.py
```diff
@@ -66,6 +67,7 @@ def __init__(self, target):
         self._stats_file_path = os.path.join(target.platform.gem5_out_dir,
                                             'stats.txt')
         self.rois = {}
+        self._dump_pos_cache = {0: 0}
```
setrofim: Use a list rather than a dict here? That would make the logic on lines 176-177 simpler (and probably faster).

qperret: Actually having a dict makes line 173 very simple to implement. Since a long run will involve a lot (thousands ?) of dumps not all of them are cached in `_dump_pos_cache`. For example, I will end-up with something like 
```
self._dump_pos_cache = { 
    0: 0,
    4: 123456,
    19: 7891234
}
```

if use `_goto_dump` on dumps 4 and 19. Putting this in a list will make things more difficult since the Xth dump in the list is not necessarily be the Xth dump in the file. If we really want to use a list, we probably want to cache each and every dump by modifying line 191 and putting it into the `while` loop. I you think it's better I'm OK with it, the only concern is maybe the memory overhead.

qperret:
Fixed commit message typo, docstrings and removed unecessary `\`.

devlib/module/gem5stats.py
setrofim: OK, didn't realize there could be gaps. In which case, I would suggest using `OrderedDict` instead.

qperret: Ah yes we can probably find a better data-structure. Actually, the order in which keys are inserted depends on how `match()` is called by the user. So we can't really assume they will be inserted in increasing order and I think this is what we need to speed-up lines 176:177 :confused:. What might be usefull is some sort of key-ordered dict here instead of a simple OrderedDict. Do you know if there is a standard python data-structure doing that ? Most likely, we need some sort of tree-based hash-map, otherwise complexity will be :-1:

setrofim: Mmm ok. Yeah, if we can't assume the keys are inserted in incremental order, then something like a red-black tree would be the way to go. There is nothing like that in the standard lib. There is a `bintrees` package that provides an implementation. However, I'd rather hold-off on introducing an additional dependency until/unless lines 176:177 prove to be a bottle neck. I guess, we can leave them as-is for now.

qperret: I don't think python provides that by default. This is what we need:
https://github.com/grantjenks/sorted_containers
Not sure if it's a good idea to add external dependencies to devlib ...

qperret:
Anything left preventing from merging ?

setrofim:
Since this is gem5-related, I'd like ether @Sticklyman1936 or @avanlaer  to OK this before merging.

Sticklyman1936:
This looks good to me. Seems like a nice way to deal with this issue. Would be interesting to see how well this sort of approach will work if we ever figure out how to deal with the gzipped stats files. However, this seems perfect for the time being. Thanks a lot for doing this!

PR #161, opened 2017-08-16
module/gem5stats: enhance match() with regex support

qperret:
The current implementation of match() in the gem5stats module returns
records matching exactly the specified keys. This PR changes this
behaviour by matching keys over regular expressions, hence resulting in
a much more powerful match() implementation.

### DIFF
--- a/devlib/module/gem5stats.py
+++ b/devlib/module/gem5stats.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import re
 import sys
 import logging
 import os.path
@@ -139,6 +140,10 @@ class Gem5StatsModule(Module):
                 self.logger.warning('Trying to match records in statistics file'
                         ' while ROI {} is running'.format(label))
         
+        # Construct one large regex that concatenates all keys because
+        # matching one large expression is more efficient than several smaller
+        all_keys_re = re.compile('|'.join(keys))
+        
         def roi_active(roi_label, dump):
             roi = self.rois[roi_label]
             return (roi.field in dump) and (int(dump[roi.field]) == 1)
@@ -148,8 +153,8 @@ class Gem5StatsModule(Module):
             for dump in iter_statistics_dump(stats_file):
                 active_rois = [l for l in rois_labels if roi_active(l, dump)]
                 if active_rois:
-                    record = {k: dump[k] for k in keys}
-                    yield (record, active_rois)
+                    rec = {k: dump[k] for k in dump if all_keys_re.search(k)}
+                    yield (rec, active_rois)
 
     def next_dump_no(self):
         '''
--- a/devlib/module/gem5stats.py
+++ b/devlib/module/gem5stats.py
@@ -108,14 +108,55 @@ class Gem5StatsModule(Module):
     
     def match(self, keys, rois_labels, base_dump=0):
         '''
-        Tries to match the list of keys passed as parameter over the statistics
-        dumps covered by selected ROIs since ``base_dump``. Returns a dict 
-        indexed by key parameters containing a dict indexed by ROI labels 
-        containing an in-order list of records for the key under consideration
-        during the active intervals of the ROI.
+        Extract specific values from the statistics log file of gem5
 
-        Keys must match fields in gem5's statistics log file. Key example:
-            system.cluster0.cores0.power_model.static_power
+        :param keys: a list of key name or regular expression patterns that
+            will be matched in the fields of the statistics file. ``match()``
+            returns only the values of fields matching at least one these
+            keys.
+        :type keys: list
+
+        :param rois_labels: list of ROIs labels. ``match()`` returns the 
+            values of the specified fields only during dumps spanned by at
+            least one of these ROIs.
+        :type rois_label: list
+
+        :param base_dump: dump number from which ``match()`` should operate. By 
+            specifying a non-zero dump number, one can virtually truncate 
+            the head of the stats file and ignore all dumps before a specific
+            instant. The value of ``base_dump`` will typically (but not 
+            necessarily) be the result of a previous call to ``next_dump_no``.
+            Default value is 0.
+        :type base_dump: int
+
+        :returns: a dict indexed by key parameters containing a dict indexed by
+        ROI labels containing an in-order list of records for the key under
+        consideration during the active intervals of the ROI. 
+        
+        Example of return value:
+         * Result of match(['sim_'],['roi_1']):
+            {
+                'sim_inst': 
+                {
+                    'roi_1': [265300176, 267975881]
+                }
+                'sim_ops': 
+                {
+                    'roi_1': [324395787, 327699419]
+                }
+                'sim_seconds': 
+                {
+                    'roi_1': [0.199960, 0.199897]
+                }
+                'sim_freq': 
+                {
+                    'roi_1': [1000000000000, 1000000000000]
+                }
+                'sim_ticks': 
+                {
+                    'roi_1': [199960234227, 199896897330]
+                }
+            }
         '''
         records = defaultdict(lambda : defaultdict(list))
         for record, active_rois in self.match_iter(keys, rois_labels, base_dump):
@@ -126,12 +167,27 @@ class Gem5StatsModule(Module):
 
     def match_iter(self, keys, rois_labels, base_dump=0):
         '''
-        Yields for each dump since ``base_dump`` a pair containing:
-        1. a dict storing the values corresponding to each of the specified keys
-        2. the list of currently active ROIs among those passed as parameters.
+        Yield specific values dump-by-dump from the statistics log file of gem5
 
-        Keys must match fields in gem5's statistics log file. Key example:
-            system.cluster0.cores0.power_model.static_power
+        :param keys: same as ``match()``
+        :param rois_labels: same as ``match()``
+        :param base_dump: same as ``match()``
+        :returns: a pair containing:
+            1. a dict storing the values corresponding to each of the found keys
+            2. the list of currently active ROIs among those passed as parameters
+
+        Example of return value:
+         * Result of match_iter(['sim_'],['roi_1', 'roi_2']).next()
+            ( 
+                { 
+                    'sim_inst': 265300176,
+                    'sim_ops': 324395787,
+                    'sim_seconds': 0.199960, 
+                    'sim_freq': 1000000000000,
+                    'sim_ticks': 199960234227,
+                },
+                [ 'roi_1 ' ] 
+            )
         '''
         for label in rois_labels:
             if label not in self.rois:
--- a/devlib/utils/gem5.py
+++ b/devlib/utils/gem5.py
@@ -13,12 +13,17 @@
 # limitations under the License.
 
 import re
+import logging
+
+from devlib.utils.types import numeric
 
 
 GEM5STATS_FIELD_REGEX = re.compile("^(?P<key>[^- ]\S*) +(?P<value>[^#]+).+$")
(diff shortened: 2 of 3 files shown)


devlib/module/gem5stats.py
```diff
@@ -137,17 +138,21 @@ def match_iter(self, keys, rois_labels):
...
         
+        # Construct one large regex that concatenates all keys because
+        # matching one large expression is more efficient than several smaller
+        all_keys_re = re.compile('|'.join(keys))
+        
         def roi_active(roi_label, dump):
             roi = self.rois[roi_label]
             return (roi.field in dump) and (int(dump[roi.field]) == 1)
-
+
```
Sticklyman1936: please remove the whitespace change :)

qperret:
Fixed whitespace and rebased to new master

devlib/module/gem5stats.py
```diff
@@ -148,8 +153,8 @@ def roi_active(roi_label, dump):
             for dump in iter_statistics_dump(stats_file):
                 active_rois = [l for l in rois_labels if roi_active(l, dump)]
                 if active_rois:
-                    record = {k: dump[k] for k in keys}
-                    yield (record, active_rois)
+                    rec = { k: dump[k] for k in dump if all_keys_re.search(k) }
```
setrofim: No spaces inside braces

qperret:
Refixed the whitespace (I went too fast the first time, sorry :smile: )

devlib/module/gem5stats.py
qperret: Done

setrofim:
Please update the docstrings for `match()` and `match_iter()` to reflect that now keys can be regex's.

qperret:
`match()` and `match_iter()` should now have better documentation.
I also re-tested the whole thing to check that the quick fixes I did earlier didn't break anything and it all looks good :)

devlib/module/gem5stats.py
```diff
@@ -107,14 +108,53 @@ def start_periodic_dump(self, delay_ns=0, period_ns=10000000):
...
-        dumps covered by selected ROIs since ``base_dump``. Returns a dict 
-        indexed by key parameters containing a dict indexed by ROI labels 
-        containing an in-order list of records for the key under consideration
-        during the active intervals of the ROI.
-
-        Keys must match fields in gem5's statistics log file. Key example:
-            system.cluster0.cores0.power_model.static_power
+        Extract specific values from the statistics log file of gem5
+
+        :param keys: list of regex matched over the fields of the statistics
```
setrofim: " a list of key name or regular expression patterns that will be matched in the fields of the statistics"

devlib/module/gem5stats.py
```diff
@@ -107,14 +108,53 @@ def start_periodic_dump(self, delay_ns=0, period_ns=10000000):
...
+        :returns: a dict indexed by key parameters containing a dict indexed by
+        ROI labels containing an in-order list of records for the key under
+        consideration during the active intervals of the ROI. 
+        
+        Example of return value:
+         * Result of match(['sim_'],['roi_1']):
+            {
+                'sim_inst': 
+                {
+                    'roi_1': ['265300176', '267975881']
```
setrofim: Here values appear to be strings. Is this correct? If so, then the implementation should be changed to cast this to `devlib.utils.types.numeric`.

devlib/module/gem5stats.py
```diff
@@ -107,14 +108,53 @@ def start_periodic_dump(self, delay_ns=0, period_ns=10000000):
...
+            file. match() returns only the values of fields matching at least
+            one these regex.
+        :type keys: list
+
+        :param rois_labels: list of ROIs labels. match() returns the values of
+            the specified fields only during dumps spanned by at least of
+            of these ROIs.
+        :type rois_label: list
+
+        :param base_dump: dump number from which match() should operate. By
```
setrofim: "the dump number from which \`\`match()`` should operate."

devlib/module/gem5stats.py
```diff
@@ -107,14 +108,53 @@ def start_periodic_dump(self, delay_ns=0, period_ns=10000000):
...
+        :param rois_labels: list of ROIs labels. match() returns the values of
+            the specified fields only during dumps spanned by at least of
+            of these ROIs.
+        :type rois_label: list
+
+        :param base_dump: dump number from which match() should operate. By 
+            specifying a non-zero dump number, one can virtually truncate 
+            the head of the stats file and ignore all dumps before a specific
+            instant. The value of ``base_dump`` will typically (but not 
+            necessarily) be the result of a previous call to ``next_dump_no``.
```
setrofim: Should specify that defaults to zero

devlib/module/gem5stats.py
qperret: Hmm good point. The current implementation returns only strings (or sets of strings). It looks like the values in the stats file are printed in many different formats, but most of them can be converted easily using `devlib.utils.types.numeric`. However, for the others, it might be more tricky. For example, some stats values are percentages, and they usually include the `%` symbol which will cause issues during the cast. I'm not sure if this is the only problematic case and if we can safely assume anything about the format of values dumped in the stats file.

@Sticklyman1936 , @avanlaer : are we guaranteed to have numeric values only in the stats file ? Is the `%` symbol the only corner case ?

devlib/module/gem5stats.py
qperret: Done

devlib/module/gem5stats.py
qperret: Done

devlib/module/gem5stats.py
qperret: Done

devlib/module/gem5stats.py
setrofim: Regarding the `'%'` symbol, the `numeric()` function can be extended to handle that: https://github.com/ARM-software/devlib/pull/162

qperret: I think it's a good idea to manage the `%` symbol directly in `numeric()`. Thanks for the PR !

devlib/utils/gem5.py
```diff
@@ -39,5 +46,6 @@ def iter_statistics_dump(stats_file):
             if res:
                 k = res.group("key")
                 v = res.group("value").split()
+                v = map(best_effort_numeric, v)
```
setrofim: After an offline discussion with @Sticklyman1936, gem5 stats values will always be numeric (internally they're all represented as doubles), so best effort is not needed here.

We should assume that they are numeric, if, for some reason, the conversion fails, we do want that to be raised as an error (as that would indicate that something went wrong during stats collection) and not have it "swallowed" here.

However, we do want to make sure that, in such an event, the entire run should not be kills, so we do want to add a try/except guard to catch and log the error.

Not sure if it's best to do it here or further up the call stack?

qperret: I think it makes sense to raise an exception here. We care about the format of the stats file in this function only so if there is a formatting issue it should be raised here. I can replace line 49 with `v = map(numeric, v)` and remove `best_effort_numeric`. Is that OK for you ? If you think it's necessary I can also `try ... catch` the cast to raise a more readable exception.
And, for safety, we should probably wait for  #162 to be merged before merging this one.

setrofim: Hm, yeah, maybe a `TargetError` with something like "encountered non-numeric gem5 stat: \<blah>" would be appropriate here, and then let the client code figure out the best way of handling that...

@Sticklyman1936 , @avanlaer ?

qperret: I missed your 3rd paragraph ...
What about modifying the `catch` part of `best_effort_numeric` to log the issue before returning the value ?

setrofim: Having had a look through the code, it seems to me that it would make more sense to propagate the error to the calling code, rather than swallow it here. However I'm not that familiar with gem5 stats and how they are typically processed, so I'll leave the decision on propagate vs catch-and-report to @Sticklyman1936  and @avanlaer.

qperret:
Rebased on new master.
@Sticklyman1936, @avanlaer any thoughts about this ?

avanlaer:
In my opinion (but want to see what @Sticklyman1936 thinks), I think it is better to catch-and-report rather than propagating. If there is a difference in stat format, you want this to be mentioned to the user with bells and whistles but not necessarily kill the simulation. If you don't properly close the simulation and make it 'crash', you sometimes lose the whole stats file (and hence the complete run, which can be painful) in the case of a compressed stats file, because the stats file is no longer properly compressed and hence readable.

Also, I think (sorry, cannot find the code to support this now) in the case of a TargetError, devlib tries to gather as much info as possible (screencap, logcat etc) which in the case of gem5 is incredibly slow and hence not very practical.

setrofim:
> Also, I think (sorry, cannot find the code to support this now) in the case of a TargetError, devlib tries to gather as much info as possible (screencap, logcat etc) which in the case of gem5 is incredibly slow and hence not very practical.

This happens in WA, rather than devlib.

Sticklyman1936:
Personally, I would opt for roughly the same solution as Anouk suggested. I would catch the exception here, and actually downgrade it to a warning. If there really is an issue with parsing the stats file, then this doesn't mean that one should kill the entire simulation as this most likely stems from a parsing error. One can always re-parse the stats offline, if really required.

Even if we do decide to propagate the error upwards here, we want to make sure it is caught and that the gem5 simulation is shut down as gracefully as possible.

qperret:
Yes, I agree that loosing the whole run for one malformed stats entry is probably not what we want. I can change the `except` part of `best_effort_numeric` to log a warning before returning the raw value. However I'm not sure what's the best way of implementing that inside `iter_statistics_dump` which is kind of self-contained right now. I could maybe use the `warnings` package here ? I'm not sure if this would be devlib-friendly though. Or I could give a `logger` argument to `iter_statistics_dump` ? @setrofim, what do you think ?

setrofim:
> I can change the except part of best_effort_numeric to log a warning before returning the raw value. 

No. The values must be numbers and only numbers. If you're unable to parse the value as part of the number then you report it in the warning, but you do no return it. Lines 46-50 should be replaced with something along the lines of:

```
             if res:
                 k = res.group("key")
                 vtext = res.group("value")
                 try:
                     v = map(numeric, vtext.split())
                     cur_dump[k] = v[0] if len(v)==1 else set(v)
                 except ValueError:
                     msg = 'Could not parse value for "{}": "{}"'
                     logger.error(msg.format(k, vtext))

```
                     

>  Or I could give a logger argument to iter_statistics_dump

Don't pass as an argument. Initialize the logger at the top level in this module:

```
import logging

logger = logging.getLogger('gem5')
```

qperret:
Updated the implementation of `iter_statistics_dump`. Let me know if it's OK.

devlib/utils/gem5.py
```diff
@@ -38,6 +43,11 @@ def iter_statistics_dump(stats_file):
             res = GEM5STATS_FIELD_REGEX.match(line) 
             if res:
                 k = res.group("key")
-                v = res.group("value").split()
-                cur_dump[k] = v[0] if len(v)==1 else set(v)
+                vtext = res.group("value").split()
```
setrofim: `split()` inside the `map` below, rather than here; so that if/when you log the error in the `except` clause, you will be logging the original value, and not the split list.

qperret: Done.

qperret:
Thanks for the review guys! I re-tested this today and it looks good. Is there anything I should add/fix before merging ?

setrofim:
I'm OK with current pull request. Will merge once either @Sticklyman1936 or @avanlaer :+1:  as well.

PR #162, opened 2017-08-16
utils/types: numeric expeanded to handle percentages

setrofim:
numeric() conversion function is expanded to handle percentage values in
the form "10.123%". The result is the numeric part converted to a float
and divided by 100, e.g. 0.10123.

### DIFF
--- a/devlib/utils/types.py
+++ b/devlib/utils/types.py
@@ -68,6 +68,15 @@ def numeric(value):
     """
     if isinstance(value, int):
         return value
+
+    if isinstance(value, basestring):
+        value = value.strip()
+        if value.endswith('%'):
+            try:
+                return float(value.rstrip('%')) / 100
+            except ValueError:
+                raise ValueError('Not numeric: {}'.format(value))
+
     try:
         fvalue = float(value)
     except ValueError:


devlib/utils/types.py
```diff
@@ -68,6 +68,15 @@ def numeric(value):
     """
     if isinstance(value, int):
```
qperret: What about adding `or isinstance(value, long) or isinstance(value, float)` ?

devlib/utils/types.py
```diff
@@ -68,6 +68,15 @@ def numeric(value):
...
         return value
+
+    if isinstance(value, basestring):
+        value = value.strip()
+        if value.endswith('%'):
+            try:
+                return float(value.rstrip('%')) / 100
```
qperret: Should we do that in a more generic way ? We could do something like
```python
corrector = 1.0
if value.endswith('%'):
    value = value.rstrip('%')
    corrector = 0.01
# [ ... rest of the function ... ]
return fvalue*corrector
```
I think the code will be easier to read and it can be extended with no effort to account for other corner cases (_pre-mille_ for example: ‰)

qperret: The `int` / `float` thing below can be replaced by a single call to `ast.literal_eval()` which will return an `int` or a `float` when appropriate.

setrofim: If we ran into a percentage we don't want to drop through to the rest of the function after striping away the `%`. Say we run into `"10%"` afterwards, you end up with `"10"`. 

- If the logic is implemented exactly as you've suggested, then we will be returning `10` as in that case it is the `ivalue`, not the `fvalue` that  is returned.
- If we appy `corrector` to the 'ivalue' we always end up with a float.

The `corrector` can be changed to default to `1` rather than `1.0`, in which case we should get the correct behavior. However, that's rather subtle, and I would argue, less clear than the more explicit current approach. Additionally, if see a percentage, we will always want to return a float, therefore there is no point doing the subsequent integer conversion and comparison -- better to return right away.

Regarding factoring out the divisor into a variable -- I suggest we hold off until we're actually adding the support for per-mille or other such value...

setrofim: > The int / float thing below can be replaced by a single call to ast.literal_eval() which will return an int or a float when appropriate

Yes, be we also want to raise `ValueError` unless the string is representing an `int` or `float`. So you'd still have to do checks afterwards, and you're not winning anything.

devlib/utils/types.py
setrofim: No, because if we get `1.0`, we want to return an `1` -- this function is intended to prefer ints, so if a float value can be represented as an integer without loss of precision, it will return the corresponding int.

devlib/utils/types.py
qperret: My bad, I didn't think carefully. However we can probably refactor that a little bit using `ast.eval_literal()`. What about something like:
```python
if isinstance(value, int) or isinstance(value, float):
    return value
if isinstance(value, basestring):
    try:
        return ast.eval_literal(value)
    except:
        pass
    if value.endswith('%'):
        cvalue = value.rstrip('%')
        try:
            return ast.eval_literal(cvalue) / 100.0
        except:
            pass # Don't raise excpetion here to extend more easily with other value.endswith('whatever')
raise ValueError('Not numeric: {}'.format(value))
```

devlib/utils/types.py
qperret: Oh ok, I can see that in the docstring ... But why do we want this behaviour ?

devlib/utils/types.py
setrofim: What happens if I call that with `"[1, 2]"`?

devlib/utils/types.py
setrofim: In short, because  ℤ ⊆ ℝ, and we want to provide as "tight" a value a possible.

There are certain values that are conceptually always ints (e.g. counts), and there are analysis tools that may expect them to be such. On the other hand, there are applications which, either for simplicity or out of necessity (e.g. they're implemented in Javascript), report all numeric values as floats. Things then break when you try to combine the two.

It's also partially a usability thing, e.g. `pandas` will format ints in a nicer, more readable way.

Basically, if something expects a float and we give it an int, chances are we're not losing anything; if something expects an int and we give it a float, things break. Hence, as the default behavior,  prefer ints.

devlib/utils/types.py
qperret: Well, you'll get a list of integers which is a not too bad behaviour IMO, but I agree you have a problem if you call it with `"['a', 'b']"`. Fair enough, `ast.eval_literal()` is maybe too powerful to reduce really simplify the code in any way ...

devlib/utils/types.py
qperret: Ok, thanks, that makes sense !

devlib/utils/types.py
setrofim: > Well, you'll get a list of integers which is a not too bad behaviour IMO

The function's contract is to return a numerical value; returning anything else (even a list of numerical values) is bad. The calling code can reasonably expect to plug the output into a mathematical formula, write value into a column in a database, etc.

qperret: Yeah that's true ...  What I was thinking is that having `numeric()` working on list might be ok (if it's documented like that at least). Anyway the above code does not suffice at all for that, and yeah, the code simplification I was aiming at with `ast.eval_literal()` is actually ruined by all the additional checks we need to add to make it work properly. So forget about it, I'm OK with your implementation. Sorry for bothering you :smile:

setrofim: No worries -- it's good to have the discussion. What you are suggesting are perfectly valid semantics that are useful in certain situations. However,  extending `numeric` with them would make it unsuitable for its intended purpose. For what you're describing, WA has the `list_of` and `list_or` families of functions, e.g. https://github.com/ARM-software/workload-automation/blob/master/wlauto/utils/types.py#L111

(These have not been migrated into devlib because, as of now, there is no clear need for them in the code line; but they could be, if and when such a need arises.)

PR #163, opened 2017-08-18
Add support for AcmeCape and Derived Measurements

marcbonnici:


devlib/instrument/gem5power.py
```diff
@@ -51,6 +51,7 @@ def __init__(self, target, power_sites):
             self.add_channel(field, 'power')
         self.target.gem5stats.book_roi(self.roi_label)
         self.sample_period_ns = 10000000
+        self.sample_rate_hz = 1 / (self.sample_period_ns * 10**-9)
```
qperret: As a side note, I'm not sure if this really applies to gem5. Indeed, because of gem5 behaviour, it's not guaranteed that the `Gem5PowerInstrument` provides samples at that rate. In practice, it provides samples every 10ms  **+** additional asynchronous samples for various reasons. That's actually why the instrument always provides a `time` channel. Without it, there is no way we can figure out how much time has elapsed between two samples. So, I'm just wondering if providing `sample_rate_hz` is a good idea here. It might be misleading I suppose.

marcbonnici: That's a good point thanks, I've removed that so it will just rely on the timestamps reported. 
On a similar note, to conform with the new `DerivedMeasurements` API, to allow them to be detected the time channel needs to be renamed from "sim_seconds" to "timestamp", is this going to cause many problems?

qperret: Hmmm yes that's an issue unfortunately ... In short, `sim_seconds` is gem5-related and cannot be changed. In more detail, with the current implementation, the channel names (including the `power_sites`) of the `Gem5PowerInstrument` must match specific fields of the statistics log file of gem5. And as you probably guessed, the field that gives the time elapsed since the last sample is `sim_seconds` ... So if this is absolutely required to change the `sim_seconds` to `timestamp`, we probably need deeper modifications of `Gem5PowerInstrument` to decorelate the channel names from the fields of the stats file. Or we can hack something for the specific case of `sim_seconds` ...

devlib/instrument/gem5power.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, target, power_sites):
             self.power_sites = power_sites
         else:
             self.power_sites = [power_sites]
-        self.add_channel('sim_seconds', 'time')
+        self.add_channel('timestamp', 'time')
```
qperret: This breaks the `Gem5PowerInstrument` since `sim_second` must be extracted from the statistics file of gem5.

marcbonnici: Thanks for pointing that out. I've added a mapping for the site name that will allow for `sim_seconds` to be used when extracting the timestamp from the data which should fix that problem.

qperret: Cool, that should fix it indeed. Thanks !

PR #165, opened 2017-08-21
Added support for SSH options

AnthonyARM:


AnthonyARM:
Oops yes, sorry about that

PR #166, opened 2017-09-08
Logcat monitor

valschneider:
The code could probably do with some cleanup, but I wanted to get some feedback before investing too much time into it.

This was made due to a need to parse logcat with lines that can come out of order. Furthermore, most of the logcat parsing done in LISA was copy-pasted from one workload to another, so I wanted to centralize that.

The idea here is to have a thread that reads from logcat, and it can also look for certain regexps (by default '.*', so every line will be logged). We can then search for all the lines matching a given regexp. The general use case is:

```
# All regexps that will be used during the workload execution
my_regexps = [...]

logcat = target.logcat_monitor(my_regexps)
logcat.start()

my_workload.run()

logcat.search(WORKLOAD_START_REGEXP)
print "workload started !"

logcat.search(WORKLOAD_END_REGEXP)
print "workload ended !"

logcat.stop()
```

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -14,7 +14,7 @@ from devlib.module import get_module
 from devlib.platform import Platform
 from devlib.exception import TargetError, TargetNotRespondingError, TimeoutError
 from devlib.utils.ssh import SshConnection
-from devlib.utils.android import AdbConnection, AndroidProperties, adb_command, adb_disconnect
+from devlib.utils.android import AdbConnection, AndroidProperties, LogcatMonitor, adb_command, adb_disconnect
 from devlib.utils.misc import memoized, isiterable, convert_new_lines, merge_lists
 from devlib.utils.misc import ABI_MAP, get_cpu_name, ranges_to_list, escape_double_quotes
 from devlib.utils.types import integer, boolean, bitmask, identifier, caseless_string
@@ -1155,6 +1155,9 @@ class AndroidTarget(Target):
 
     def clear_logcat(self):
         adb_command(self.adb_name, 'logcat -c', timeout=30)
+
+    def get_logcat_monitor(self, regexps=None):
+        return LogcatMonitor(self, regexps)
 
     def adb_kill_server(self, timeout=30):
         adb_command(self.adb_name, 'kill-server', timeout)
--- a/devlib/trace/logcat.py
+++ b/devlib/trace/logcat.py
@@ -1,0 +1,58 @@
+import os
+import re
+import shutil
+
+from devlib.trace import TraceCollector
+from devlib.utils.android import LogcatMonitor
+
+class LogcatCollector(TraceCollector):
+
+    def __init__(self, target, regexps=None):
+        super(LogcatCollector, self).__init__(target)
+        self.regexps = regexps
+        self._collecting = False
+        self._prev_log = None
+
+    def reset(self):
+        """
+        Clear Collector data but do not interrupt collection
+        """
+        if not self._monitor:
+            return
+
+        if self._collecting:
+            self._monitor.clear_log()
+        elif self._prev_log:
+            os.remove(self._prev_log)
+            self._prev_log = None
+
+    def start(self):
+        """
+        Start collecting logcat lines
+        """
+        self._monitor = LogcatMonitor(self.target, self.regexps)
+        if self._prev_log:
+            # Append new data collection to previous collection
+            self._monitor.start(self._prev_log)
+        else:
+            self._monitor.start()
+
+        self._collecting = True
+
+    def stop(self):
+        """
+        Stop collecting logcat lines
+        """
+        if not self._collecting:
+            raise RuntimeError('Logcat monitor not running, nothing to stop')
+
+        self._monitor.stop()
+        self._collecting = False
+        self._prev_log = self._monitor.logfile
+
+    def get_trace(self, outfile):
+        """
+        Output collected logcat lines to designated file
+        """
+        # copy self._monitor.logfile to outfile
+        shutil.copy(self._monitor.logfile, outfile)
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -24,6 +24,9 @@ import time
 import subprocess
 import logging
 import re
+import threading
+import tempfile
+import Queue
 from collections import defaultdict
 
 from devlib.exception import TargetError, HostError, DevlibError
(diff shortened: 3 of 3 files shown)


devlib/utils/android.py
```diff
@@ -508,3 +509,89 @@ def _check_env():
...
+            regexp = '({})'.format(regexp)
+
+        self._logcat = self.device.background('logcat -e {}'.format(regexp))
+
+        while not self._stopped.is_set():
+            line = self._logcat.stdout.readline(1024)
+            self._add_line(line)
+
+    def _add_line(self, line):
+        self._datalock.acquire()
```
setrofim: Rather than doing explicit acquire/release, you can use the lock as a context:

```
with self._datalock:
       # do stuff
```

This is cleaner and will also ensure that the lock is released should an exception occur with it held.

devlib/target.py
```diff
@@ -1156,6 +1156,9 @@ def dump_logcat(self, filepath, filter=None, append=False, timeout=30):  # pylin
     def clear_logcat(self):
         adb_command(self.adb_name, 'logcat -c', timeout=30)
 
+    def logcat_monitor(self, regexps=['.*']):
```
setrofim: `get_logcat_monitor` or `monitor_logcat` -- method names should generally be verb phrases.

devlib/utils/android.py
```diff
@@ -508,3 +509,89 @@ def _check_env():
...
+class LogcatMonitor(threading.Thread):
+    def __init__(self, device, regexps=['.*']):
+        super(LogcatMonitor, self).__init__()
+
+        self.device = device
+
+        self._stopped = threading.Event()
+        self._seeking = threading.Event()
+
+        self._seeked = None
```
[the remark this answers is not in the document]
valschneider: Apologies for the English blasphemy :)

valschneider:
@setrofim Thanks for the feedback !
I wanted to add something regarding the regexp matching: my initial 'pitch' was that, since we know in advance the regexps we're going to try to find in logcat, we can feed that to the monitor and reduce the amount of lines to collect - and in fine reduce the amount of lines to go through for individual line matching.

As such, I think the regexp(s) should be a mandatory parameter. Now, there is the problem of overhead - continuously running unoptimised regexps on the device could noticeably impact the test/benchmark we're running, but I was under the impression that sending every single line via ADB to the host machine (for later regexp matching on the host) would be just as inefficient if not worse. What do you think ?

setrofim:
You're making the following assumption:

> since we know in advance the regexps we're going to try to find in logcat

This may be valid for your use case, but it won't always be in general. Logcat contains a wealth of information about the activity in the system and our strategy is often to collect all of it. Even if the primary goal is to extract specific information from it, we want the rest in cause we need to debug some issue.

> As such, I think the regexp(s) should be a mandatory parameter

It isn't in your current implementation -- you're giving it a default, so it's not mandatory for the user.

> but I was under the impression that sending every single line via ADB to the host machine (for later regexp matching on the host) 

Well, you'd be transmitting more data, but doing less on-target processing on it. Whether or not this will result in lighter or heavier processing load, I would not want to guess without doing some profiling. But this is moot -- you're missing my point.

You're defaulting the regex to `'.*'` which matches every line, so this will return everything anyway. My proposal is to, by default, not use regex. That will be functionally equivalent -- returning every line, but without the pointless matching of `'.*'` on every line.

valschneider:
> This may be valid for your use case, but it won't always be in general. Logcat contains a wealth of information about the activity in the system and our strategy is often to collect all of it. Even if the primary goal is to extract specific information from it, we want the rest in cause we need to debug some issue.

Fair enough, I'll go with logging everything by default then.

valschneider:
Here's what's new:
- No regexp matching by default
- New lines are added to a queue, and queue is flushed to a file either
   - If queue size reaches a threshold
   - Whenever a user requests data from the monitor (e.g. call to `get_log()` or `search()`)

And here are my comments:
- At this point it's pretty much an implementation of TraceCollector, so I was wondering if I shouldn't just move the code from utils/android to trace/logcat ?
- For now the log file is a tempfile that is never destroyed. I might change it a bit to fit `get_trace()` of TraceCollector - I'd prefer to avoid forcing the user to work with a file for very simply use cases (which covers a good bunch of what I'll do once this is merged).

setrofim:
As I've mentioned in an earlier comment, there should be a `TraceCollector` implementation for this. I don't have a strong opinion right now on whether it is better to turn this into a `TraceCollector` subclass directly, or whether to create a `TraceCollector` wrapper for the monitor.

> For now the log file is a tempfile that is never destroyed.

This should be done in `TraceCollector`'s `reset()`.

> I'd prefer to avoid forcing the user to work with a file for very simply use cases (which covers a good bunch of what I'll do once this is merged)

Sure, that makes sense. There is nothing to stop you from adding other methods, in addition to those defined by the `TraceCollector`'s API, that implement the semantics you need. As long as the API is fully implemented. 

It might also be worth considering extending the API to handle these sorts of use cases. Though we'd need to make sure it is sufficiently detailed to be implementable by all trace collectors.

valschneider:
> As I've mentioned in an earlier comment, there should be a TraceCollector implementation for this. I don't have a strong opinion right now on whether it is better to turn this into a TraceCollector subclass directly, or whether to create a TraceCollector wrapper for the monitor.

- I created a **trace/logcat.py** file and moved the LogcatMonitor there. I added the LogcatCollector wrapper in there as well. I think it's cleaner but if you disagree I can have it split over **trace/logcat** and **utils/android**
- I also removed the getter in **target** since this is all in **trace/logcat**, but again I could add it back if it makes more sense.

As far as I can tell, the LogcatCollector will be used to simply record what is going on a device, whereas the LogcatMonitor will be used when you're looking for specific things within logcat.

setrofim:
Since the monitor is not not itself a `TraceCollector`, I don't think it should be under `trace/` and I would move it back to `utils/android`. Other than that, looks good.

valschneider:
@setrofim I think this version addresses all of your comments

devlib/trace/logcat.py
```diff
@@ -0,0 +1,58 @@
...
+        self.regexps = regexps
+        self._collecting = False
+        self._prev_log = None
+
+    def reset(self):
+        """
+        Clear Collector data but do not interrupt collection
+        """
+        if not self._monitor:
+            raise RuntimeError('Collector has never been started, nothing to reset')
```
setrofim: It is valid for reset to be called prior to the initial run, so just return here, rather than raise an error.

valschneider:
> It is valid for reset to be called prior to the initial run, so just return here, rather than raise an error.
Done

PR #169, opened 2017-09-13
utils/android: Add grant_app_permissions

valschneider:
This is mostly useful to avoid having to manually click/tap on the permission requests that may pop up when opening apps, which would ruin automation

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -445,6 +445,23 @@ def adb_command(device, command, timeout=None,adb_server=None):
     output, _ = check_output(full_command, timeout, shell=True)
     return output
 
+def grant_app_permissions(target, package):
+    """
+    Grant an app all the permissions it may ask for
+    """
+    dumpsys = target.execute('dumpsys package {}'.format(package))
+
+    permissions = re.search(
+        'requested permissions:\s*(?P<permissions>(android.permission.+\s*)+)', dumpsys
+    )
+    permissions = permissions.group('permissions').replace(" ", "").splitlines()
+
+    for permission in permissions:
+        try:
+            target.execute('pm grant {} {}'.format(package, permission))
+        except TargetError:
+            logger.debug('Cannot grant {}'.format(permission))
+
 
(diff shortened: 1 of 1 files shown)


devlib/utils/android.py
```diff
@@ -445,6 +445,24 @@ def adb_command(device, command, timeout=None,adb_server=None):
...
+
+    permissions = re.search(
+        'requested permissions:\s*(?P<permissions>(android.permission.+\s*)+)', dumpsys
+    )
+    permissions = permissions.group('permissions').replace(" ", "").splitlines()
+
+    for permission in permissions:
+        try:
+            target.execute('pm grant {} {}'.format(package, permission))
+        except:
```
Sticklyman1936: It would be good if this didn't catch everything blindly. What is the error you expect to see here?

valschneider: A quick run with exoplayer raises:

```
TargetError: Got exit code 1
from target command: pm grant com.google.android.exoplayer2.demo android.permission.INTERNET
STDOUT: 
STDERR: Operation not allowed: java.lang.SecurityException: Permission android.permission.INTERNET is not a changeable permission type
```

I think it's safe enough to search for 'java.lang.SecurityException' in stderr and propagate the exception if it's not found.

setrofim: That might break when future version of Android decide to rename the exception or format the error differently. I think catching `TargetError` here should be sufficient, as this wraps a single operation, and programming errors in the underlying implementation will manifest as some other exception type.

Sticklyman1936:
Thanks for fixing. This is now good by me, but I'll let Sergei have the final say. :)

bjackman:
Low-priority though: should probably have some error handling here - ATM if you try this on a package that hasn't been installed I think you'll get a TypeError where `permissions` is None. Not really a problem TBH.

setrofim:
Arguably, this should be checked by the calling code (i.e. do not try to call this with `None`).

setrofim:
Fixed. Thanks for catching this.

PR #170, opened 2017-09-13
FPS and rendering stats DerivedMeasurements

setrofim:
This addes DerivedMeasurements implementations for calculating FPS and related
rendering stats from SurfaceFlinger and GfxInfo frame data collected by
respective instruments.

This also extends the derived measurements API, but adding a method for
processing "raw" output, which is less strict about its inputs, than the
exiting "process", and adds a DerivedMetric type for the results, which, unlike
Measurement, does require defining a Channel (which doesn't really make sense
for a summary static).


devlib/__init__.py
```diff
@@ -20,7 +20,7 @@
 from devlib.instrument.gem5power import Gem5PowerInstrument
```
bjackman: > `drived: renamed derived_measurments --> energy `

s/drived/derived/

devlib/derived/__init__.py
```diff
@@ -12,6 +12,45 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
+from devlib.instrument import MeasurementType, MEASUREMENT_TYPES
+
+
```
bjackman: > The notion of an InstrumentChannel doesn't really make
sense in the context of DerivedMeasurments, which are not directly
measured on the target.

Hmm.. but `DerivedEnergyMeasurements` does derive per-channel metrics, so we technically we lose (although it's actually in the `DerivedMetric`'s `name`) a meaningful mapping between metric and channel there, right? I don't know if that actually matters at all...

devlib/instrument/__init__.py
```diff
@@ -72,9 +72,25 @@ def __repr__(self):
...
+    MeasurementType('count', 'count'),
+    MeasurementType('percent', 'percent'),
+
+    # Time measurement. While there is typically a single "canonical" unit
+    # used for each type of measurmenent, time may be measured to a wide variety
+    # of events occuring at a wide range of scales. Forcing everying into a
+    # single scale will lead to inefficient and awkward to work with result tables.
+    # Coversion functions between the formats are specified, so that downstream
+    # processors that expect all times time be at a particular scale can automatically
+    # covert without being familar with individual instruments.
```
bjackman: I wonder if it would be useful to differentiate between times that are periods ('this frame took 12ms to render') and times that are stamps ('this frame was rendered at t=1005.3s')?

devlib/derived/__init__.py
setrofim: Yes, that's true, but in the general case, there won't necessarily be a `DerivedMetric` --> `InstrumentChannel` mapping, as a metric may be derived from multiple channels, or from auxiliary output. 

I have considered adding a `channels` attribute that would default to an empty list, but could be overwritten to identify the channels that were involved in deriving the metric. However, being unable to think of a good use for it, I decided to err on the side of simplicity.

I'm open to adding that, you think it would be useful.

devlib/__init__.py
setrofim: fixed.

devlib/derived/__init__.py
bjackman: Yeah, I was thinking of an optional `channel` attribute like you suggested (and a list is better), but you're right it's probably unnecessary complexity unless we know a use case. Can always add it later if one arises.

devlib/instrument/__init__.py
setrofim: hm, possibly, however, periods can also be large ("this gem5 simulation took several days to run"), so it would not be enough to simply have "time" measurement in "seconds" and "period" measurement in "microseconds". You'd arguably still need the three levels here.

It may be useful to add an additional "timestamp" measurement type, but again, I'm hesitant to do so without something that actually requires it. In order for that to make sense if there is something that cares about absolute vs relative time difference, but not about the other details of the metric (otherwise just use metric name).

bjackman: OK, that makes sense too.

devlib/derived/fps.py
```diff
@@ -0,0 +1,205 @@
...
+            frame_count += 1
+
+            if start_vsync is None:
+                start_vsync = frame_data.Vsync_time_us
+            end_vsync = frame_data.Vsync_time_us
+
+            frame_time = frame_data.FrameCompleted_time_us - frame_data.IntendedVsync_time_us
+            pff = 1e9 / frame_time
+            if pff > self.drop_threshold:
+                per_frame_fps.append(pff)
```
bjackman: Should be `per_frame_fps.append([pff])` so that writer.writerows doesn't error with 'sequence expected'

devlib/derived/fps.py
```diff
@@ -0,0 +1,205 @@
...
+        if self.filename:
+            return os.path.join(outdir, self.filename)
+
+        frames_basename = os.path.basename(frames_file)
+        rest, ext = os.path.splitext(frames_basename)
+        csv_basename = rest + self.suffix + ext
+        return os.path.join(outdier, csv_basename)
+
+
+class DerivedGfxInfoStats(DerivedFpsStats):
```
bjackman: This could probably do with a bit of explanation and a link to https://developer.android.com/training/testing/performance.html

devlib/derived/fps.py
```diff
@@ -0,0 +1,205 @@
...
+        for frame_data in measurements_csv.iter_values():
+            frame_count += 1
+
+            if start_vsync is None:
+                start_vsync = frame_data.Vsync_time_us
+            end_vsync = frame_data.Vsync_time_us
+
+            frame_time = frame_data.FrameCompleted_time_us - frame_data.IntendedVsync_time_us
+            pff = 1e9 / frame_time
+            if pff > self.drop_threshold:
```
bjackman: I'm not totally convinced about this logic - FrameCompleted - IntendedVsync is supposed to be a real measure of the time spent rendering, right? So if it goes goes over a certain time, that sounds like a real indicator of slow rendering (i.e. not just a static frame)?

bjackman: Oh also https://developer.android.com/training/testing/performance.html says you can only use FrameCompleted - IntendedVsync when the flags column is 0

setrofim: Ah., I missed that. Thanks for catching! I've made the fix.

devlib/derived/fps.py
setrofim: Fixed.

devlib/derived/fps.py
setrofim: See documentation: https://github.com/ARM-software/devlib/pull/170/files#diff-3cab921f2a5cb7f3f292f5600a3e49c3R179

bjackman: Sorry, I always forget about the .rst files!

devlib/derived/fps.py
setrofim: > I'm not totally convinced about this logic - FrameCompleted - IntendedVsync is supposed to be a real measure of the time spent rendering, right? So if it goes goes over a certain time, that sounds like a real indicator of slow rendering (i.e. not just a static frame)?

In some cases, we see bogus frames where the timestamps are obviously wrong. This is to filter those out.

devlib/utils/rendering.py
```diff
@@ -85,8 +85,12 @@ def write_frames(self, outfile, columns=None):
...
             frames = self.frames
         else:
-            header = [c for c in self.header if c in columns]
-            indexes = [self.header.index(c) for c in header]
+            indexes = []
+            for c in columns:
+                if c not in self.header:
+                    msg = 'In  valid column "{}"; must be in {}'
+                    raise ValueError(msg.format(c, self.header))
+                indexes.append(self.haader.index(c))
```
bjackman: s/haader/header/

devlib/utils/rendering.py
```diff
@@ -85,8 +85,12 @@ def write_frames(self, outfile, columns=None):
             header = self.header
             frames = self.frames
         else:
-            header = [c for c in self.header if c in columns]
-            indexes = [self.header.index(c) for c in header]
+            indexes = []
+            for c in columns:
+                if c not in self.header:
+                    msg = 'In  valid column "{}"; must be in {}'
```
bjackman: Should be 'invalid'

devlib/derived/fps.py
```diff
@@ -0,0 +1,208 @@
...
+            writer.writerow(['fps'])
+            writer.writerows(per_frame_fps)
+
+        return [DerivedMetric('fps', fps, 'fps'),
+                DerivedMetric('total_frames', frame_count, 'frames'),
+                MeasurementsCsv(csv_file)]
+
+    def _process_with_pandas(self, measurements_csv):
+        data = pd.read_csv(measurements_csv.path)
+        data = data[data.Flags_flag == 0]
```
bjackman: Should be `data.Flags_flags` (missing s)

setrofim: All recent comments should be fixed.

devlib/utils/rendering.py
```diff
@@ -83,8 +85,12 @@ def write_frames(self, outfile, columns=None):
...
             frames = self.frames
         else:
-            header = [c for c in self.header if c in columns]
-            indexes = [self.header.index(c) for c in header]
+            indexes = []
+            for c in columns:
+                if c not in self.header:
+                    msg = 'Invalid column "{}"; must be in {}'
+                    raise ValueError(msg.format(c, self.header))
+                indexes.append(self.header.index(c))
```
bjackman: I think 
```
header = columns
```
 is missing

setrofim: Ah yes. Thanks for catching that! Fixed.

bjackman:
Couple of things I've noticed while fiddling with this:

- This reports stuff in us but https://developer.android.com/training/testing/performance.html says all timestamps are in nanoseconds

- AFAICT there's not code to detect duplicate frames in the 'framestats' info. So if the app is running consistently at 30fps then this will collect (up to) 120 frames' stats every ~2 seconds: the first time that's the first ~60 frames, the second time it will be the first ~120 frames, the 3rd time it will be frames ~60-~180.

setrofim:
Yup, I've been looking at the same thing, and came to the same conclusion. I have just pushed an additional commit that should fix those issues. I'm not sure it would sort the mis-calculated FPS though.

Unfortunately, while attempting to test this, I've run into some revent-related issues in WA, which I am now trying to resolve. If you've got the time, please could you test out the latest  patch with your workload and send me the results?

bjackman:
Sure, I'll take a look now.

devlib/utils/rendering.py
```diff
@@ -205,7 +206,11 @@ def _process_raw_file(self, fh):
                 for line in fh:
                     if line.startswith('---PROFILEDATA---'):
                         break
-                    self.frames.append(map(int, line.strip().split(',')[:-1]))  # has a trailing ','
+                    entries = map(int, line.strip().split(',')[:-1])  # has a trailing ','
+                    if entries[-1] <= last_vsync:
```
bjackman: Ah, the reason I only saw 7 frames is because this assumes the vsync timestamp is the last column; it's actually `entries[1]`. Fixing that gets us all the entries from `frames.csv.raw` in `frames.csv`, but there are still not as many as expected. Will keep investigating.

bjackman:
Hmm after fixing the `entry / 1000` thing and the index of the intended vsync column in the 'last_vsync' overlap detection I still found that I was only seeing frame data for a minimal number of frames.

I dumped all the unique framestats entries into a csv file and took a look at that, and it looks like I am getting only sporadic frame data (I have frame stats spread roughly evenly over 40 seconds but only for 56 frames):

https://gist.github.com/bjackman/3bec7647fc8ed549412f8bcefbb0ff5f

This makes me think that it's not the devlib instrumentation logic but the way gfxinfo is hooked up - I'm using exoplayer so maybe the frames being rendered are not being attributed to the right package

setrofim:
Will re-submit FPS derived stats separately from the other comits, as the gfxinfo issue is preventing orthogonal fixes from being merged.

PR #171, opened 2017-09-13
instrument: Clear up Instrument.reset semantics

bjackman:
- Fix missing parameter in the documentation

- Clarify meaning of `sites` and `kinds` in the documentation.

- With the current implementation the `channels` argument is
  useless: if `sites` and `kinds` are not also specified then all
  channels are enabled anyway. Fix that by making those parameters
  ignored when `channels` is provided.

### DIFF
--- a/devlib/instrument/__init__.py
+++ b/devlib/instrument/__init__.py
@@ -258,24 +258,29 @@ class Instrument(object):
         pass
 
     def reset(self, sites=None, kinds=None, channels=None):
+        self.active_channels = []
         if kinds is None and sites is None and channels is None:
             self.active_channels = sorted(self.channels.values(), key=lambda x: x.label)
-        else:
-            if isinstance(sites, basestring):
-                sites = [sites]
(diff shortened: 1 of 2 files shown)


devlib/instrument/__init__.py
```diff
@@ -258,24 +258,29 @@ def teardown(self):
...
-            if isinstance(sites, basestring):
-                sites = [sites]
-            if isinstance(kinds, basestring):
-                kinds = [kinds]
-            self.active_channels = []
-            for chan_name in (channels or []):
+        elif channels is not None:
+            if sites is not None or kinds is not None:
+                raise ValueError(
+                    'sites and kinds should not be set if channels is set')
```
valschneider: nitpick: I'd put the closing parenthesis on a new line.

bjackman: Just did `git --no-pager grep -n -e \(\$ -- ` to find other instances, looks like the rest of the code closes it on the same line

PR #173, opened 2017-09-19
instrument/frames: Fix parsing MeasurementsCsv

bjackman:
The collector's write_frames method uses a fixed column order, only using the
'columns' param as a filter i.e. if its columns are ['a', 'b', 'c'] and you pass
columns=['c', 'a'], the output columns will be ['a', 'c'].

Therefore passing self.active_channels breaks MeasurementsCsv parsing when its
order does not match the collector's column order.

Since the collector writes a header in the CSV file, just pass columns=None to
MeasurementsCsv so it will parse that header to detect the column order.

### DIFF
--- a/devlib/instrument/frames.py
+++ b/devlib/instrument/frames.py
@@ -44,7 +44,7 @@ class FramesInstrument(Instrument):
         self.collector.process_frames(raw_outfile)
         active_sites = [chan.label for chan in self.active_channels]
         self.collector.write_frames(outfile, columns=active_sites)
-        return MeasurementsCsv(outfile, self.active_channels, self.sample_rate_hz)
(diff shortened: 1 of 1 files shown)


bjackman:
This was tested using https://github.com/ARM-software/devlib/pull/170, not sure if there are compat issues with other bits?

setrofim:
I'ved updated that pull request with

https://github.com/ARM-software/devlib/pull/170/commits/41bb452084040339641ea510b3ee22345602ef0a

Which should ensure that the order of columns written by the collector matches active_channels.

bjackman:
OK that works too. On the other hand it was quite nice that the CSV file had the same order as the gfxinfo output you see in Android...

setrofim:
It should still have that by default; active_channels should default to that order...

bjackman:
(y)

PR #175, opened 2017-09-20
Fix race in LogcatMonitor

bjackman:


bjackman:
CC @valschneider

bjackman:
nope! still a race, updating shortly

bjackman:
Fixed. I hope the commit message makes sense..

bjackman:
Fixed. I hope the commit message makes sense..

setrofim:
What happens if `stop()` is called before `start() `? It *is* illegal, but we want a better failure mode than "wait forever". 

Proposed solution: Make sure `_stopped` is set inside `__init__`, unset it inside `start`,  immediately return from `stop` if `_stopped` is set.

valschneider:
@setrofim Couldn't we just use this:

```
if not self.is_active()
    return
```

The way I see it, `self.is_active()` answers the question "have I called start?"

bjackman:
Updated with the check for `is_alive`

PR #180, opened 2017-09-28
android: Add AndroidScreen class with orientation methods

bjackman:
This can be used to save, set and restore the screen
orientation.

This can be useful for ensuring the reproducibility of results that
depend on the actual content that was rendered. A simple example of
this is that a device with an OLED screen uses measurably more energy
playing a plain-white Youtube video if it is in landscape orentation,
due to having more white pixels lit up.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -14,7 +14,8 @@ from devlib.module import get_module
 from devlib.platform import Platform
 from devlib.exception import TargetError, TargetNotRespondingError, TimeoutError
 from devlib.utils.ssh import SshConnection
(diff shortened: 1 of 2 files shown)


bjackman:
CC @derkling  @valschneider @mdigiorgio 

I changed the API and implementation from the [LISA class this was based on](https://github.com/ARM-software/lisa/blob/master/libs/utils/android/screen.py#L27), I _think_ that this is clearer but please let me know if you disagree

bjackman:
Huh.. looks like devlib already has this. Off I go then :grin:

PR #189, opened 2017-10-11
AndroidTarget: Add boolean 'state' to airplane mode intent broadcast

bjackman:
Recently I've been seeing some errors with enabling airplane mode on
hikey960 - firstly I get a 'broken pipe' error followed by a 'cmd:
Can't find service: settings'. This missing boolean was the first
thing I found that had changed wrt. the code I used to use, and
adding it back appears to fix the issue.


derkling:
LGTM

PR #190, opened 2017-10-11
android: Use pexpect for LogcatMonitor

bjackman:
I wrote this for another reason which turned out not to be necessary. However it turns out that the current LogcatMonitor can't be killed with Ctrl+C while it's in the `threading.Event.wait` in `wait_for`, which is irritating. So I'm submitting this now. It also reduces lines of code which I guess is Good.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -20,6 +20,7 @@ Utility functions for working with Android devices through adb.
 """
 # pylint: disable=E1103
 import os
+import pexpect
 import time
 import subprocess
 import logging
@@ -543,19 +544,18 @@ def _check_env():
         adb = _env.adb
         aapt = _env.aapt
 
-class LogcatMonitor(threading.Thread):
+class LogcatMonitor(object):
     """
     Helper class for monitoring Anroid's logcat
 
     :param target: Android target to monitor
     :type target: :class:`AndroidTarget`
 
+    :param regexps: List of uncompiled regular expressions to filter on the
                     device. Logcat entries that don't match any will not be
                     seen. If omitted, all entries will be sent to host.
     :type regexps: list(str)
     """
-
-    FLUSH_SIZE = 1000
 
     @property
     def logfile(self):
@@ -565,16 +565,6 @@ class LogcatMonitor(threading.Thread):
         super(LogcatMonitor, self).__init__()
 
         self.target = target
-
-        self._started = threading.Event()
-        self._stopped = threading.Event()
-        self._match_found = threading.Event()
-
-        self._sought = None
-        self._found = None
-
-        self._lines = Queue.Queue()
-        self._datalock = threading.Lock()
         self._regexps = regexps
 
     def start(self, outfile=None):
@@ -585,15 +575,10 @@ class LogcatMonitor(threading.Thread):
         :type outfile: str
         """
         if outfile:
-            self._logfile = outfile
+            self._logfile = open(outfile)
         else:
-            fd, self._logfile = tempfile.mkstemp()
-            os.close(fd)
-        logger.debug('Logging to {}'.format(self._logfile))
+            self._logfile = tempfile.NamedTemporaryFile()
 
-        super(LogcatMonitor, self).start()
-
-    def run(self):
         self.target.clear_logcat()
 
         logcat_cmd = 'logcat'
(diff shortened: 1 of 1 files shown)


bjackman:
CC @valschneider

devlib/utils/android.py
```diff
@@ -605,86 +590,32 @@ def run(self):
...
+        logcat_cmd = get_adb_command(self.target.conn.device, logcat_cmd)
 
-        while not self._stopped.is_set():
-            line = self._logcat.stdout.readline(1024)
-            if line:
-                self._add_line(line)
+        logger.debug('logcat command ="{}"'.format(logcat_cmd))
+        self._logcat = pexpect.spawn(logcat_cmd, logfile=self._logfile)
 
     def stop(self):
```
derkling: Should not this be called also from a dtor, i.e. the `__del__` method?

devlib/utils/android.py
```diff
@@ -20,6 +20,7 @@
 """
```
valschneider: From the commit msg: "[...] a separate thread along with the synchronization that brings."  That should be "the synchronization that **it** brings", right ?

devlib/utils/android.py
```diff
@@ -20,6 +20,7 @@
...
 # pylint: disable=E1103
 import os
+import pexpect
 import time
 import subprocess
```
valschneider: Do we still need that import ?

valschneider: Wait nevermind that, it's not a file with just the monitor in it.

devlib/utils/android.py
```diff
@@ -704,16 +635,11 @@ def wait_for(self, regexp, timeout=30):
...
-        found = self._match_found.wait(timeout)
-        self._match_found.clear()
-        self._sought = None
-
-        if found:
-            return [self._found]
-        else:
+        try:
+            self._logcat.expect(regexp, timeout=timeout)
+            return [self._logcat.match.groups(0)]
```
valschneider: Hm that's a change of behaviour with the current code, but I guess it goes with how pexpect works.
By returning `match.groups(0)`, you can't pass regexps with named sections or several matching groups.

Unless I'm wrong on how pexpect returns regexp matches, it means you have different return behaviours depending if you immediately found a match (return `self.search(regexp), which is a list of lines)` or if you had to wait to find a match (return `[self._logcat.match.groups(0)]`) which is a list of matches.

bjackman: Hmm, not certain if I understand you - are you saying that this should instead return `[self._logcat.match.group(0)]` (`group` not `groups`, so it returns the full match)? If so then yes I think you are right - I probably just typo'd' this.

valschneider: What I meant is that we shouldn't be returning the match in this case, but the whole line that caused the match to happen. That's what `search()` does:
```
return [line for line in self.get_log() if re.match(regexp, line)]
```

devlib/utils/android.py
bjackman: I've never used that method before... but it could be useful since it means a broken user that forgets to call `stop` won't leave a `pexect` lying around. @setrofim what do you think? If we go for it though, I'll do it in a separate PR.

setrofim: Adding a call to `stop` inside `pexpect` is a good idea to make sure `logfile` is always closed properly, even if the client code neglects to explicitly close the monitor.

However, that would mean that `stop` could be called multiple times, so you should make sure that it would be safe to do so (that the second call does not raise errors).

And yeah, probably best to do this in a separate PR.

devlib/utils/android.py
bjackman: OK, after a bit of fiddling I discovered that [pexpect's regex matching is a little unusual](https://pexpect.readthedocs.io/en/stable/overview.html#find-the-end-of-line-cr-lf-conventions) so I've just set this to return the result of `search`. A little inefficient but it's on the host so I don't care, and it's an easy way to get consistency.

devlib/utils/android.py
bjackman: Ummmmmmmmmmmmmm I think [it's 'along with the synchronisation [that] that brings' - the first 'that' introduces a relative clause but is sort of optional, the second 'that' is a relative pronoun referring to the separate thread.](https://s-media-cache-ak0.pinimg.com/originals/78/d7/fc/78d7fc8f03a3e1a920e7b195ddce2644.jpg)

devlib/utils/android.py
valschneider: Fair enough. Since `search` is just a one-liner, maybe we could just save the index of the last line in the log before we wait for a match, and then do a search from this index forward ?

devlib/utils/android.py
bjackman: Clearer example would be "the queen is ashamed to admit _that_ the language _she_ represents is competely absurd" (where the "that" that introduces a relative clause could just be dropped and the "she" is a relative pronoun)

(#nerdsniped)

bjackman: (OK back to work)

valschneider: Fair enough, so much for me trying to be nitpicky on the Queen's language :D

devlib/utils/android.py
bjackman: OK, I'll do that

bjackman: OK that's done

devlib/utils/android.py
```diff
@@ -701,19 +632,18 @@ def wait_for(self, regexp, timeout=30):
...
             return res
 
-        # Did not find any match, wait for one to pop up
-        self._sought = regexp
-        found = self._match_found.wait(timeout)
-        self._match_found.clear()
-        self._sought = None
+        # Store the number of lines we've searched already, so we don't have to
+        # re-grep them after 'expect' returns
+        next_line_num = len(self.get_log())
```
valschneider: Mmm unlikely but not impossible: between the `if len(res) > 0` and this index store, we could've received a matching line. That's something the `_datalock` was taking care of in the previous implementation. Could it be possible to freeze the `pexpect` process before doing the `search`, unfreeze it if we found lines or after we've saved the current index ?

bjackman: That isn't a problem - `expect` still checks lines that happened before it got called.

valschneider: Oh, right, didn't read that properly. Looks good to me !

bjackman: Oh yeah damn you're right. I'll fix it by gluing together the `self.search()` and `self.get_log()`

bjackman: OK, should be fixed now. Thanks a lot for the careful review :)

valschneider: I think you could just store `next_line_num` before doing a call to `search` - which I believe is equivalent to what you're doing now (?)

valschneider: Mm actually it might just be better to keep what you have right now but do
```
res = [line for line in log if re.match(regexp, line)]
```
instead of
```
res = [line for line in self.get_log() if re.match(regexp, line)]
```

bjackman: OH damn sorry that's exactly what I meant to do, just rushed it. Updating once again.

valschneider:
That LGTM now :)

setrofim:
Is this ready for merge now? Most importantly, is the Queen happy with the English in the commit message? ;)

bjackman:
Yup, ready for merge. Spoke to her royal majesty, she says LGTM and is relieved to have the Ctrl+C thing fixed.

PR #193, opened 2017-10-18
target: Add is_network_connected method

bjackman:
This is for implementing the `requires_network` parameter in WA3..

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -31,6 +31,8 @@ KVERSION_REGEX =re.compile(
     r'(?P<version>\d+)(\.(?P<major>\d+)(\.(?P<minor>\d+)(-rc(?P<rc>\d+))?)?)?(.*-g(?P<sha1>[0-9a-fA-F]{7,}))?'
 )
 
+GOOGLE_DNS_SERVER_ADDRESS = '8.8.8.8'
+
 
 class Target(object):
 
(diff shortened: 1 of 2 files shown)


devlib/target.py
```diff
@@ -701,6 +703,33 @@ def _install_module(self, mod, **params):
     def _resolve_paths(self):
         raise NotImplementedError()
 
+    def is_network_connected(self):
+        self.logger.debug('Checking for internet connectivity...')
+        return self._ping_server(GOOGLE_DNS_SERVER_ADDRESS)
+
+    def _ping_server(self, ip_address, timeout=1, packet_count=1):
```
derkling: why not exposing this also as a public API?

devlib/target.py
```diff
@@ -701,6 +703,33 @@ def _install_module(self, mod, **params):
...
+        command = 'ping -q -c {} -w {} {}'.format(packet_count, timeout, ip_address)
+
+        # Ideally we'd like to just run with check_exit_code=False, then check
```
derkling: Not entirely sure about how it plays with self.execute, but... what about just using something like:

```python
command = 'ping -q -c {} -w {} {}; echo $?'.format(packet_count, timeout, ip_address)
output = self.execute(command)
if output == '0':
    # Success
else
   # Failure
```

devlib/target.py
derkling: Should not be safer to usage a slightly longer timeout... maybe 3s?

devlib/target.py
setrofim: `execute` takes care of that. All you need to do is

```
try:
     self.execute(command)
     return True
except TargetError:
     return False
```

devlib/target.py
setrofim: > why not exposing this also as a public API?

I think keeping this internal for now is the right move. We can always open it up later, when there is a clear need.

bjackman:
Don't think I was thinking clearly when I originally wrote this patch! 

New version removes the pointless intermediate method, and just uses `2>&1` to get stderr. That means we can ignore the exit code - we don't care whether the ping actually succeeded (if we did, we would have to drastically increase the packet count and timeout to get a reliable test), just about whether the system thinks it's theoretically possible to get packets to `8.8.8.8`. 

In the end the use case for this is basically for failing early when we definitely aren't connected to the internet, not checking that we definitely are (which is a slightly nebulous thing anyway).

devlib/target.py
```diff
@@ -701,6 +703,20 @@ def _install_module(self, mod, **params):
...
+
+        ip_address = GOOGLE_DNS_SERVER_ADDRESS
+
+        command = 'ping -q -c 1 -w 1 {} 2>&1'.format(ip_address)
+
+        output = self.execute(command, check_exit_code=False)
+        if 'network is unreachable' in output.lower():
+            self.logger.debug('Cannot find IP address {}'.format(ip_address))
+            return False
+        else:
```
setrofim: What if the command fails for some other reason (e.g. ping is not on target)? Should probably check that the command succeed, as you're disabling `check_exit_code`. Also, would probably be a good idea to invoke ping via busybox...

bjackman: >  Should probably check that the command succeed

Hmm, yeah we should probably check we were actually able to execute `ping`. I just don't want to introduce a failure if the ICMP packet happens to get lost or delayed, and I don't want to introduce a delay by sending several. I'll see what I can think of.

>  Also, would probably be a good idea to invoke ping via busybox...

Yeah.. the problem is that we need root to ping - the `ping` command typically has `setuid` set.

setrofim: > Yeah.. the problem is that we need root to ping - the ping command typically has setuid set.

Sorry, I'm not sure I understand this. Could you elaborate? Why do you need root?

bjackman: [You need a certain capability to open the type of socket that you need to send ICMP packets](https://superuser.com/a/1035983/446515).

setrofim: Ah, ok, I see. That makes sense. Thanks.

bjackman:
OK, third attempt.. this time we don't ignore the exit code, but we still check the output for certain strings. I'm fairly optimistic that those strings are pretty universal.

PR #200, opened 2017-11-08
Gem5 fixes

Sticklyman1936:
Minor gem5 fixes to address various issues when interacting.

### DIFF
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -757,9 +757,9 @@ class Gem5Connection(TelnetConnection):
         """
         gem5_logger.info("Mounting VirtIO device in simulated system")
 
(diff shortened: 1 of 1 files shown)


devlib/utils/ssh.py
```diff
@@ -695,6 +695,9 @@ def _gem5_shell(self, command, as_root=False, timeout=None, check_exit_code=True
 
         gem5_logger.debug("gem5_shell command: {}".format(command))
 
+        if as_root:
+            command = 'echo "{}" | su'.format(command)
```
setrofim: Should be using `escape_double_quotes()` on the `command` to make sure the echo quote isn't closed early if there happen to be a double quote character in the command itself.

PR #206, opened 2017-11-20
shutils: consider different hierarchies when running command into cgr…

mdigiorgio:
…oups

The target platform can have different hierarchies with different cgroup
controllers mounted. If we have a cgroup that uses different controllers
belonging to different hierarchies, that cgroup will be present under the
different hierarchies. Therefore, in such cases we need to take into all those
paths having that cgroup and move the task into all of them.


derkling:
LGTM

PR #210, opened 2017-11-22
Gem5

marcbonnici:


devlib/utils/ssh.py
```diff
@@ -727,11 +727,11 @@ def _gem5_shell(self, command, as_root=False, timeout=None, check_exit_code=True
 
         output = output[command_index + len(command):].strip()
 
-        # It is possible that gem5 will echo the command. Therefore, we need to
-        # remove that too!
-        command_index = output.find(command)
-        if command_index != -1:
-            output = output[command_index + len(command):].strip()
+        # If gem5 echoes the command we need to remove that too!
```
Sticklyman1936: `# If the gem5 system echoes the executed command, we need to remote that too!`

devlib/utils/ssh.py
```diff
@@ -727,11 +727,11 @@ def _gem5_shell(self, command, as_root=False, timeout=None, check_exit_code=True
...
+        if self.shell_echo:
+            command_index = output.find(command)
+            if command_index != -1:
+                output = output[command_index + len(command):].strip()
```
Sticklyman1936: Do we need to do something more here to make sure that this works as intended? We might want to limit for far it is allowed to look into the output, to try and avoid it matching things in the middle of the output. Just a thought - maybe I am thinking too much into this.

devlib/utils/ssh.py
```diff
@@ -349,6 +350,8 @@ def __init__(self,
         self.is_rooted = True
         self.password = None
         self.port = None
+        # Flag to indicate whether commands are echoed
```
Sticklyman1936: +`by the simulated system`?

setrofim:
I think `shell_echo` is a bit too generic a name for this.  It may be misinterpreted as controlling the echoing rather than stripping the output. Maybe something like `strip_echoed_commands` instead...

devlib/utils/ssh.py
setrofim: I don't think it's worth changing anything here, since we have no means of testing the change would actually help. I think we need to wait until this becomes an issue.

PR #212, opened 2017-12-01
 Target: Implement as_root parameter for push/pull in base class 

bjackman:
The AndroidTarget implementation is actually generic, so just move it
into the parent class so that LinuxTarget gets it too.


bjackman:
Oops didn't mean to submit this yet, will reopen once tested!

PR #228, opened 2018-01-15
[RFC] Add Initial Support for ChromeOs

marcbonnici:


devlib/target.py
```diff
@@ -1688,3 +1688,90 @@ def process_node(node, path, value):
...
+        self.android_connection_settings = {}
+        for setting in adb_conn_params:
+            if connection_settings.get(setting, None):
+                self.android_connection_settings[setting] = connection_settings[setting]
+
+        # If adb device is not explicitly specified use same as ssh host
+        if not connection_settings.get('device', None):
+            self.android_connection_settings['device'] = connection_settings.get('host', None)
+
+        self.android_target = AndroidTarget(connection_settings=self.android_connection_settings,
```
setrofim: I would rename this to just `android` (or maybe `android_container` though that might be a bit long), as `target.android_target` would look weird in the calling code.

doc/target.rst
```diff
@@ -595,3 +595,37 @@ Android Target
...
    support the majority of newer devices.
+
+
+ChromeOs Target
+---------------
+
+.. class:: ChromeOsTarget(connection_settings=None, platform=None, working_directory=None, executables_directory=None, android_working_directory=None, android_executables_directory=None, connect=True, modules=None, load_default_modules=True, shell_prompt=DEFAULT_SHELL_PROMPT, package_data_directory="/data/data")
+
+    :class:`ChromeOsTarget` is a subclass of :class:`LinuxTarget` with
+    additional features specific to a device running ChromeOs including its
```
setrofim: ChoromeOs --> ChromeOS

doc/target.rst
```diff
@@ -595,3 +595,37 @@ Android Target
...
+    own android container which can be accessed via the ``android_target``
+    attribute. When making calls to or accessing properties and attributes
+    of the ChromeOs target, by default they will be applied to the linux
```
setrofim: ChoromeOs --> ChromeOS, linux --> Linux

setrofim:
Also, this is more of a question then a comment, but is the Android container an essential part of android device? At the moment the Android connection is always instantiated unconditionally, which is fine if we can rely on Android always being there. but just wanted to check.

devlib/target.py
derkling: `android_container` looks better to me... just to be more clear that's an instance of Android running within an host system.

derkling:
AFAIN not all Chromebooks will support Android apps... dunno if this can impact the code above...

marcbonnici:
Updated to try and detect whether the device supports android apps by detecting the presence of `/opt/google/containers/android/` upon connection to the device before conditionally trying to connect to the android side.

PR #229, opened 2018-01-16
LocalLinuxTarget: fix executables folder location

derkling:
Since:
   cafc0a4 target: do not create shutil in package directory
we generate 'shutils' in /tmp, which is also the executables_path
used used by deafult by a LocalLinuxTarget.

This ultimately results into a:

   self.install(shutils_ofile)
      ...
         shutil.copy(source, dest)

raising an exception since source == dest.

Let's fix this by setting /tmp/devlib-target as a default base path for
all devlib deployed stuff into a localhost target.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1644,10 +1644,10 @@ class LocalLinuxTarget(LinuxTarget):
 
     def _resolve_paths(self):
         if self.working_directory is None:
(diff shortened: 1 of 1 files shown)


derkling:
Update to address comments on commit message's typos ;-)

PR #233, opened 2018-01-18
gem5Connection: Catch EOF errors & make more user friendly

Sticklyman1936:
Previously, when the gem5 simulation crashed, one would get errors
relating to pexpect reaching EOF, rather than an informative message
stating the gem5 itself had crashed. With this change, we catch some
of the common cases where this can happen, and inform the user if gem5
itself has crashed. In the event that the gem5 simulation itself has
not reported an error, we instead re-throw the original pexpect error.

### DIFF
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -554,6 +554,19 @@ class Gem5Connection(TelnetConnection):
 
         self.connect_gem5(port, gem5_simulation, gem5_interact_dir, gem5_out_dir)
 
+    # Handle the EOF exception raised by pexpect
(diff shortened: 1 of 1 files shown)


devlib/utils/ssh.py
```diff
@@ -554,6 +554,19 @@ def _connect_gem5_platform(self, platform):
 
         self.connect_gem5(port, gem5_simulation, gem5_interact_dir, gem5_out_dir)
 
+    # Handle the EOF exceptino raised by pexpect
```
marcbonnici: /s "exceptino" -> "exception"

PR #237, opened 2018-02-08
arm-probe instrument

vingu-linaro:
Add new energy instrument that is based on arm-probe tool to manage AEP
This tools uses a config file and manages power topology description when computing power figures
It can also aggregate channels

### DIFF
--- a/devlib/instrument/arm_energy_probe.py
+++ b/devlib/instrument/arm_energy_probe.py
@@ -1,0 +1,130 @@
+#    Copyright 2018 Linaro Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+# pylint: disable=W0613,E1101,access-member-before-definition,attribute-defined-outside-init
+from __future__ import division
+import os
+import csv
+import subprocess
+import signal
+import struct
+import sys
+
+import tempfile
+import shutil
+
+from devlib.instrument import Instrument, CONTINUOUS, MeasurementsCsv
+from devlib.exception import HostError
+from devlib.utils.misc import which
+
+from devlib.utils.parse_aep import AEP_parser
+
+class ArmEnergyProbeInstrument(Instrument):
+    """Collects power traces using the ARM Energy Probe.
+            This instrument requires ``arm-probe`` utility to be installed on the host and be in the PATH.
+            arm-probe is available here:
+                ``https://git.linaro.org/tools/arm-probe.git``.
+
+            Details about how to build and use it is available here:
+                ``https://git.linaro.org/tools/arm-probe.git/tree/README``
+
+            ARM energy probe (AEP) device can simultaneously collect power from up to 3 power rails and
+            arm-probe utility can record data from several AEP devices simultaneously.
+
+            To connect the energy probe on a rail, connect the white wire to the pin that is closer to the
+            Voltage source and the black wire to the pin that is closer to the load (the SoC or the device
+            you are probing). Between the pins there should be a shunt resistor of known resistance in the
+            range of 5 to 500 mOhm but the voltage on the shunt resistor must stay smaller than 165mV.
+            The resistance of the shunt resistors is a mandatory parameter to be set in the ``config`` file.
+    """
+
+    mode = CONTINUOUS
+
+    MAX_CHANNELS = 12 # 4 Arm Energy Probes
+
+    def __init__(self, target, config_file='./config-aep', ):
+        super(ArmEnergyProbeInstrument, self).__init__(target)
+        self.arm_probe = which('arm-probe')
+        if self.arm_probe is None:
+            raise HostError('arm-probe must be installed on the host')
+        #todo detect is config file exist
+        self.attributes = ['power', 'voltage', 'current']
+        self.sample_rate_hz = 10000
+        self.config_file = config_file
+
+        self.parser = AEP_parser()
+        #TODO make it generic
+        topo = self.parser.topology_from_config(self.config_file)
+        for item in topo:
+            if item == 'time':
+                self.add_channel('timestamp', 'time')
+            else:
+                self.add_channel(item, 'power')
+
+    def reset(self, sites=None, kinds=None, channels=None):
+        super(ArmEnergyProbeInstrument, self).reset(sites, kinds, channels)
+        self.output_directory = tempfile.mkdtemp(prefix='energy_probe')
+        self.output_file_raw = os.path.join(self.output_directory, 'data_raw')
+        self.output_file = os.path.join(self.output_directory, 'data')
+        self.output_file_figure = os.path.join(self.output_directory, 'summary.txt')
+        self.output_file_error = os.path.join(self.output_directory, 'error.log')
+        self.output_fd_error = open(self.output_file_error, 'w')
+        self.command = 'arm-probe --config {} > {}'.format(self.config_file, self.output_file_raw)
+
+    def start(self):
+        self.logger.debug(self.command)
+        self.armprobe = subprocess.Popen(self.command,
+                                       stderr=self.output_fd_error,
+                                       preexec_fn=os.setpgrp,
+                                       shell=True)
+
+    def stop(self):
+       self.logger.debug("kill running arm-probe")
+       os.killpg(self.armprobe.pid, signal.SIGTERM)
+
+    def get_data(self, outfile):  # pylint: disable=R0914
+        self.logger.debug("Parse data and compute consumed energy")
+        self.parser.prepare(self.output_file_raw, self.output_file, self.output_file_figure)
+        self.parser.parse_AEP()
+        self.parser.unprepare()
+        skip_header = 1
+
+        all_channels = [c.label for c in self.list_channels()]
+        active_channels = [c.label for c in self.active_channels]
+        active_indexes = [all_channels.index(ac) for ac in active_channels]
+
+        with open(self.output_file, 'rb') as ifile:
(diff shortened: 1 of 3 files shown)


devlib/instrument/parse_aep.py
```diff
@@ -0,0 +1,509 @@
...
+                try:
+                    power += data[child]
+                except IndexError:
+                    continue
+            data.append(power)
+
+        return data
+
+    @staticmethod
+    def delta_nrj(array, delta, min=[], max=[], hide=[], virtual=[]):
```
setrofim: the defaults here should be `None` and then updated to empty lists in the function body. See:

http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

devlib/instrument/parse_aep.py
```diff
@@ -0,0 +1,509 @@
...
+import os
+import sys
+import getopt
+import subprocess
+import signal
+import serial
+import time
+import math
+
+class AEP_parser:
```
setrofim: should be `class AEP_parser(object):`.

devlib/instrument/parse_aep.py
```diff
@@ -0,0 +1,509 @@
+#!/usr/bin/env python
```
setrofim: As mentioned above, this probably belongs in  the same file the instrument. But even if this does end up in a separate module, `devlib/instrument/` should *only* contain instruments.  This would belong under `devlib/utils/`

devlib/instrument/arm_energy_probe.py
```diff
@@ -0,0 +1,130 @@
...
+
+import tempfile
+import shutil
+
+from devlib.instrument import Instrument, CONTINUOUS, MeasurementsCsv
+from devlib.exception import HostError
+from devlib.utils.misc import which
+
+sys.path.append(os.path.dirname(sys.modules[__name__].__file__))
+from parse_aep import AEP_parser
```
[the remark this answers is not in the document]
vingu-linaro: Yes this python class is also used in other tools to parse results from arm-probe

devlib/instrument/arm_energy_probe.py
```diff
@@ -0,0 +1,130 @@
...
+
+class ArmEnergyProbeInstrument(Instrument):
+
+    mode = CONTINUOUS
+
+    name = 'arm_energy_probe'
+    description = """Collects power traces using the ARM Energy Probe.
```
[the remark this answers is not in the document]
vingu-linaro: ok

devlib/instrument/arm_energy_probe.py
```diff
@@ -0,0 +1,130 @@
...
+
+                     This instrument requires ``arm-probe`` utility to be installed in the workload automation
```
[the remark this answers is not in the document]
vingu-linaro: ok

devlib/instrument/arm_energy_probe.py
```diff
@@ -0,0 +1,130 @@
...
+                     host and be in the PATH. arm-probe is available here ``https://git.linaro.org/tools/arm-probe.git`` .
```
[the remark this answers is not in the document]
vingu-linaro: yes. I will do

devlib/instrument/parse_aep.py
```diff
@@ -0,0 +1,509 @@
...
+
+    def __del__(self):
+        self.unprepare()
+
+
+    @staticmethod
+    def topology_from_data(array, topo):
+    # Extract topology information for the data file
+    # The header of a data file looks like this:
+# configuration: <file path>
```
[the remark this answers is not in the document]
vingu-linaro: yes

devlib/instrument/parse_aep.py
vingu-linaro: ok. i'm going to have a look

devlib/instrument/arm_energy_probe.py
```diff
@@ -0,0 +1,130 @@
...
+import sys
+
+import tempfile
+import shutil
+
+from devlib.instrument import Instrument, CONTINUOUS, MeasurementsCsv
+from devlib.exception import HostError
+from devlib.utils.misc import which
+
+sys.path.append(os.path.dirname(sys.modules[__name__].__file__))
```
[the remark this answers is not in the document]
vingu-linaro: ok

devlib/instrument/arm_energy_probe.py
```diff
@@ -0,0 +1,130 @@
...
+        self.sample_rate_hz = 10000
+        self.config_file = config_file
+
+        self.parser = AEP_parser()
+        #TODO make it generic
+        topo = self.parser.topology_from_config(self.config_file)
+        for item in topo:
+            if item == 'time':
+                self.add_channel('timestamp', 'time')
+                continue
```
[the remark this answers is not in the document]
vingu-linaro: no it has to be a continue but it can be an else as well.
might be clearer with else ?
+            if item == 'time':
+                self.add_channel('timestamp', 'time')
￼+            else:
+                 self.add_channel(item, 'power')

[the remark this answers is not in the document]
setrofim: oh, right, I see. yes, with an `else` it would be clearer.

devlib/instrument/parse_aep.py
vingu-linaro: In fact, they are mandatory

devlib/instrument/parse_aep.py
vingu-linaro: ok

vingu-linaro:
I have just pushed a new version of the patch that normally takes into account your comments.
You are right that this provides support for the Energy Probe  but via arm-probe rather than caiman.
The arm-probe provides more features compared to caiman:
- Multi devices support
- aggregation of channels. We have some platforms with 2 power rails for the same voltage domain. Only the add of these 2 rails make sense
- auto zero channel
- Take into account power topology when computing energy/power

setrofim:
OK, if that is the case, do we still need the `caiman` version? Does that offer something this doesn't?

devlib/instrument/arm_energy_probe.py
```diff
@@ -0,0 +1,130 @@
...
+import shutil
+
+from devlib.instrument import Instrument, CONTINUOUS, MeasurementsCsv
+from devlib.exception import HostError
+from devlib.utils.misc import which
+
+from devlib.utils.parse_aep import AEP_parser
+
+class ArmEnergyProbeInstrument(Instrument):
+    """Collects power traces using the ARM Energy Probe.
```
setrofim: Please ensure the triple double (sextuple?) quotes are on a line of their own, and the docstring text is aligned with the quotes, not indented with respect to them.

devlib/utils/parse_aep.py
```diff
@@ -0,0 +1,528 @@
...
+import os
+import sys
+import getopt
+import subprocess
+import signal
+import serial
+import time
+import math
+
+class AEP_parser(object):
```
setrofim: No underscores in class names, acronyms should not be capitalised.  So should be `AepParser`

devlib/utils/parse_aep.py
```diff
@@ -0,0 +1,528 @@
...
+                self.fs = open(summaryfile, "w")
+            except IOError:
+                print "WARN: Unable to create %s" % (summaryfile)
+                self.fs = sys.stdout
+        else:
+                self.fs = sys.stdout
+
+        self.prepared = True
+
+
```
setrofim: Only single empty line between class members.

devlib/utils/parse_aep.py
```diff
@@ -0,0 +1,528 @@
...
+        self.fo.write(str(array[0]))
+        for i in range(1, len(array)):
+            if hide[i]:
+                continue
+
+            self.fo.write(" "+str(array[i]))
+
+        self.fo.write("\n")
+
+    def parse_AEP(self, start=0, lenght=-1):
```
setrofim: Method names should be all lower case: `parse_aep`

devlib/utils/parse_aep.py
```diff
@@ -0,0 +1,528 @@
...
+                # hide useless columns and detect channels that are children
+                # of other channels
+                hide, duplicate = self.filter_column(label, unit, topo)
+
+                # Create virtual power domains
+                virtual = self.create_virtual(topo, label, hide, duplicate)
+                if self.parse:
+                    self.output_label(label, hide)
+
+                if self.verbose:
```
setrofim: Do not do this. Instantiate a global logger and use `logger.debug`, allowing the calling code to deal with verbosity.

devlib/utils/parse_aep.py
```diff
@@ -0,0 +1,528 @@
...
+        # store AEP Platform channel info in the results table
+        results_table["Platform"] = total_nrj, average_power
+
+        return results_table
+
+    def topology_from_config(self, topofile):
+        try:
+            ft = open(topofile, "r")
+        except IOError:
+            print "WARN: Unable to open config file %s" % (topofile)
```
setrofim: `logger.warn'`

devlib/utils/parse_aep.py
```diff
@@ -0,0 +1,528 @@
...
+            # extract power figures
+            self.delta_nrj(data, nrj, min, max, hide)
+
+            # write data into new file
+            if self.parse:
+                self.output_power(data, hide)
+
+        # if there is no data just return
+        if label_line or len(nrj) == 1:
+            print "No data found in the data file. Please check the Arm Energy Probe"
```
setrofim: Raise an exception here.

devlib/utils/parse_aep.py
```diff
@@ -0,0 +1,528 @@
...
+            self.fo.close()
+
+        self.prepared = False
+
+
+    def __del__(self):
+        self.unprepare()
+
+
+    @staticmethod
```
setrofim: Should these statics just be internal utility function? If you want to keep them as as statics, please move them to above the `__init__`. 

Also, please move the `__del__` to below the regular methods. In general the order of class members should be:

1. class attributes
2. static/class methods
3. `__init__`
4. regular methods
5. internal (single leading underscore) methods
6. "special" (surrounded by double underscores) methods.

vingu-linaro:
Regarding caiman version, I would say that apart from being the official tools from Arm energy probe, I don't see other features that this tools doesn't provide

vingu-linaro:
Looks like I have done something wrong and close the pull request

PR #239, opened 2018-02-14
Add a new Arm Energy Probe instrument

vingu-linaro:
Add new energy instrument that is based on arm-probe tool to manage AEP
Main advantages of this tool are:
- uses a config file for describing channels and shunt resistors value
- manages power topology description in the config file. This topology
is then used when computing power figures
- can create virtual power channel and aggregate channels
- support multiple AEP
- support auto-zero of AEP's channel

Signed-off-by: Vincent Guittot <vincent.guittot@linaro.org>


vingu-linaro:
I did something wrong with the previous pull request and close it. This one is a new pull request and take into account comments done in previous one

PR #241, opened 2018-02-26
shutils.in: change cgroups_tasks_move filter

Elieva:
cgroups_tasks_move allow to exclude some tasks from the migration
by filtering them either by their PID or by the COMMAND name.
This patch allow to filter them as well by their CMDLINE.

Signed-off-by: Elieva Pignat <Elieva.Pignat@arm.com>

### DIFF
--- a/devlib/bin/scripts/shutils.in
+++ b/devlib/bin/scripts/shutils.in
@@ -157,7 +157,7 @@ cgroups_tasks_move() {
 
 	[ "x$FILTERS" = "x" ] && exit 0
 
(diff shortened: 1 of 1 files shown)


setrofim:
LGTM, but would be good to get a +1 from @derkling, @valschneider, or @qperret before merging.

derkling:
I've actually asked @Elieva about a similar feature, but having a better look at the code now, I think we should do it in a different way... as well as perhaps better specify the accepted values for the current `cgroups_tasks_move` implementation.

So, please hold on on merging this... we will come back shortly with some other feedback.

PR #242, opened 2018-02-27
cgroups.py: Add filters for the tasks() function

Elieva:
Closes #241 
The tasks() function allows to get the tasks that are in a cgroup.
Filters for the tasks TID, name and cmdline have been added to the
parameters of the function such that it is possible to select the tasks
that match these patterns.

Signed-off-by: Elieva Pignat <Elieva.Pignat@arm.com>

### DIFF
--- a/devlib/module/cgroups.py
+++ b/devlib/module/cgroups.py
@@ -14,6 +14,7 @@
 #
 # pylint: disable=attribute-defined-outside-init
 import logging
+import re
 from collections import namedtuple
 
 from devlib.module import Module
@@ -168,7 +169,37 @@ class Controller(object):
             if cgroup != dest:
                 self.move_tasks(cgroup, dest, grep_filters)
 
-    def tasks(self, cgroup):
+    def tasks(self, cgroup,
+              filter_tid='',
+              filter_tname='',
+              filter_tcmdline=''):
+        """
+        Report the tasks that are included in a cgroup. The tasks can be
+        filtered by their tid, tname or tcmdline if filter_tid, filter_tname or
+        filter_tcmdline are defined respectively. In this case, the reported
+        tasks are the ones in the cgroup that match these patterns.
+
+        Example of tasks format:
+        TID,tname,tcmdline
+        903,cameraserver,/system/bin/cameraserver
+
+        :params filter_tid: regexp pattern to filter by TID
+        :type filter_tid: str
+
+        :params filter_tname: regexp pattern to filter by tname
+        :type filter_tname: str
+
+        :params filter_tcmdline: regexp pattern to filter by tcmdline
+        :type filter_tcmdline: str
+
+        :returns: a dictionary in the form: {tid:(tname, tcmdline)}
+        """
+        if not isinstance(filter_tid, str):
+            raise TypeError('filter_tid should be a str')
+        if not isinstance(filter_tname, str):
+            raise TypeError('filter_tname should be a str')
+        if not isinstance(filter_tcmdline, str):
+            raise TypeError('filter_tcmdline should be a str')
         try:
             cg = self._cgroups[cgroup]
         except KeyError as e:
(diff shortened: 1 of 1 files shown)


devlib/module/cgroups.py
```diff
@@ -168,7 +169,31 @@ def move_all_tasks_to(self, dest, exclude=[]):
...
+        Report the tasks that are included in a cgroup. The tasks can be
+        filtered by their tid, tname or tcmdline if filter_tid, filter_tname or
+        filter_tcmdline are defined respectively. In this case, the reported
+        tasks are the ones in the cgroup that match these patterns.
+
+        Example of tasks format:
+        TID,tname,tcmdline
+        903,cameraserver,/system/bin/cameraserver
+
+        :params filter_tid: pattern to filter by TID
```
derkling: Maybe we should specify that's a regexp?

devlib/module/cgroups.py
```diff
@@ -180,13 +205,19 @@ def tasks(self, cgroup):
         tasks = {}
         for task in entries:
             tid = task.split(',')[0]
+            if filter_tid and not re.search(filter_tid, tid):
```
derkling: Just to be a bit more sure it's a "sort of regexp, should we use instead:
```python
if type(filter_tid) != str and re.search(filter_tid, tid):
```

devlib/module/cgroups.py
```diff
@@ -180,13 +205,19 @@ def tasks(self, cgroup):
...
             try:
                 tname = task.split(',')[1]
+                if filter_tname and not re.search(filter_tname, tname):
+                    continue
             except: continue
             try:
                 tcmdline = task.split(',')[2]
             except:
                 tcmdline = ''
+            if filter_tcmdline and not re.search(filter_tcmdline, tcmdline):
```
derkling: This should be in the try block above... we don't want to filter the empty string.

devlib/module/cgroups.py
derkling: Should we also add a sanity check? It should be either None or a string...

devlib/module/cgroups.py
derkling: You can skip the check on type if you do the sanity check at the beginning...

devlib/module/cgroups.py
Elieva: specify that is a regex: Done
i have added sanity checks at the start of the function for the filter. I change as well the default value of the filter for the empty string.

devlib/module/cgroups.py
Elieva: Done: sanity check at the start of the function

devlib/module/cgroups.py
Elieva: I am not sure about this one. If someone wants to filter some pattern in the cmdline, it makes sense to me that a task that has no cmdline should not be in the result.

devlib/module/cgroups.py
```diff
@@ -180,13 +211,19 @@ def tasks(self, cgroup):
         tasks = {}
         for task in entries:
```
setrofim: I would re-write this loop to split only once for each task at the beginning: 

```
tid_str, tname, tcmdline = task.split(',', 2)
tid = int(tid_str)
```

(also note the limit on the split to handle the case where cmdline itself contains commas).

devlib/module/cgroups.py
```diff
@@ -168,7 +169,37 @@ def move_all_tasks_to(self, dest, exclude=[]):
...
+        :type filter_tcmdline: str
+
+        :returns: a dictionary in the form: {tid:(tname, tcmdline)}
+        """
+        if type(filter_tid) != str:
+            raise TypeError('filter_tid should be a str')
+        if type(filter_tname) != str:
+            raise TypeError('filter_tname should be a str')
+        if type(filter_tcmdline) != str:
+            raise TypeError('filter_tcmdline should be a str')
```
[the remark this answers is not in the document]
Elieva: I think it is better to keep filter_tid as a string such that a regexp can be given which allows more flexible filtering (not only equality)

Change type() by isinstance(): Done

devlib/module/cgroups.py
Elieva: Done

devlib/module/cgroups.py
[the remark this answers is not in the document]
setrofim: >I think it is better to keep filter_tid as a string such that a regexp can be given which allows more flexible filtering (not only equality)

OK, makes sense.

PR #245, opened 2018-03-07
gem5: Update to telnet connection method

avanlaer:
gem5 prints a string, to communicate which telnet port can be used to connect
to the gem5 system.  The exact string has recently changed. This commit ensures
both the old and new string are taken into account when devlib parses gem5
output to determine the telnet port.

### DIFF
--- a/devlib/platform/gem5.py
+++ b/devlib/platform/gem5.py
@@ -174,7 +174,11 @@ class Gem5SimulationPlatform(Platform):
             # Open the stderr file
             with open(self.stderr_filename, 'r') as f:
                 for line in f:
+                    # Look for two different strings, exact wording depends on
+                    # version of gem5
                     m = re.search(r"Listening for system connection on port (?P<port>\d+)", line)
+                    if not m:
+                        m = re.search(r"Listening for connections on port (?P<port>\d+)", line)
                     if m:
                         port = int(m.group('port'))
                         if port >= 3456 and port < 5900:


avanlaer:
I cannot immediately see how to add a reviewer so, @marcbonnici, @Sticklyman1936  do you mind looking at this? :)

devlib/platform/gem5.py
```diff
@@ -174,7 +174,11 @@ def _intercept_telnet_port(self):
             # Open the stderr file
             with open(self.stderr_filename, 'r') as f:
                 for line in f:
+                    # Look for two different strings, exact wording depends on
+                    # version of gem5
                     m = re.search(r"Listening for system connection on port (?P<port>\d+)", line)
+                    if not m:
+                        m = re.search(r"Listening for connections on port (?P<port>\d+)", line)
```
marcbonnici: Could condense these into a single expression, e.g. something like:
`Listening for (system )?connection[s]? on port (?P<port>\d+)`

avanlaer: To be fair, I like splitting it up in 2 lines, because then it is more readable, if you would be okay with that?

setrofim:
I think this is OK to be merged. IIRC, @Sticklyman1936  has looked at this last week, he just didn't ACK it in the interface.

avanlaer:
Yes, sounds good to me. Sorry, didn't see the thumbs up earlier.

Sticklyman1936:
LGTM

PR #255, opened 2018-04-06
utils/android: Append platform-tools at the end of PATH

JaviMerino:
devlib currently prepends platform-tools at the beginning of PATH.
This makes WA3 use a different adb (the one in platform-tools) than
the one I use everywhere else.  Because they are different, whenever
the "other" adb starts, it kills the sessions of the previous one.

Instead of prepending and using a different adb than the reset of the
system, append platform-tools to PATH, so that the adb used by WA3 and
the rest of the system is the same.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -491,7 +491,7 @@ def _initialize_with_android_home(env):
     logger.debug('Using ANDROID_HOME from the environment.')
     env.android_home = android_home
     env.platform_tools = os.path.join(android_home, 'platform-tools')
-    os.environ['PATH'] = env.platform_tools + os.pathsep + os.environ['PATH']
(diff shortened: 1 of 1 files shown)


setrofim:
The pre-pending is intentional. The devlib dependency is on an Android SDK -- it discovers an SDK via `$ANDROID_HOME` and uses the tools with in that SDK. This should continue to be the default behaviour. If you would like to use a tool outside the SDK, that is an exceptional condition that should be explicitly configured for your environment, not  the default;

derkling:
I've also get used to set `$ANDROID_HOME` in my env and that's actually quite handy when it comes down to prefer my own SDK instead of the, possibly outdated, one installed by a distro.

PR #259, opened 2018-04-11
devlib/utils/ssh.py: Fix set terminal window size

alimon:
Ensure that the terminal window size is set to 500x200, if not long
commands (prompt + command > 80 chars) will fail because the pexpect
search to match the command string in order to get the result and the
command is wrapped to first 80 chars.

In order to fix this scenario enables checkwinsize in the shell and use
stty to set the new terminal window size.

Signed-off-by: Aníbal Limón <anibal.limon@linaro.org>

### DIFF
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -72,8 +72,14 @@ def ssh_get_shell(host, username, password=None, keyfile=None, port=None, timeou
                 raise TargetError(message.format(host))
             time.sleep(5)
 
+    conn.sendline('shopt -s checkwinsize')
+    conn.prompt()
     conn.setwinsize(500,200)
     conn.sendline('')
+    conn.prompt()
+    conn.sendline('stty rows 500')
+    conn.prompt()
+    conn.sendline('stty cols 200')
     conn.prompt()
     conn.setecho(False)
     return conn


alimon:
Related to,

https://github.com/ARM-software/workload-automation/pull/725

devlib/utils/ssh.py
```diff
@@ -72,9 +72,15 @@ def ssh_get_shell(host, username, password=None, keyfile=None, port=None, timeou
                 raise TargetError(message.format(host))
             time.sleep(5)
 
+    conn.sendline('shopt -s checkwinsize')
```
douglas-raillard-arm: Since we hit similar issue I had a look at this PR, and `shopt` is a bash-specific builtin, which does not exist in the `busybox` shell AFAIK (or even zsh, which uses `set` for similar purposes). I don't think that bash feature has equivalent in all shells, so it should probably not be relied on.

douglas-raillard-arm:
@marcbonnici That code has been replaced by the paramiko connection for quite a while now, so I think we can close this PR. A similar problem probably exists in the telnet connection but I honestly don't think telnet works at all in the current state.

If we wanted to resurrect it, we would need to be able to multiplex multiple command output in the same terminal (for background command API), which is challenging. And at that rate, it may just be simpler to setup SLIP (IP over serial port) and use SSH

douglas-raillard-arm:
Closing as:
* the PR hasn't been updated in 6 years
* this relies on a non-portable feature of bash
* SSH connection does not use a terminal anymore and therefore completely avoids the window size problem. The problem may still exist for other connections such as telnet but I don't think they are actually working anymore and will likely be removed.

PR #268, opened 2018-05-30
cgroups: add execute as root

vingu-linaro:
On latest hikey960 android image, only root can execute: cat /proc/cgroups

Signed-off-by: Vincent Guittot <vincent.guittot@linaro.org>

### DIFF
--- a/devlib/module/cgroups.py
+++ b/devlib/module/cgroups.py
@@ -394,7 +394,7 @@ class CgroupsModule(Module):
     def list_subsystems(self):
         subsystems = []
         for line in self.target.execute('{} cat /proc/cgroups'\
-                .format(self.target.busybox)).splitlines()[1:]:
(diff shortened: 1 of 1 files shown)


devlib/module/cgroups.py
```diff
@@ -394,7 +394,7 @@ def __init__(self, target):
     def list_subsystems(self):
         subsystems = []
         for line in self.target.execute('{} cat /proc/cgroups'\
-                .format(self.target.busybox)).splitlines()[1:]:
+                .format(self.target.busybox), as_root=True).splitlines()[1:]:
```
setrofim: Please change to `as_root=self.target.is_rooted`, so that this will continue to work for unrooted targets that do not require root to read `/proc/cgroups`.

vingu-linaro:
I have updated the patch

PR #270, opened 2018-05-31
[RFC] Add support for Python 3

setrofim:
Add support for running on Python 3 while maintaining Python 2
compatibility.

NOTE: this introduces a new dependency on the `future` package (which provides 2-3 interop)

This has been tested with WA Python 3 port (which exercises a large portion of
the devlib API) with Android and Linux targets, running on both 2.7 and 3.6 Python versions.


setrofim:
@derkling, @valschneider: would be cool if you could test this with your existing infrastructure to make sure I haven't inadvertently broke something for Python 2.

setrofim:
Further testing has revealed some issues pertaining to the processing of CSV files, which have now been fixed.

PR #273, opened 2018-06-06
Add Juno Shell Prompt Regex

sdpenguin:
Also checks for 'juno' in addition to 'root' and 'shell'.
Concurrent support added for parsing target params in the platform tuple
in WA which makes use of this.


setrofim:
"juno" is already part of the `DEFAULT_SHELL_PROMPT`

PR #274, opened 2018-06-07
Add gem5 checkpoint support

Sticklyman1936:
Issues have been addressed, and the code has been tested from both WA and from a customised python script directly utilising devlib.

### DIFF
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -532,6 +532,10 @@ class Gem5Connection(TelnetConnection):
         """
         gem5_logger.info("Gracefully terminating the gem5 simulation.")
         try:
+            # Unmount the virtio device BEFORE we kill the
+            # simulation. This is done to simplify checkpointing at
+            # the end of a simulation!
+            self._unmount_virtio()
             self._gem5_util("exit")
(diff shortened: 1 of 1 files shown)


setrofim:
This looks reasonable to me, given the current code structure. One *could* monkey-patch `take_checkpoint` into the target, but I'm not sure it's worth it.

As we discussed offline, `Gem5Connection` should probably be split up into separate "connection" and "simulation" objects, the latter being initialized and installed into the Target by the Plaform. When the code has been thus refactored, `take_checkpoint` would be part of the simulation, rather than a connection.

For now, I would be happy to pull these changes as-is.

devlib/utils/ssh.py
```diff
@@ -532,6 +532,10 @@ def close(self):
         """
         gem5_logger.info("Gracefully terminating the gem5 simulation.")
         try:
+            # Unmount the virtio device BEFORE we kill the
+            # simulation. This is done to simplify checkpointing at
+            # the end of a simulation!
+            target._unmount_virtio()
```
marcbonnici: This should probably be `self._unmount_virtio()`?

Issue #276, opened 2018-06-07
Instrument.reset does not populate self.active_channels for string kinds


pietos01:
Calling `Instrument.reset(kinds='some string')` leaves `self.active_channels` as `[]` which is probably not the expected behaviour. This is caused by [the last nested `else`](https://github.com/ARM-software/devlib/blob/fe0d6eda2acb7345e0499943e8effac187aac91c/devlib/instrument/__init__.py#L319) which refers to [the condition `if isinstance(kinds, basestring)`](https://github.com/ARM-software/devlib/blob/fe0d6eda2acb7345e0499943e8effac187aac91c/devlib/instrument/__init__.py#L317) and might have been overlooked because of having been confused with [the top-level `else`](https://github.com/ARM-software/devlib/blob/fe0d6eda2acb7345e0499943e8effac187aac91c/devlib/instrument/__init__.py#L314). Anyhow, an `else` does not seem to be needed there.

PR #277, opened 2018-06-07
Instrument.reset: fix when called with string kind & improve readability

pietos01:
Issue #276 illustrates the risk of having too many nested levels and
execution paths which also impact the readability of the code. We
modify the implementation to solve the bug on top of which we:

  - Reduce the maximum order of nested levels from 4 to 3;

  - Express more clearly the potential paths of execution
    (less nested conditions);

  - Replace unnecessary `for`-loops by list comprehensions,
    removing the need for an initialisation of `active_channels`
    and making clearer what each path of execution ends up with;

  - Removed unnecessary `list` copies of `self.channels.values()`;

  - Used the fact that the message of a `KeyError` is the unknown
    key.

Some comments have also been added pointing out places in the code
where the logger could have been called: these are proposition to
be accepted (and translated into code) or removed in an upcoming
commit.

We feel like an exception is unnecessarily raised in the current implementation and would propose to remove it; this is a tough call as it is possible (although unlikely) that user code relies on this exception in a `try` clause ...

### DIFF
--- a/devlib/instrument/__init__.py
+++ b/devlib/instrument/__init__.py
@@ -1,4 +1,4 @@
-#    Copyright 2015 ARM Limited
+#    Copyright 2018 ARM Limited
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -298,29 +298,26 @@ class Instrument(object):
         pass
 
     def reset(self, sites=None, kinds=None, channels=None):
-        self.active_channels = []
-        if kinds is None and sites is None and channels is None:
-            self.active_channels = sorted(self.channels.values(), key=lambda x: x.label)
-        elif channels is not None:
+        if channels is not None:
             if sites is not None or kinds is not None:
-                raise ValueError(
-                    'sites and kinds should not be set if channels is set')
-            for chan_name in channels:
-                try:
-                    self.active_channels.append(self.channels[chan_name])
-                except KeyError:
-                    msg = 'Unexpected channel "{}"; must be in {}'
-                    raise ValueError(msg.format(chan_name, self.channels.keys()))
+                raise ValueError('sites and kinds should not be set if channels is set')
+
+            try:
+                self.active_channels = [self.channels[ch] for ch in channels]
+            except KeyError as e:
+                msg = 'Unexpected channel "{}"; must be in {}'
+                raise ValueError(msg.format(e, self.channels.keys()))
+        elif sites is None and kinds is None:
+            self.active_channels = sorted(self.channels.itervalues(), key=lambda x: x.label)
         else:
             if isinstance(sites, basestring):
                 sites = [sites]
             if isinstance(kinds, basestring):
                 kinds = [kinds]
-            else:
-                for chan in self.channels.values():
-                    if (kinds is None or chan.kind in kinds) and \
-                       (sites is None or chan.site in sites):
-                        self.active_channels.append(chan)
+
+            wanted = lambda ch : ((kinds is None or ch.kind in kinds) and
+                                  (sites is None or ch.site in sites))
+            self.active_channels = filter(wanted, self.channels.itervalues())
 
     # instantaneous
 


setrofim:
> We feel like an exception is unnecessarily raised in the current implementation and would propose to remove it;

This is library code, it has no awareness of the context it's running in, and therefore should make no assumptions about how an exceptional condition should be handled -- it's up to the calling code to decide whether the this should be ignored, logged, propagated, handled in some other way.

pietos01:
The commits were squashed into a single one. Sorry about this.

> This is library code, it has no awareness of the context it's running in, and therefore should make no assumptions about how an exceptional condition should be handled

Indeed, this is undeniable. The point was about how we define _an exceptional condition_; it is common to ignore function parameters based on the value of other parameters. In this case, any of `kinds` or `sites` being non-`None` does not prevent `reset()` from doing its job. Therefore, it may be argued that the caller does not even deserve an exception.

That being said, the exception (which, to be clear, I am not saying is wrong) will be kept and [the commented question](https://github.com/pietos01/devlib/blob/0d26523f3c762902219e04615488bb868a59dc97/devlib/instrument/__init__.py#L302) will be removed.

pietos01:
What about logging? I see that `Instrument.__init__()` instantiates a `logger` which is not used in the rest of the class (to be fair, there is not much to log outside of `reset()` anyway). Is it expected that this `logger` should only be used by child classes inheriting from `Instrument` (_e.g._ `AcmeCapeInstrument` uses it for [info](https://github.com/ARM-software/devlib/blob/fe0d6eda2acb7345e0499943e8effac187aac91c/devlib/instrument/acmecape.py#L103), [debug](https://github.com/ARM-software/devlib/blob/fe0d6eda2acb7345e0499943e8effac187aac91c/devlib/instrument/acmecape.py#L78), [warning](https://github.com/ARM-software/devlib/blob/fe0d6eda2acb7345e0499943e8effac187aac91c/devlib/instrument/acmecape.py#L114) and [error](https://github.com/ARM-software/devlib/blob/fe0d6eda2acb7345e0499943e8effac187aac91c/devlib/instrument/acmecape.py#L94) messages)?

setrofim:
>  In this case, any of kinds or sites being non-None does not prevent reset() from doing its job. 
Yes it does. The job of `reset()` among other things is to configure `active_channels`. This is done either via `sites` and `kinds` (indicating "I want all channels for these sites/kinds") or via `channels` (indicating "I want *only* these channels"). So if both `channels` *and* `sites`/`kinds`  are specified, the user's intent is ambiguous.

setrofim:
>  Is it expected that this logger should only be used by child classes inheriting from Instrument (e.g. AcmeCapeInstrument uses it for info, debug, warning and error messages) and not Instrument itself?

It *can* be used the `Instrument` where appropriate. In case of `AcmeCapeInstrument`, it logs errors in communicating with iio-capture, which is part of its implementation, and not something the calling code has any control over. Note that even in that case, the error still gets propagated as an exception. The reason it is logged at that  level, is that that there is a chance that it won't be propagated, if it is decided that sufficient output has been generated to attempt further processing; in which case it may be propagated later, but we still want to log here to indicate where the error originated. This is a somewhat unusual scenario as we typically want to fail early.

In the case of `reset`, the error originates from the arguments passed by the calling code. It is the calling code's responsibility to ensure that the arguments it passes into `reset` make sense, and if they don't, an error should be raised.

setrofim:
> it is common to ignore function parameters based on the value of other parameters.

This is a terrible idea. One should never just ignore parameter specified by the caller. If a caller explicitly specifies something, the assumption should be that they expect you do to something sensible with it; if you can't, you should raise an error. Doing anything else may result in outcomes being subtly different from what the caller expects and the rest of the code working under false assumptions.

For example, in this case -- should `sites`/`kinds` or `channels` take precedence when all are specified? Different individual will have different ideas about what is "intuitive" or "obvious".  One could, of course, specify this in the documentation, but people don't read the documentation. The safe thing to do is to force the calling code be unambiguous in what it wants.

devlib/instrument/__init__.py
```diff
@@ -298,29 +298,28 @@ def teardown(self):
...
             if sites is not None or kinds is not None:
-                raise ValueError(
-                    'sites and kinds should not be set if channels is set')
-            for chan_name in channels:
-                try:
-                    self.active_channels.append(self.channels[chan_name])
-                except KeyError:
-                    msg = 'Unexpected channel "{}"; must be in {}'
-                    raise ValueError(msg.format(chan_name, self.channels.keys()))
+		# self.logger.error(...) ?
```
setrofim: Please remove

devlib/instrument/__init__.py
```diff
@@ -298,29 +298,28 @@ def teardown(self):
...
+                raise ValueError('sites and kinds should not be set if channels is set')
+
+            try:
+                self.active_channels = [self.channels[ch] for ch in channels]
+            except KeyError as e:
+                msg = 'Unexpected channel "{}"; must be in {}'
+		# self.logger.error(...) ?
```
setrofim: Please remove.

PR #281, opened 2018-06-13
Fix Python3 Byte Handling

sdpenguin:
Convert bytes to strings (utf-8 encoding) to make compatible with
Python3 in arm.py
Drop problematic characters when decoding stdout and stderr in misc.py
by setting errors='replace' in the string decode method.

### DIFF
--- a/devlib/platform/arm.py
+++ b/devlib/platform/arm.py
@@ -14,6 +14,7 @@
 #
 from __future__ import division
 import os
+import sys
 import tempfile
 import time
 import pexpect
--- a/devlib/platform/arm.py
+++ b/devlib/platform/arm.py
@@ -88,6 +88,9 @@ class VersatileExpressPlatform(Platform):
     def _init_android_target(self, target):
         if target.connection_settings.get('device') is None:
             addr = self._get_target_ip_address(target)
+            if sys.version_info[0] == 3:
+                # Convert bytes to string for Python3 compatibility
+                addr = addr.decode("utf-8")
             target.connection_settings['device'] = addr + ':5555'
 
     def _init_linux_target(self, target):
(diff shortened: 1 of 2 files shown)


devlib/platform/arm.py
```diff
@@ -88,6 +88,8 @@ def init_target_connection(self, target):
     def _init_android_target(self, target):
         if target.connection_settings.get('device') is None:
             addr = self._get_target_ip_address(target)
+            # Convert bytes to string for Python3 compatibility
+            addr = addr.decode("utf-8")
```
setrofim: This needs to be guarded by `if sys.version_info[0] == 3:`

devlib/platform/arm.py
```diff
@@ -103,7 +105,9 @@ def _get_target_ip_address(self, target):
             tty.sendline('su')  # this is, apprently, required to query network device
                                 # info by name on recent Juno builds...
             self.logger.debug('Waiting for the Android shell prompt.')
-            tty.expect(target.shell_prompt)
+            # In Python3 pexpect does not accept bytes-like objects, so convert to str
+            shell_prompt_string = target.shell_prompt.pattern
+            tty.expect(shell_prompt_string)
```
setrofim: Erm I'm unclear on the need for this -- `target.shell_prompt` is a regex, not a bytes-like object.

sdpenguin: Yes, it may not be a bytes-like object, but it is not a string, and the Python3 version of tty.expect requires a string input, so we need to extract the string using the pattern property.

devlib/platform/arm.py
sdpenguin: Done

devlib/platform/arm.py
setrofim: According to the docs: 

http://pexpect.readthedocs.io/en/stable/api/pexpect.html#pexpect.spawn.expect

the pattern argument could be a compiled re.

sdpenguin: For some reason, it doesn't work and gives:
TypeError(cannot use a string pattern on a bytes-like object)

PR #283, opened 2018-06-13
Fix shell_prompt matching on Python 3

setrofim:


devlib/utils/types.py
```diff
@@ -124,3 +125,38 @@ def bitmask(value):
     if not isinstance(value, int):
         raise ValueError(value)
     return value
+
+
+regex_type = type(re.compile(''))
```
sdpenguin: Don't forget 'import re'

devlib/utils/types.py
```diff
@@ -146,7 +147,8 @@ def bytes_regex(value):
         if isinstance(value, regex_type):
             if isinstance(value.pattern, bytes):
                 return value
-            return re.compile(value.pattern.encode(sys.stdout.encoding))
+            return re.compile(value.pattern.encode(sys.stdout.encoding),
+                              value.flags)
```
sdpenguin: ValueError(cannot use UNICODE flag with a bytes pattern)

PR #286, opened 2018-06-14
PEP 396 + commit ID inside installed version

setrofim:


setup.py
```diff
@@ -37,6 +38,26 @@
 except OSError:
     pass
 
+
+with open(os.path.join(devlib_dir, '__init__.py')) as fh:
+    # exract the version by parsing the text of the file,
```
Sticklyman1936: s/exract/extract/g

PR #287, opened 2018-06-15
Sched tunables

derkling:
Add some new SchedModule APIs to get/set scheduler attributes and features.

### DIFF
--- a/devlib/module/sched.py
+++ b/devlib/module/sched.py
@@ -253,6 +253,58 @@ class SchedModule(Module):
 
         return SchedProcFSData.available(target)
 
+    def get_kernel_attributes(self, matching=None, check_exit_code=True):
+        """
+        Get the value of scheduler attributes.
+
+        :param matching: an (optional) substring to filter the scheduler
+        attributes to be returned.
+
+        The scheduler exposes a list of tunable attributes under:
+            /proc/sys/kernel
+        all starting with the "sched_" prefix.
+
+        This method returns a dictionary of all the "sched_" attributes exposed
+        by the target kernel, within the prefix removed.
+        It's possible to restrict the list of attributes by specifying a
+        substring to be matched.
+
+        returns: a dictionary of scheduler tunables
+        """
+        command = 'sched_get_kernel_attributes {}'.format(
+            matching if matching else ''
+        )
+        output = self.target._execute_util(command, as_root=self.target.is_rooted,
+                                           check_exit_code=check_exit_code)
+        result = {}
+        for entry in output.strip().split('\n'):
+            if ':' not in entry:
+                continue
+            path, value = entry.strip().split(':', 1)
+            if value in ['0', '1']:
+                value = bool(int(value))
+            elif value.isdigit():
+                value = int(value)
+            result[path] = value
+        return result
+
+    def set_kernel_attribute(self, attr, value, verify=True):
+        """
+        Set the value of a scheduler attribute.
+
+        :param attr: the attribute to set, without the "sched_" prefix
+        :param value: the value to set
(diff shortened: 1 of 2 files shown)


devlib/module/sched.py
```diff
@@ -268,6 +268,58 @@ def set_kernel_attribute(self, attr, value, verify=True):
         path = '/proc/sys/kernel/sched_' + attr
         self.target.write_value(path, value, verify)
 
+    @property
+    @memoized
+    def has_debug(self):
+        if self.target.config.get('SCHED_DEBUG') != 'y':
+            logger.warn("target's kernel compiled with !CONFIG_SCHED_DEBUG")
```
setrofim: Why warn here? Should `SCHED_DEBUG` part of the condition for returning `True`?

devlib/module/sched.py
```diff
@@ -268,6 +268,58 @@ def set_kernel_attribute(self, attr, value, verify=True):
...
+        :returns: a dictionary of features and their "is enabled" status
+        """
+        if not self.has_debug:
+            raise RuntimeError("sched_features not available")
+        feats = self.target.read_value('/sys/kernel/debug/sched_features')
+        features = {}
+        for feat in feats.split():
+            value = True
+            if feat.startswith('NO'):
+                feat = feat.replace('NO_', '')
```
setrofim: Change to `feat.replace('NO_', '', 1)` to guard against the case where a feature has `NO_` somewhere in its name.

devlib/module/sched.py
```diff
@@ -268,6 +268,58 @@ def set_kernel_attribute(self, attr, value, verify=True):
...
+
+        :param feature: the feature name to set
+        :param enable: true to enable the feature, false otherwise
+
+        :raise ValueError: if the specified enable value is not bool
+        :raise RuntimeError: if the specified feature cannot be set
+        """
+        if not self.has_debug:
+            raise RuntimeError("sched_features not available")
+        if not isinstance(enable, bool):
```
setrofim: Maybe coerce to `boolean`  (https://github.com/ARM-software/devlib/blob/master/devlib/utils/types.py#L44) here instead?

devlib/module/sched.py
```diff
@@ -268,6 +268,58 @@ def set_kernel_attribute(self, attr, value, verify=True):
...
+            raise ValueError('enable parameter can only be boolean')
+        feat_value = feature.upper()
+        if enable is False:
```
setrofim: `if not enable:` would be more Pythonic

devlib/module/sched.py
derkling: You can have 'file_exists' failing but just because the debugfs is not mounted... this was an attemp to warn the use in case debugfs is not available... but perhaps we should move it into a dedicated call and avoid explicit warnings?

devlib/module/sched.py
derkling: Good point!

devlib/module/sched.py
derkling: Right, good point.

devlib/module/sched.py
derkling: yes... although sometimes I've got the impression that if we wanna ensure a specific value/object (i.e. a boolean false in this case) it's usually better to check it explicit. I'm always a bit confused on that.
But I guess here we are ensure by the before coerce to boolean that this `not` will mean `False`.

setrofim: Yes, given the coercion above, there is no practical difference. My comment was purely stylistic.

derkling:
This seems to address all the changes required... although github keep reporting "chages requested"... not sure why :/

PR #288, opened 2018-06-20
Few additions to AndroidTarget

valschneider:
A collection of bits and bobs we have laying in LISA but that really belongs here.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1155,23 +1155,56 @@ class AndroidTarget(Target):
 
     # Android-specific
 
+    def input_tap(self, x, y):
+        command = 'input tap {} {}'
+        self.execute(command.format(x, y))
+
+    def input_tap_pct(self, x, y):
+        width, height = self.screen_resolution
+
+        x = (x * width) // 100
+        y = (y * height) // 100
+
+        self.input_tap(x, y)
+
+    def input_swipe(self, x1, y1, x2, y2):
+        """
+        Issue a swipe on the screen from (x1, y1) to (x2, y2)
+        Uses absolute screen positions
+        """
+        command = 'input swipe {} {} {} {}'
+        self.execute(command.format(x1, y1, x2, y2))
+
+    def input_swipe_pct(self, x1, y1, x2, y2):
+        """
+        Issue a swipe on the screen from (x1, y1) to (x2, y2)
+        Uses percent-based positions
+        """
+        width, height = self.screen_resolution
+
+        x1 = (x1 * width) // 100
+        y1 = (y1 * height) // 100
+        x2 = (x2 * width) // 100
+        y2 = (y2 * height) // 100
+
+        self.input_swipe(x1, y1, x2, y2)
+
     def swipe_to_unlock(self, direction="diagonal"):
         width, height = self.screen_resolution
-        command = 'input swipe {} {} {} {}'
         if direction == "diagonal":
             start = 100
             stop = width - start
             swipe_height = height * 2 // 3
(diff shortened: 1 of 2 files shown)


devlib/target.py
```diff
@@ -934,6 +934,10 @@ class AndroidTarget(Target):
     os = 'android'
     ls_command = ''
 
+    # See https://developer.android.com/reference/android/content/Intent.html#setFlags(int)
```
valschneider: I put those here but I'm not sure it's the best place. Should that live in utilities/android ?

devlib/target.py
```diff
@@ -1155,23 +1159,57 @@ def capture_screen(self, filepath):
...
     # Android-specific
 
+    def input_tap(self, x, y):
+        command = 'input tap {} {}'
+        self.execute(command.format(x, y))
+
+    def input_tap_pct(self, x, y):
+        width, height = self.screen_resolution
+
+        x = (x * width) / 100
```
valschneider: Eeep, I just realised those want to be `//`. Thanks Python3...

devlib/target.py
```diff
@@ -1155,23 +1159,57 @@ def capture_screen(self, filepath):
...
+        x1 = (x1 * width) // 100
+        y1 = (y1 * height) // 100
+        x2 = (x2 * width) // 100
+        y2 = (y2 * height) // 100
+
+        self.input_swipe(x1, y1, x2, y2)
+
     def swipe_to_unlock(self, direction="diagonal"):
         width, height = self.screen_resolution
         command = 'input swipe {} {} {} {}'
```
marcbonnici: `command` can now be removed from here.

devlib/target.py
derkling: Since this is a collection of Intent's flags... should not this belong to an "intent" module (meaning python module, not devlib)? Something that along time can be improved to be a proxy of the link you reported above.

valschneider:
- Killed the stray `command` spotted by @marcbonnici

devlib/target.py
valschneider: Those would definitely benefit from being properly separated. *utills/android.py* has some definitions of Android versions, but IMO those flags are more meant for internal use whereas it makes sense to expose Android versions. 

As for a dedicated intent module, I'm not sure if it's worth creating one just for them. @marcbonnici / @setrofim, your thoughts on where to store these nasties ?

marcbonnici: It would certainly be nice to have somewhere to contain these nicely however I'm inclined to agree that it might not be worth creating something external for just these couple of values unless it is likely that many more of these will be required in the future and in other locations?

I might suggest moving them to the beginning of the file with the other global variables and adding a prefix to identify them rather than have them as class attributes for now? Although I'm not sure if that would make things much cleaner.

setrofim: I think that, for now, `utils/android.py` is the right place for them, as that's where we stick the various bits of Android miscellanea. In the future, it might make sense to  reorganise things further. However, I agree with @valschneider  and @marcbonnici that it's probably not worth creating a separate module just for those (e.g. at the moment, it's not clear whether it would make more sense to have `utils/android/intent.py` or `utils/android/flags.py`).

valschneider:
I moved the flags in an `INTENT_FLAGS` dict in **utils/android.py**:

```python
INTENT_FLAGS = {
    'ACTIVITY_NEW_TASK' : 0x10000000,
    'ACTIVITY_CLEAR_TASK' : 0x00008000
}
```

derkling:
LGTM

PR #291, opened 2018-06-22
module: Add devfreq module

mdigiorgio:
Add support for devfreq. This is used for example to get/set the frequency
and/or governor of the GPU.

### DIFF
--- a/devlib/module/devfreq.py
+++ b/devlib/module/devfreq.py
@@ -1,0 +1,261 @@
+#    Copyright 2018 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from devlib.module import Module
+from devlib.exception import TargetError
+from devlib.utils.misc import memoized
+
+class DevfreqModule(Module):
+
+    name = 'devfreq'
+
+    @staticmethod
+    def probe(target):
+        path = '/sys/class/devfreq/'
+        if not target.file_exists(path):
+            return False
+
+        # Check that at least one policy is implemented
+        if not target.list_directory(path):
+            return False
+
+        return True
+
+    @memoized
+    def list_devices(self):
+        """Returns a list of devfreq devices supported by the target platform."""
+        sysfile = '/sys/class/devfreq/'
+        return self.target.list_directory(sysfile)
+
+    @memoized
+    def list_governors(self, device):
+        """Returns a list of governors supported by the device."""
+        sysfile = '/sys/class/devfreq/{}/available_governors'.format(device)
+        output = self.target.read_value(sysfile)
+        return output.strip().split()
+
+    def get_governor(self, device):
+        """Returns the governor currently set for the specified device."""
+        if isinstance(device, int):
+            device = 'device{}'.format(device)
+        sysfile = '/sys/class/devfreq/{}/governor'.format(device)
+        return self.target.read_value(sysfile)
+
+    def set_governor(self, device, governor):
+        """
+        Set the governor for the specified device.
+
+        :param device: The device for which the governor is to be set. This must be
+                    the full name as it appears in sysfs, e.g. "e82c0000.mali".
+        :param governor: The name of the governor to be used. This must be
+                         supported by the specific device.
+
+        Additional keyword arguments can be used to specify governor tunables for
+        governors that support them.
+
+        :raises: TargetError if governor is not supported by the device, or if,
+                 for some reason, the governor could not be set.
+
+        """
+        supported = self.list_governors(device)
+        if governor not in supported:
+            raise TargetError('Governor {} not supported for device {}'.format(governor, device))
+        sysfile = '/sys/class/devfreq/{}/governor'.format(device)
+        self.target.write_value(sysfile, governor)
+
+    @memoized
+    def list_frequencies(self, device):
+        """
+        Returns a list of frequencies supported by the device or an empty list
+        if could not be found.
+        """
+        cmd = 'cat /sys/class/devfreq/{}/available_frequencies'.format(device)
+        output = self.target.execute(cmd)
+        available_frequencies = [int(freq) for freq in output.strip().split()]
+
+        return available_frequencies
+
+    def get_min_frequency(self, device):
+        """
+        Returns the min frequency currently set for the specified device.
+
+        Warning, this method does not check if the device is present or not. It
+        will try to read the minimum frequency and the following exception will
+        be raised ::
+
+        :raises: TargetError if for some reason the frequency could not be read.
+
+        """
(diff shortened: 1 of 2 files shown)


devlib/module/devfreq.py
```diff
@@ -0,0 +1,201 @@
...
+
+    name = 'devfreq'
+
+    @staticmethod
+    def probe(target):
+        path = '/sys/class/devfreq/'
+        return target.file_exists(path)
+
+    def __init__(self, target):
+        super(DevfreqModule, self).__init__(target)
```
setrofim: Can remove this, as only invoking the super here.

devlib/module/devfreq.py
```diff
@@ -0,0 +1,201 @@
...
+
+    @memoized
+    def list_frequencies(self, device):
+        """
+        Returns a list of frequencies supported by the device or an empty list
+        if could not be found.
+        """
+        cmd = 'cat /sys/class/devfreq/{}/available_frequencies'.format(device)
+        output = self.target.execute(cmd)
+        available_frequencies = map(int, output.strip().split())  # pylint: disable=E1103
```
setrofim: `from past.builtins import map` or change to a list comprehension in order to be compatible between Py2 and Py3 (see: http://python-future.org/compatible_idioms.html#map).

devlib/module/devfreq.py
```diff
@@ -0,0 +1,201 @@
...
+        available_frequencies = self.list_frequencies(device)
+        try:
+            value = int(frequency)
+            if exact and available_frequencies and value not in available_frequencies:
+                raise TargetError('Can\'t set {} frequency to {}\nmust be in {}'.format(device,
+                                                                                        value,
+                                                                                        available_frequencies))
+            sysfile = '/sys/class/devfreq/{}/max_freq'.format(device)
+            self.target.write_value(sysfile, value)
+        except ValueError:
```
setrofim: Move to just after the conversion attempt, and remove a level of indent from the remaining code.

mdigiorgio:
Updated the patch with requested changes.

devlib/module/devfreq.py
```diff
@@ -0,0 +1,199 @@
...
+            raise ValueError('Frequency must be an integer; got: "{}"'.format(frequency))
+
+        if exact and available_frequencies and value not in available_frequencies:
+            raise TargetError('Can\'t set {} frequency to {}\nmust be in {}'.format(device,
+                                                                                    value,
+                                                                                    available_frequencies))
+        sysfile = '/sys/class/devfreq/{}/max_freq'.format(device)
+        self.target.write_value(sysfile, value)
+
+    def set_governor_for_devices(self, devices, governor):
```
derkling: If you can have multiple devices, as for example we can have multiple CPUs for the cpufreq module, we found it quite useful to use an shutil script to reduce the number of host->target calls.

Could it make sense to add something like `cpufreq_{set,get}_all_frequencies`:
  https://github.com/ARM-software/devlib/blob/master/devlib/bin/scripts/shutils.in#L18
which has a Python API similar to:
  https://github.com/ARM-software/devlib/blob/master/devlib/module/cpufreq.py#L379
?

devlib/module/devfreq.py
```diff
@@ -0,0 +1,199 @@
...
+        raised ::
+
+        :raises: TargetError if for some reason the frequency could not be read.
+        """
+        sysfile = '/sys/class/devfreq/{}/max_freq'.format(device)
+        return self.target.read_int(sysfile)
+
+    def set_max_frequency(self, device, frequency, exact=True):
+        """
+        Set's the maximum value for device frequency. Actual frequency will
```
derkling: Maybe better:

```
Sets the maximum value for device frequency.

Actual frequency will...
```

?

devlib/module/devfreq.py
```diff
@@ -0,0 +1,199 @@
...
+        :raises: ValueError if ``frequency`` is not an integer.
+
+        """
+        available_frequencies = self.list_frequencies(device)
+        try:
+            value = int(frequency)
+        except ValueError:
+            raise ValueError('Frequency must be an integer; got: "{}"'.format(frequency))
+
+        if exact and available_frequencies and value not in available_frequencies:
```
derkling: Since `available_frequencies` is (eventually) an empty list, should not be possible to skip the second condition?

devlib/module/devfreq.py
```diff
@@ -0,0 +1,199 @@
...
+from devlib.exception import TargetError
+from devlib.utils.misc import memoized
+
+class DevfreqModule(Module):
+
+    name = 'devfreq'
+
+    @staticmethod
+    def probe(target):
+        path = '/sys/class/devfreq/'
```
derkling: Could be the folder empty?

Maybe we should check there is at least a subfolder.
Probably you could better use your `list_devices` to check that we get a non empty list of subfolders?

devlib/module/devfreq.py
```diff
@@ -0,0 +1,199 @@
...
+
+        :raises: TargetError if governor is not supported by the device, or if,
+                 for some reason, the governor could not be set.
+
+        """
+        supported = self.list_governors(device)
+        if governor not in supported:
+            raise TargetError('Governor {} not supported for device {}'.format(governor, device))
+        sysfile = '/sys/class/devfreq/{}/governor'.format(device)
+        self.target.write_value(sysfile, governor)
```
derkling: Maybe we should force `as_root`?

devlib/module/devfreq.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        available_frequencies = self.list_frequencies(device)
+        try:
+            value = int(frequency)
+            if exact and available_frequencies and value not in available_frequencies:
+                raise TargetError('Can\'t set {} frequency to {}\nmust be in {}'.format(device,
+                                                                                        value,
+                                                                                        available_frequencies))
+            sysfile = '/sys/class/devfreq/{}/min_freq'.format(device)
+            self.target.write_value(sysfile, value)
```
derkling: `as_root`?

devlib/module/devfreq.py
setrofim: I think devfreq devices tend to be a lot less homogeneous that CPUs, so it might not make sense to set a particular frequency for all at once; then again, I guess, there no harm in having that as an option...

devlib/module/devfreq.py
mdigiorgio: What do you mean? `probe()` is just checking whether the path exists.

setrofim: I think @derkling  is questioning whether it is enough that the directory exists, or whether `probe()` should check that is also non-empty (i.e. not only that `devfreq` support has been enabled in the kernel, but also that at least one policy is implemented).

We've recently run into a issue like this with `cpufreq`. See https://github.com/ARM-software/devlib/commit/0d63386343e56ff3f20c4a65f71bcd3f281b1ffa

derkling: Yes, right.

derkling: IOW: we should not enable the module if there is really nothing to possibly "control"...

devlib/module/devfreq.py
mdigiorgio: There is no `as_root` option for `write_value`

setrofim: `write_value` already forces `as_root=True`. One wonders whether it should be `as_root=self.is_rooted` or even be exposed as an argument; however, given that we have lived with it as-is up to this point, I'm reluctant to suggest the change until there is some evidence that it is needed.

In any case, that's unrelated to this pull.

mdigiorgio:
Thanks for the review, I've addressed your comments and updated the PR.

PR #295, opened 2018-06-28
Add IIO-based ACME instrument (BaylibreAcmeInstrument)

pietos01:
The commit message describes the reason for implementing another `Instrument` for the Baylibre ACME board.

This pull request has yet to be reviewed and modified before being accepted, as described in the following comments.

The main interrogation being about either

- keeping the "native order of magnitude" as reported by the ACME (mV, mA and mW) which is the range that makes the most sense knowing the range in which the board can be used but this would require adding more units to `_measurement_types` (in `Instrument/__init__.py`) which does not seem to be designed to handle many variants of the elementary units;

- converting the samples to V, A and W and potentially loosing in accuracy (most certainly negligible) while using a non-ideal scale; this is what is currently implemented.


devlib/instrument/baylibre_acme.py
```diff
@@ -0,0 +1,424 @@
...
+
+        self.probes = {p.nickname : p for p in probes}
+
+    # channel management
+
+    def list_channels(self):
+        # TODO: not sure what to do:
+        pass
+        #      or
+        # return {p: self.probes[p].list_channels() for p in self.probes}
```
pietos01: If we were to implement the output as a `dict`, it wouldn't comply with `Instrument`.

If we decided to output channels such as `'<probe>/<site>_<kind>'`, we can't prepend `p` to the channel name in this function (I believe) so that the channel itself should be able to output that string. A way would be to overwrite its `'<site >'` with `'<probe>/<site>'` but then `IIOINA226Instrument.reset` could not rely on the corresponding method of its parent ... Am I missing something?

This is an obvious drawback of my design choice of using an `Instrument` to hold `Instrument`s ...

devlib/instrument/baylibre_acme.py
```diff
@@ -0,0 +1,424 @@
...
+        args = {k : v for k, v in locals().iteritems() if k != 'self'}
+
+        def pseudo_list(v, i):
+            try:
+                return v[i]
+            except TypeError:
+                return v
+
+        for i, p in enumerate(self.probes.values()):
+            p.setup(**{k : pseudo_list(v, i) for k, v in args.iteritems()})
```
pietos01: Is this too "obfuscated"?

devlib/instrument/baylibre_acme.py
```diff
@@ -0,0 +1,424 @@
...
+    def get_data(self, outfile=None):
+        def compose_fname(basefile, probe_name):
+            if basefile:
+                base, ext = os.path.splitext(basefile)
+                return '{}_{}{}'.format(base, probe_name, ext)
+            return basefile
+
+        pairs = ((p, self.probes[p].get_data(compose_fname(outfile, p)))
+                 for p in self.active_probes)
+        return collections.OrderedDict(pairs)
```
pietos01: With multiple active probes, we are returning a dictionary of `MeasurementCsv`s which does not comply with the parent interface!

I thought about merging all the probes into a single CSV but:

- as probes can have different sampling rates, the CSV for a given probe may have more samples (*i.e.* more lines) than the CSV for another probe for a single run. A solution would be to prevent the user from having different sampling rates but isn't this a bit extreme?

- Each probe reports its own timestamps so that samples from different probes (even if they are taken at the same sampling frequency) are not taken at the same times which would mean that a CSV for multiple probes would contain multiple timestamp columns!

doc/instrumentation.rst
```diff
@@ -269,4 +269,468 @@ Available Instruments
...
+
+   acme.reset(probes='battery', sites=['timestamp', 'bus', 'shunt'])
+   acme.start()
+   time.sleep(10)
+   acme.stop()
+   data = acme.get_data()['battery']
+   acme.teardown()
+
+   # Compute numpy array containing the power channel with 64-bit FPU accuracy:
+   power = data['shunt_voltage'] * data['bus_voltage'] / shunt_resistor
```
pietos01: Could we imagine having a `high_resolution_power` option for `BaylibreAcmeInstrument` which would allow computing "offline" the power signal or is this too far from the data acquisition to be incorporated into an `Instrument`?

setrofim:
How is this different from the existing `acmecape` instrument?

https://github.com/ARM-software/devlib/blob/master/devlib/instrument/acmecape.py

doc/instrumentation.rst
```diff
@@ -269,4 +269,468 @@ Available Instruments
...
+The Infrastructure
+^^^^^^^^^^^^^^^^^^
+
+Retrieving measurement from the ACME through devlib requires:
+
+  - a BBB running the `image built for using the ACME`_ (micro SD card required);
+  - an ACME cape on top of the BBB;
+  - at least one ACME probe [#acme_probe_variants]_ connected to the ACME cape;
+  - a BBB-host interface (typically USB or Ethernet) [#acme_name_conflicts]_;
+  - a host (the one running devlib) with `libiio (the Linux IIO interface)`_ installed, and a Python environment able to find the libiio Python wrapper *i.e.* able to ``import iio`` as communications between the BBB and the host rely on the `Linux Industrial I/O Subsystem`_ (IIO).
```
derkling: Can you better detail how to get the python IIO support?
It seems that, on my Ubuntu 16.04, I cannot just pip install it...

derkling: Ok, the support seems to be provided by the python-libiio package... maybe we can add an example note for Ubuntu systems?

devlib/instrument/baylibre_acme.py
setrofim: I would question the need for the `IIOINA226Instrument` to be an actual `Instrument`. Each probe is one <site> exposing timestamp, voltage (I don't see a reason to expose both bus and shunt voltages to the outside world), current, and power. `IIOINA226Instrument` is basically an interface to a probe, but it does not need to implement the `Instrument` API, as that just makes things more complicated and confusing.

devlib/instrument/baylibre_acme.py
```diff
@@ -0,0 +1,424 @@
...
+        def pseudo_list(v, i):
+            try:
+                return v[i]
+            except TypeError:
+                return v
+
+        for i, p in enumerate(self.probes.values()):
+            p.setup(**{k : pseudo_list(v, i) for k, v in args.iteritems()})
+
+    def reset(self, probes=None, sites=None, kinds=None, channels=None):
```
setrofim: The breaks the API by adding `probes`; and, as mentioned above, probes == sites.

derkling:
Maybe not necessarily related to this code but I was actually hoping that the different approach proposed by this new module would have solved an issue I have sometimes also with the existing instrument. 

While testing the API with these commands:
```python
In [1]: import iio
In [2]: import devlib
In [3]: acme = devlib.BaylibreAcmeNetworkInstrument("192.168.0.1", probe_names=['ch0', 'ch1', 'ch2'])
In [4]: int_times = acme.integration_times_available
In [5]: ratios    = acme.oversampling_ratios_available
In [6]: acme.setup(shunt_resistor=10000, integration_time_bus=int_times[1], integration_time_shunt=int_times[1], oversampling_ratio=ratios[1])
In [7]: acme.reset()
In [8]: acme.start()
```

after a few seconds I get this error:

```python
In [9]: Exception in thread Thread-262:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/data/Code/lisa/libs/devlib/devlib/instrument/baylibre_acme.py", line 233, in collect_samples
    self.iio_buffer.refill()
  File "/usr/lib/python2.7/dist-packages/iio.py", line 474, in refill
    _buffer_refill(self._buffer)
  File "/usr/lib/python2.7/dist-packages/iio.py", line 32, in _checkNegative
    raise Exception("Error: " + _strerror(-result))
Exception: Error: Resource temporarily unavailable

Exception in thread Thread-264:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/data/Code/lisa/libs/devlib/devlib/instrument/baylibre_acme.py", line 233, in collect_samples
    self.iio_buffer.refill()
  File "/usr/lib/python2.7/dist-packages/iio.py", line 474, in refill
    _buffer_refill(self._buffer)
  File "/usr/lib/python2.7/dist-packages/iio.py", line 32, in _checkNegative
    raise Exception("Error: " + _strerror(-result))
Exception: Error: Resource temporarily unavailable
Exception in thread Thread-263:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/data/Code/lisa/libs/devlib/devlib/instrument/baylibre_acme.py", line 233, in collect_samples
    self.iio_buffer.refill()
  File "/usr/lib/python2.7/dist-packages/iio.py", line 474, in refill
    _buffer_refill(self._buffer)
  File "/usr/lib/python2.7/dist-packages/iio.py", line 32, in _checkNegative
    raise Exception("Error: " + _strerror(-result))
Exception: Error: Resource temporarily unavailable
```


Now, if I try to query the device using the iio-capture script, I get:

```bash
$ /usr/bin/iio-capture -n 192.168.0.1 -d 1 iio:device0
Unsupported write attribute 'in_oversampling_ratio'
```

Which is the same error I sometimes get also by using the current API... but it usually happens after some time (days) I've been using the CAPE. Usually, after a reset I get it working as expected... but with this new API it seems to happen almost always, right after a reboot and at the really first usage.

For completeness of information, these are the values I get from the code above:

```python
In [13]: ratios
Out[13]: (1, 4, 16, 64, 128, 256, 512, 1024)
In [14]: int_times
Out[14]: (0.00014, 0.000204, 0.000332, 0.000588, 0.0011, 0.002116, 0.004156, 0.008244)
```

I'm wondering if you even notice this behaviour and if you know how I can better debug the issue.

devlib/instrument/baylibre_acme.py
```diff
@@ -0,0 +1,424 @@
...
+import iio
+import numpy as np
+
+from devlib import Instrument, InstrumentChannel, CONTINUOUS, MeasurementsCsv
+
+
+class IIOInstrumentChannel(InstrumentChannel):
+
+    def __init__(self, iio_channel, site, measure, iio_dtype):
+        # NB: iio_channel.id could be used to figure out measure as, with
```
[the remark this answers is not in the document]
setrofim: We should go with A, V, and W. Code that processes power data should not know or care which instrument generated it. All data should be reported consistently.

I don't believe the loss precision is a concern in practice, especially given how noisy power measurements tend to be.  If I'm wrong, and it is an issue, then we *can* switch to reporting in milli's; but that would require updating every other energy instrument to do the same.

devlib/instrument/baylibre_acme.py
setrofim: Align the timestamps and re-sample to the highest rate the individual outputs before combining them. This ought to be relatively straight-forward to do with `pandas`.

doc/instrumentation.rst
pietos01: To install it, I downloaded the [archive provided](http://swdownloads.analog.com/cse/travis_builds/master_latest_libiio-trusty.tar.gz) which contains the Python files in `usr/lib/python2.7/site-packages` which I copied to the corresponding location and made sure that my path (in the Python environment) was accessing that directory. This is probably not the cleanest way and seemed a bit counter-intuitive.

If there is an APT package then it's much better! Also, I couldn't find an module on PyPI for libiio (which would allow the better option of using pip) ...

derkling:
Regarding the issue mentioned above, I've update my BBB image to the last release:

      **acme-beaglebone-black_b3-sdcard-image.xz**

which I assume it means "Beta3" release

Precisely, I'm now using this SDCard image:
   https://gitlab.com/baylibre-acme/ACME-Software-Release/blob/c0d29e638e9749b267dadd838fee8510f4a67687/README.md
and the board seems to be stable now. :+1: 

Perhaps we should add this tag as a minimum required version, just to ensure that people are running the right SW stack on the device side?

pietos01:
> @setrofim: How is this different from the existing [`acmecape`](https://github.com/ARM-software/devlib/blob/master/devlib/instrument/acmecape.py) instrument?

Please refer to the [commit message](https://github.com/pietos01/devlib/commit/2dc6f38161252ada3bab2ed8fa4ae38955335c24) for this PR. I can further detail the reason why we would want to interface our Python source directly with the drivers instead of escaping from Python to run a script but I believe that commit message covers the main arguments.

> @derkling: Regarding the issue mentioned above, I've update my BBB image to the last release [...] and the board seems to be stable now.

Good! Going one step further in "using the latest", I notice from the line numbers and the type of the exceptions raised (`Exception` instead of `OSError`) that the `iio.py` you are using is not the [latest version](https://github.com/analogdevicesinc/libiio/blob/master/bindings/python/iio.py) and I therefore wonder how old your Python wrapper is and what unexpected issues this could raise!

From a quick check, I can say that it is not more recent than [November 2016](https://github.com/analogdevicesinc/libiio/commit/6847e2233e20a81f842e446104f6added2451c61), when the move from `Exception` to `OSError` was made. It could therefore contain bugs or limited features that have already been fixed or enhanced ... as an example, it does not include [a commit seeming to imply a port to Python 3](https://github.com/analogdevicesinc/libiio/commit/8179c7a1335afc5b744a9d793bc391bed1e97630).

Furthermore, `apt` `show`s that `python-libiio` depends on the`libiio0` APT package and both are from version `0.6.40-1` which seems to be from around the first half of 2016 according to [their git tags](https://github.com/analogdevicesinc/libiio/tags?after=v0.8&before=v0.6).

Although I still wonder what the best and cleanest way to install the IIO Python wrapper is, it looks like the APT `python-libiio` package might not be it.

> Perhaps we should add this tag as a minimum required version, just to ensure that people are running the right SW stack on the device side?

Keep in mind that there are 4 "blocks" of software:
1. The SD card image used to run the BBB;
2. IIO running on the device (at boot, this is tied to the SD card image);
3. IIO running on the host (`libiio.so.0` on Linux corresponding to the `libiio0` APT package);
4. IIO Python wrapper (basically `iio.py`)

Let's ignore the possibility of manually upgrading 2. after booting 1. and of having a custom 1.: in his setup, one only has control over 1., 2. and 3.

I see that there exist [`iio.version`](https://github.com/analogdevicesinc/libiio/blob/6ecff5d46e1b12c2859f0b63a73282940e3402bb/bindings/python/iio.py#L360) and [`iio.Context.version`](https://github.com/analogdevicesinc/libiio/blob/6ecff5d46e1b12c2859f0b63a73282940e3402bb/bindings/python/iio.py#L804) that would respectively allow us to programmatically verify 3. and 2. and, if I'm not mistaken, 4. could be read through `iio.__version__` but we have no way to verify 1. through `iio` and in general, I can't think of a clean way to do so. However 1. seems to have been the origin of the behaviour you were experiencing. I imagine that there exists some way of figuring out the SD card "version" (or stamp) from accessing the file system (SSH) of the BBB but this seems overkill and I would give up on verifying the SW stack versions ourselves. IMHO, if a bug comes from IIO or the ACME image, I don't consider it our responsibility to mitigate or even detect it.

Anyhow, I agree that a comment in the documentation about keeping everything up to date on both sides would help and would be particularly helpful if we figured out the "proper way" of installing the **latest version** libIIO + the Python wrapper!

devlib/instrument/baylibre_acme.py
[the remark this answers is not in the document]
pietos01: > We should go with A, V, and W.

+1

> I don't believe the loss precision is a concern in practice, especially given how noisy power measurements tend to be.

You are right

setrofim:
>  IMHO, if a bug comes from IIO or the ACME image, I don't consider it our responsibility to mitigate or even detect it.

Disagree. It is precisely the purpose of `devlib` as an abstraction layer to deal with this sort of shit. Whilst not everything is under our control, we ought to do as much as we reasonably can to make it easier for our users.  Note: I don't necessarily disagree with most of your analysis above, just with the general attitude of of "it's not our job to deal with the issues in the dependencies we introduce". 

Documenting the issue is a good start, but let's face it -- most people don't read the docs, so if we can do  something in code to mitigate, or at least, to detect and point out the issue, we should.

>  I imagine that there exists some way of figuring out the SD card "version" (or stamp) from accessing the file system (SSH) of the BBB but this seems overkill and I would give up on verifying the SW stack versions ourselves

Well, it's just the case of `cat`'ing `/etc/os-release` on BBB, but given that we don't already require login creds for it, I agree that doing that might be overkill (unless we do a best-effort attempt using the default creds, and ignore any failures?). In any cases, do we not only about (1) due to its influence on (2), and as we can read (2) directly, we can just forget about it?

derkling:
Thanks @pietos01 for the exhaustive analysis, however I have to fix one of my previous statements.
The issue at the end happened again even with the most recent image for the BBB.

As you pointed out, I have a pretty old host side support, precisely:

```bash
python-libiio:  0.6.40-1
libiio-dev:     0.6.40-1
```

thus, maybe the issue is related to those libs... I can try to update them and run some more experiments.

If it turns out to fix my issue it means that we can get it right with just the fixes on point 3 and 4 you proposed. Let see...

pietos01:
> I [...] disagree with [...] the general attitude of "it's not our job to deal with the issues in the dependencies we introduce".

Re-reading what I wrote, I realise I phrased that very badly ... My concern was (and is) that I don't see how we could handle such bugs. The only things that I believe we could realistically do are:

- Catch exceptions that `iio` throws and rephrase them to make them clear to a `devlib` user (Actually, this is error handling for errors that are already expected by the code, not for bugs ...)

- Verify the version(s) that is (are) used by the user (and I expressed what I thought to be obstacles to this but your latest answer addresses these about (1); see below);

Putting version verification aside, how would you programmatically detect (not even talking about *handling*) a bug coming from `iio`?

> Documenting the issue is a good start

Will do! Saying to "use the latest", at least.

> most people don't read the docs

:disappointed: 

>  if we can do something in code to mitigate, or at least, to detect and point out the issue, we should.

So how should we do it (*cfr.* first point of this comment)?

About verifying the version(s) used, should we hard code versions IDs for (2), (3), (4) and, potentially, (1) and mandate the user to have these?

If yes:

- in what way? should we `raise`? `warnings.warn`? write to the `logger`? ...? Are we okay with keeping on running even if the versions are not what we want or do we halt execution?

- what happens when a new release of `iio` comes out or if Baylibre release a new BBB image? I guess we could implement an "at least" condition but then we would accept mismatched versions between (1), (2), (3) and (4) ... how to handle those?

- as a side note: in the case we refuse to proceed if the versions do not please our code, should we implement a "switch" allowing the user to turn off these checks as to avoid the frustration of *"YES! I know what I'm doing, stupid devlib! Let me do it!"*?

About doing anything more than checking the versions, I am out of ideas ...

>  it's just the case of `cat`'ing `/etc/os-release` on BBB

I did not know that, very cool; is there a standard that ensures that this file will always change between releases of the ACME SD card image? From the ones that have been released, it seems to be the case as `/etc/os-release` contains

```
ID="acme-baylibre"
NAME="Poky (Yocto Project Reference Distro)"
VERSION="2.1.X (krogoth)"
VERSION_ID="2.1.X"
PRETTY_NAME="Poky (Yocto Project Reference Distro) 2.1.X (krogoth)"
```
where `2.1.X` is `2.1.1`, `2.1.2` or `2.1.3` for, respectively, the releases `b1`, `b2` and `b3` of the image.

So, as the default (and only) user on the BBB image is `root` (no password), we could easily read `/etc/os-release` from a `ssh root@hostname`.

Although I must admit I didn't do it (yet!), a basic good security practice would be to, at least, set a password for the `root` user. In that case, should we prompt the `devlib` user for that password? Should we make it a parameter to the constructor? If yes, I believe we should make it optional as one would not necessarily be comfortable inputting such information ... But there is always the "best-effort" option you mention.

> In any cases, do we not only care about (1) due to its influence on (2), and as we can read (2) directly, we can just forget about it?

I don't believe so.

Running the following:

``` python
import devlib
acme = devlib.BaylibreAcmeNetworkInstrument(<...>)
print(acme.iio_context.version)
```
gives `(0L, 6L, u'v0.6   ')` for, at least `b3` and `b2`.

Furthermore, the [changelogs of the releases](https://github.com/baylibre-acme/ACME/releases) only ever mention fixes that are not directly related to IIO itself. So a bug could be fixed by updating (1) without modifying (2)-(4).

BTW, `iio.__version__` is not set so can't be used to check (4) and `iio.version` is reported as `(0L, 14L, u'8ac1893')` for my version which has a per-commit resolution -> how do we completely compare 2 versions as `<` won't have the expected behaviour or do we only care about the release version numbers?

**In summary** about error handling in `BaylibreAcmeInstrument`:

- exceptions raised by "modern" (2 years ago, when they moved from `Exception` to `OSError`, they effectively broke their API; by modern, I mean what the latest code does) `iio` can be handled [and are](https://github.com/pietos01/devlib/blob/2dc6f38161252ada3bab2ed8fa4ae38955335c24/devlib/instrument/baylibre_acme.py#L232); please come back to me if you come across any extra I overlooked;

- we know we are able to verify the version of (2) and (3) through `iio` and (1) through `ssh`; but

    * how do we handle version mismatch?

    * how do we handle failed `ssh` connections?

    * do we give up on (4)?

- what could we do other than version checking to "detect bugs in `iio`"?

pietos01:
> maybe the issue is related to those libs...

Maybe ... maybe not: When I was experimenting with the ACME, I came across a *Device is busy* exception (can't remember the `errno` but it was an `OSError`) and I tried to figure out what was causing it. From what I speculated, it looked something somewhere in the stack didn't like me trying to access the board when an `iio.Buffer` was created (you can see that that thing creates some activity because when it is instantiated it in Python, a LED lights up on the BBB and stays lighten up until you `del` it). You can play a bit with `iio` and try this out for yourself: it is very reproducible.

By using the `setup`-`reset`-`start`-`stop`-`teardown` sequence from `devlib`, I try to enforce a flow that would not trigger this but I am worried that the issue is not limited to IIO buffers and that they are simply the best way to trigger a given bug (related to shared resource access, maybe?) that could also be triggered in others ways *e.g.* bad timing, ... Resulting in "random" bugs ... to be honest, I hope it is not the case and I must say that the frequency at which I came across these exceptions has dropped since using `BaylibreAcmeInstrument` but I admit I didn't use it that much as a user, yet.

I [see you experienced the issue by directly writing to the files](https://github.com/BayLibre/iio-capture/issues/3) as well so it could come from deep into the stack ... I am very eager to learn what causes this as we could then avoid triggering it from `devlib`!

Anyway, a reboot always seems to solve the issue (from my experience, at least) so until it's fixed, could we maybe imagine a mitigation approach where we could allow `devlib` to `ssh root@acme reboot`?

setrofim:
> About verifying the version(s) used, should we hard code versions IDs for (2), (3), (4) and, potentially, (1) and mandate the user to have these?

We should mandate the minimum, rather than specific, versions. Mandating specific version is too re-restrictive, and something we'll have to keep updating in the code to remain current. Here we're not trying to make sure that the user's setup is one that we definitely know to work; here, we're trying to make sure that the user's setup is not one we definitely know to be broken.

> in what way? should we raise? warnings.warn? write to the logger? ...? Are we okay with keeping on running even if the versions are not what we want or do we halt execution?

Raise a `HostError`   or a `TargetError`  (depending on where the version is detected -- BBB is a bit of a grey area, but I would define it as being part of the host rather than the target, as it is linked to a particular setup) stating that the discovered version is too old and should be updated.

> Catch exceptions that iio throws and rephrase them to make them clear to a devlib user (Actually, this is error handling for errors that are already expected by the code, not for bugs ...)

Well, here it's a known bug in an external project, so from `devlib` point of view, this is an expected error. Basically anything that' is not a bug in `devlib` itself or in the code using it, should result in a `DevlibError`.

> how do we handle version mismatch?

We just need to try to make sure each version is sufficiently recent.

> how do we handle failed ssh connections?

Log at debug level that version on BBB could not be verified because an ssh connection connection could not be established, and otherwise ignore -- this is best-effort.

>  In that case, should we prompt the devlib user for that password? Should we make it a parameter to the constructor?

Prompting the user is definitely out of the question -- `devlib` may be used in non-interactive environments, and should not assume a user is present when the code runs. An optional parameter is a possibility, but might be overkill.

>  how do we completely compare 2 versions as < won't have the expected behaviour or do we only care about the release version numbers?

Just check the major and minor parts and ignore the commit.

> what could we do other than version checking to "detect bugs in iio"?

Nothing. I wasn't suggesting we pro-actively try to detect bugs in `iio`; just that we should, where possible, implement detection/mitigation for bugs we are already aware of.

pietos01:
I dropped the idea of inheriting from `Instrument` at the level of the probes and of the ACME and relying on the inherited `reset` combined with INA226 channels inheriting from `InstrumentChannel` to enable channels. Instead, only the ACME inherits from `Instrument` and uses `Instrument.reset` to enable which abstract the actual underlying IIO channels. Therefore, from the point of view of a `devlib` user, a `BaylibreAcmeInstrument` has 3 channels per probe (V, I and P) and a shared `timestamp` channel. All the logic and remapping to the actual underlying is then done in the rest of the implementation of `BaylibreAcmeInstrument.reset`. The concept of probe was removed from the API and probes are presented to the user as `site`s (except in the constructor, which still takes a `probe_names` which I kept as the board of the ACME says "Probe").

I added version verification through IIO and SSH. For the SSH one (checking `/etc/os-release` for the board version), I used `subprocess` to launch `ssh`: would it make more sense to use a `devlib.Target` instead (not tested)? If yes, should this be the `target` passed to `Instrument.__init__`? I also didn't bother implementing board version control for `BaylibreAcmeLocalInstrument` and `BaylibreAcmeXMLInstrument`. As discussed above, minimal versions have been hardcoded in the class: I used the ones I have on my machine.

The resampling between probes is done, as suggested, and a single `DataFrame` is obtained by the instrument with the desired channels as columns and a universal timestamp index. I took the opportunity of having a `DataFrame` to use `to_csv` and to add optional keyword parameters to `get_data` to be passed on to `to_csv`.

I also implemented my idea of having a "high resolution" mode where enabling the "power" channel from devlib actually enables the "shunt" and "bus" probes and does the multiplication under the hood with the accuracy of a FPU instead of the fixed-point from the INA226. This can be disabled through `BaylibreAcmeInstrument.setup(high_resolution=False)`.

The "constants" in the classes were also changed to capital names as mentioned earlier in this discussion.

Some exceptions now use the `devlib` `HostError` and `TargetError` classes.

This PR should not be accepted yet as I still have to update the documentation (and I'm waiting for the code to be fixed).

pietos01:
I noticed that importing the content of `baylibre_acme.py` into the namespace `devlib` requires from the user to have `iio` installed even if they don't use an acme as `pip install`ing `devlib` makes `import devlib` raise an exception complaining  about `iio` not found (coming from the `import iio` in `baylibre_acme.py`). How would we fix this? It does not make sense to require from anyone using `devlib` (including as a dependency) to have `iio` installed and, as `libiio` can't be found on PyPI, there is no way to automatically install it as a dependency from `pip`. The only solution I see requires removing the `import` of the ACME instruments in the `devlib` namespace which would make this instrument inconsistent with the other ones when being imported. Is there another way?

Secondly, for installing on Ubuntu, I think that the cleanest way is to manually download the latest `.deb` package for `libiio`, `apt install` it and, in case of a "non-standard" distribution of Python (*e.g.* `conda`), manually copy (from `/usr/lib/python2.7/site-packages/`) the Python wrapper to where it belongs.

marcbonnici:
> I used subprocess to launch ssh: would it make more sense to use a devlib.Target instead (not tested)? If yes, should this be the target passed to Instrument.__init__?

I think creating a separate devlib target to the BBB is a bit over kill and the target passed into `Instrument.__init__` should be the device under test rather than be re-purposed for the measurement device. I would suggest that instead of using subprocess it would make more sense to import Devlib's `SshConnection` and utilise that to perform the query or to use `pxssh` directly to handle the connection. 

> I noticed that importing the content of baylibre_acme.py into the namespace devlib requires from the user to have iio installed even if they don't use an acme as pip installing devlib makes import devlib raise an exception complaining about iio not found (coming from the import iio in baylibre_acme.py). How would we fix this?

The imports should be guarded and then should be checked during initialisation raising an error message if appropriate. For an example of this please see the imports of `daqpower` in https://github.com/ARM-software/devlib/blob/master/devlib/instrument/daq.py

pietos01:
> I would suggest that instead of using `subprocess` it would make more sense to import Devlib's `SshConnection` and utilise that to perform the query

Done!

> The imports should be guarded and then should be checked during initialisation raising an error message if appropriate.

Excellent idea! Done!

marcbonnici:
> The imports should be guarded and then should be checked during initialisation raising an error message if appropriate.

Apologies, I meant that the guard should be placed around of `iio` inside of baylibre_acme.py rather than  the instrument itself in `__init__.py`. This will allow for the instrument to still be available for importing from devlib even if the user doesn't have the required module installed, and only if it is attempted to actually be used the error message prompting the user to install the missing module can be raised.

pietos01:
> I meant that the guard should be placed around of `iio` inside of `baylibre_acme.py` rather than the instrument itself in `__init__.py`.

Oh! Done!

setup.py
```diff
@@ -92,6 +92,8 @@
         'pyserial',  # Serial port interface
         'wrapt',  # Basic for construction of decorator functions
         'future', # Python 2-3 compatibility
+        'pandas',
+        'numpy',
```
pietos01: I added these as dependencies as they are used by `BaylibreAcmeInstrument`; is this fine? Should they be in `extras_require`?

marcbonnici: Good question, I think for now they should probably go into ``extras_require`` and guarded against as well.

pietos01: Although I'm not sure where this code is being used, I see that `devlib/derived/fps.py` is `import`ing `pandas` so it seems like the dependency already existed in the code but wasn't expressed in `setup.py`?

I am a bit confused: from [the documentation](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) (I've never used this particular parameter to `setup` myself), if they are in `extras_require`, we have to define [*entry points*](http://setuptools.readthedocs.io/en/latest/setuptools.html#entry-points) which become the keys to the `dict` passed as the `extras_require` argument and then "trigger" these points in the source (using `pkg_resources`? I'm not so sure ...), right?

But I don't see such calls in the codebase for the existing dependencies (`'daq'`, `'doc'` and `'monsoon'`) so I'm not really sure how I should implement this using the "proper" tools ...

devlib/instrument/baylibre_acme.py
```diff
@@ -0,0 +1,547 @@
...
+            self.ssh_connection = None
+
+    def check_version(self):
+        super(BaylibreAcmeNetworkInstrument, self).check_version()
+
+        cmd = r"""sed -nr 's/^VERSION_ID="(.+)"$/\1/p' < /etc/os-release"""
+        try:
+            ver_str = self._ssh(cmd).rstrip()
+            ver = tuple(map(int, ver_str.split('.')))
+        except:
```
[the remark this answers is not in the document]
pietos01: The content of the `try` block gets information from the BBB and the idea is to log "if anything went wrong"; we expect the SSH connection to fail in some cases and it could fail for various reasons at various levels. Even if the connection fails, the code on the host should keep executing (best effort).

What I could add there is a capture of the `Exception` (which is still as general as the lonely `except`) and log it's message. Is this fine?

pietos01:
I fixed the code and documentation based on the feedback with the exception of the `except`.

devlib/instrument/baylibre_acme.py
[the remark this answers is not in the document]
marcbonnici: In that case yea I think catching `Exception` and logging it's error messages would be better in this situation. That way it might give some insight as to what went wrong and it would also not ignore things like `KeyboardInterrupt`s etc.

devlib/instrument/baylibre_acme.py
```diff
@@ -0,0 +1,557 @@
...
+    # initialization and teardown
+
+    def setup(self, shunt_resistor,
+              integration_time_bus,
+              integration_time_shunt,
+              oversampling_ratio,
+              buffer_samples_count=None,
+              buffer_is_circular=False,
+              absolute_timestamps=False,
+              high_resolution=True):
```
setrofim: I maybe misunderstanding the intent here, but it seems this code should be part of `__init__` (or maybe `reset`?) instead. `setup` is not guaranteed to be called on every use of the instrument -- it's intended for "one-time setup" actions, such as installing executable on the target, and so typically would not be modifying the state of `self`, unless overriding something that would otherwise be set earlier on the assumption that such one-time setup has already been performed (e.g. setting path to the installed executable inside `__init__`).

pietos01: I must admit I struggled to partition the different steps into `__init__`, `setup` and `reset`.

IIO uses a buffer to transfer data between the ACME-BBB and the host. Although required for a transfer to take place, this buffer is created and deleted independently of the IIO context and the establishment of a connection. The properties of the transfer itself are dictated by the properties of the buffer.

Here is my reasoning:
- A constructor (`__init__`) sets an instance to a valid initial state and is obviously not intended to be called more than once on an instance. Therefore, it makes sense to have `__init__` creating the IIO context (and the underlying connection to the BBB) as it will only happen once. On the other hand, because a succession of buffers with varying parameters can be used with the same IIO context, I didn't think creating the buffer in `__init__` made sense. Indeed, in IIO, there is no need for taking the IIO context down and up again just to change the buffer parameters (which we would effectively have to do if the code was in `__init__`);
- [`reset`](https://devlib.readthedocs.io/en/latest/instrumentation.html#Instrument.reset) has a well defined API preventing me to add the extra parameters required by the creation of an IIO buffer;
- The [documentation for `Instrument.setup`](https://devlib.readthedocs.io/en/latest/instrumentation.html#Instrument.setup) states that
    > Parameters this method takes are **particular to subclasses (see documentation for specific instruments below)**. **What actions are performed by this method are also instrument-specific**. Usually these will be things like installing executables, **starting services**, deploying assets, etc. Typically, this method needs to be invoked at most once per reboot of the target (**unless `teardown()` has been called**), **but see documentation for the instrument you’re interested in**.

    which lead me to encapsulate the IIO buffer creation in `setup` (actually, in the current implementation only the IIO buffer setup happens in `setup`, the creation itself is done "inside" `start` so that the buffer only exists when necessary) in, what I thought to be, a way complying with the standard `Instrument`.

The way the flow works is:

1. an IIO context has to be created and its IIO channels are discovered - `__init__`
1. an IIO buffer with a given set of characteristics has to be created - `setup`
1. the IIO channels have to be enabled - `reset` [IIO is now ready to capture]
1. the capture of the selected channels takes place in the given context through the given buffer - `start`-`stop`
1. the IIO buffer is `free`d - `teardown` (actually `stop`)
1. IIO channels are potentially disabled - `teardown`
1. loop to 2. if desired;
1. the IIO context is deleted

And I tried to wrap this IIO flow in the methods `Instrument` provides which, I believed, seemed to work nicely together. Hence the little "calling graph" I joined to the documentation.

setrofim: > I must admit I struggled to partition the different steps into `__init__`, `setup` and `reset`.

The distinction is between when/how often things need to happen:

Some things are fairly persistent - need to only be done once for a particular device (e.g. installing an executable) or a single boot of the device (e.g. starting a daemon/service,  setting up a persistent communications channel, etc). This is what `setup` is intended for. In particular, it may not be called every time an `Instrument` is instantiated. This so that you don't have to keep doing these things, e.g. every time you want to re-run a notebook.

It is expected that a single interactive session (a WA run, notebook/script execution, etc) would have an instance of an `Instrument` persist throughout the session (rather than needed to re-instantiate it each time measurements need to be collected). So anything that needs to be done to configure the `Instrument` for interaction with a particular target should be done inside '__init__'. 

Finally, `reset` is used to prepare the `Instrument` for collecting a particular set(s) of  measurements. This typically just involves specifying which channels (for instruments that support multiple channels) should be collected.

To summarise:

- `setup` is called once per target setup/configuration, and may not be called every time an `Instrument` is created.
- `__init__` is called once per "interactive session" with the target.
- `reset` is called prior to collecting a particular set of measurements.

`setup` was intended to avoid repeating unnecessary steps in situations where target state can be assumed to persist across code invocations. Current applications though, do not make such assumptions, so right now,`setup` will get invoked once per instance of an `Instrument`. I think, for now, current implementation might be OK.

PR #298, opened 2018-07-05
utils/misc: Ensure outputs are converted to strings

marcbonnici:
If the process was killed, either the output or error can
be `None` which causes an error when attempting to join the outputs.


setrofim:
`str(None)` is `"None"`. Rather than that, I would do `output or ''` to get empty strings.

marcbonnici:
Fixed and updated existing error message aswell.

Issue #308, opened 2018-08-13
Leaked ssh connections


qais-yousef:
After running several tests on my target using LISA I suddenly started encountering ssh related problems.

I had 2 mode of failres: failed to scp and failed to connect altogether.

ssh'ing directly into the target from the terminal (ie: outside LISA test environment) worked.

After a bit of digging it seemed that there were many ssh connection looming around that cause the target to reach a limit after which it drops further ssh connection requests.

I am not entirely sure how everything works - but I believe devlib handles the connection steps with the target? If not please let me know and I'll file a bug against the correct module.

NOTE: I might have ctrl+c to interrupt some tests. Not sure if this contributed to the issue or just continuously running tests on a target without rebooting will cause this to happen eventually.

marcbonnici:
Hi,
Thanks for reporting this, I have submitted a PR (https://github.com/ARM-software/devlib/pull/310)  to ensure that the connections are closed automatically upon exit. 
Would you be be able to let me know if this fixes the issue you've been having?

qais-yousef:
Thanks for the fast response Marc. I'll test this and let you know.

qais-yousef:
Hi Marc

With the proposed fix I can no longer see ssh sessions leaking. I tested it with a simple `ps aux | grep ssh` where you could see a left over ssh session. And by running overnight test which didn't encounter the dropped connection problem.

I don't know what's the protocol - I can close this issue or feel free to close it once your PR is merged.

Cheers.

marcbonnici:
Thank you very much for testing that out for me.

That's fine I can close the issue when we merge the PR.

qais-yousef:
Actually just after I wrote the last comment - the test got stuck and had to ctrl+c and got this backtrace - something went wrong at atexit and the connection is still open.

```
^CError in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/usr/lib/python2.7/atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "/home/qais/src/lisa/libs/devlib/devlib/utils/ssh.py", line 237, in close
    self.conn.logout()
  File "/usr/local/lib/python2.7/dist-packages/pexpect/pxssh.py", line 443, in logout
    index = self.expect([EOF, "(?i)there are stopped jobs"])
  File "/usr/local/lib/python2.7/dist-packages/pexpect/spawnbase.py", line 341, in expect
    timeout, searchwindowsize, async_)
[...]

PR #310, opened 2018-08-14
ssh: Use atexit to automatically close ssh connections

marcbonnici:
As stated in https://github.com/ARM-software/devlib/issues/308 devlib can leave ssh connections open after they are 
no longer in use, so now register the close method with atexit so the 
connections are no longer left open upon exit.

### DIFF
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -237,7 +237,11 @@ class SshConnection(object):
 
     def close(self):
         logger.debug('Logging out {}@{}'.format(self.username, self.host))
-        self.conn.logout()
+        try:
+            self.conn.logout()
+        except:
+            logger.debug('Connection lost.')
+            self.conn.close(force=True)
 
     def cancel_running_command(self):
         # simulate impatiently hitting ^C until command prompt appears
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -25,6 +25,7 @@ import shutil
 import socket
 import sys
 import time
+import atexit
 
 # pylint: disable=import-error,wrong-import-position,ungrouped-imports,wrong-import-order
 import pexpect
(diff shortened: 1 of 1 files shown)


devlib/utils/ssh.py
```diff
@@ -180,6 +181,7 @@ def __init__(self,
         logger.debug('Logging in {}@{}'.format(username, host))
         timeout = timeout if timeout is not None else self.default_timeout
         self.conn = ssh_get_shell(host, username, password, self.keyfile, port, timeout, False, None)
+        atexit.register(self.close)
```
setrofim: Should be implemented inside `__del__` instead, i.e. close the connection when the object is destroyed, not when the interpreter exists.

devlib/utils/ssh.py
```diff
@@ -235,7 +237,11 @@ def background(self, command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, as
 
     def close(self):
         logger.debug('Logging out {}@{}'.format(self.username, self.host))
-        self.conn.logout()
+        try:
+            self.conn.logout()
+        except:
+            logger.warn('Connection lost.')
```
setrofim: `logger.warning` (`warn` was deprecated, IIRC). Also, I would probably change this to `debug` instead --  a warning should be used to make the user aware of something that might require a remedial action. In this case, since we're closing the connection anyway, it doesn't really matter.

devlib/utils/ssh.py
marcbonnici: When implementing with `__del__` the method was not called unless the program exited cleanly and still appeared to leave the the connections open.

devlib/utils/ssh.py
marcbonnici: Updated.

devlib/utils/ssh.py
setrofim: OK, I was wrong on this. The docs for `__del__` state that it's not guaranteed to be called for objects that exist on interpreter exit: https://docs.python.org/3/reference/datamodel.html#object.__del__

I withdraw my previous comment.

PR #311, opened 2018-08-16
target: ignore extra lines in read_tree_values

smuckle:
If a file has multiple lines the read_tree_values fn currently
returns the last line of output as the value for the file. Files
may have additional newlines in them however, so return the first
line instead in the hopes of getting the correct output.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -660,7 +660,8 @@ class Target(object):
             if ':' not in entry:
                 continue
             path, value = entry.strip().split(':', 1)
-            result[path] = value
+            if path not in result:
+                result[path] = value
         return result
 
     def read_tree_values(self, path, depth=1, dictcls=dict, check_exit_code=True):


derkling:
Hi Steve, thanks for this contribution!

To better explain the problem, maybe it can be useful for @setrofim if you provide a small example of the problematic output you got. This can also help Sergei (if possible) to suggest a more generic solution.

valschneider:
I don't think I hit that issue when I used `read_tree_values` in the [sched module](https://github.com/ARM-software/devlib/blob/master/devlib/module/sched.py), but from my use of it I would expect the entire content of the file to be saved in the output dictionnary - the caller can then use that data however it wishes.

We could store the whole file contents into `result[path]` (newlines and all), or turn that into a list and just append every line in there.

smuckle:
This was an issue trying to run the EAS generic tests on pixel 2 because the msm idle driver returns an extra newline at the end of the output from /sys/devices/system/cpu/cpu*/cpuidle/state*/name . This causes the cpuidle state name to be empty (read_tree_values picks the second, empty line), and since those states get stored in an OrderedDict with the name as the key, they all get collapsed into one. LISA then parses a trace with idle states during the test and aborts when it sees more than the one idle state it knows about.

Since the current semantics of read_tree_values are to return single lines of output for each node in the tree, I'd argue extending it and its clients to process multiple lines of output per node is an orthogonal change, and could be made separately :) .

valschneider:
Right, disregarding possible improvements it does make much more sense (IMHO) to fetch the first line rather than the last one.

setrofim:
I'm not convinced this is the right solution. The original code assumes that keys in `result` are unique (being paths in the file system). This is clearly clearly not the case for your output, but he underlying assumption that a unique path corresponds to a unique file (i.e. no two files share a path) still seems sound. Duplicates in this case are clearly unexpected and handling them by simply taking one value (be it the first or the last) and ignoring the rest, is just glossing over the underlying problem.

Maybe I'm misunderstanding the issue. Please could you provide a sample output of running the `read_tree_values` shell function on a path that exhibits this problem?

setrofim:
I've submitted a pull with an alternative fix for this issue: https://github.com/ARM-software/devlib/pull/314. Please could you try that one and let me know if it works for you.

PR #316, opened 2018-08-24
module/cpufreq: Add a contextmanager for temporary governor changes

valschneider:
We may sometime want to temporarily use another governor, and then
restore whatever governor was used previously - for instance, RT-app
calibration ([1]) needs to use the performance governor, but we don't
want this to interfere with e.g. our current experiment.

[1]: https://github.com/ARM-software/lisa/blob/master/libs/wlgen/wlgen/rta.py#L118

### DIFF
--- a/devlib/module/cpufreq.py
+++ b/devlib/module/cpufreq.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from contextlib import contextmanager
+
 from devlib.module import Module
 from devlib.exception import TargetStableError
 from devlib.utils.misc import memoized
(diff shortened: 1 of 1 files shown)


devlib/module/cpufreq.py
```diff
@@ -95,6 +97,37 @@ def set_governor(self, cpu, governor, **kwargs):
...
+        # in the same clock domain, so only manipulating one cpu per domain
+        # is enough
+        domains = set(self.get_affected_cpus(cpu)[0] for cpu in cpus)
+        prev_governors = {cpu : (self.get_governor(cpu), self.get_governor_tunables(cpu))
+                          for cpu in domains}
+
+        for cpu in domains:
+            self.set_governor(cpu, governor, **kwargs)
+
+        yield
```
setrofim: Should surround this in a  try/finally block to ensure the governor is restored (or at least an attempt is made to restore it) even if an exception occurs during context execution.

valschneider: For some reason I wrote this thinking that was done implicitly with context managers... Let me fix my code & my thoughts

valschneider:
Added a try/finally around the yield.

Issue #317, opened 2018-08-29
target: working_directory issue(s)


v0lker:
hi,

i stumbled upon this when i was trying to connect to a linux machine with a user that has an unusual home directory path (`/foo/user/`). not actually a problem as it is easy enough to work around, but it seems wrong:

`LinuxTarget::_resolve_paths()` will set the working directory to `/home/user/devlib-target` and then `Target::connect()` will try to `mkdir -p` that directory.

1. if /home/user doesn't exist, this will try to create it, but it won't succeed for ordinary users, as normally, /home is not writeable for just anybody

this could be solved by adding `as_root=True` to the mkdir invocation. but then, the directory would be owned by root and not any good for the user, so it needs additional fixups (changing ownership of the  directories involved).

2. i think, it should use the users home - which could potentially be retrieved by executing `pwd` or awking the home out of /etc/passwd - and not try to recursively create the devlib-target sub directory.

i'm happy enough to provide a PR.

setrofim:
You can override working directory on target creation:

    target = LinuxTarget(working_directory='/foo/user/devlib-working')

However, extracting it from `/etc/passwd` is not a bad idea; please feel free to submit the PR.

marcbonnici:
Resolved with https://github.com/ARM-software/devlib/pull/319.

PR #319, opened 2018-09-07
target.py: cope with non-root users and with non-standard home directories

v0lker:
proposed fix for https://github.com/ARM-software/devlib/issues/317

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -948,10 +948,7 @@ class LinuxTarget(Target):
 
     def _resolve_paths(self):
         if self.working_directory is None:
-            if self.connected_as_root:
-                self.working_directory = '/root/devlib-target'
-            else:
-                self.working_directory = '/home/{}/devlib-target'.format(self.user)
+            self.working_directory = self.path.join(self.execute("pwd").strip(), 'devlib-target')
         self._file_transfer_cache = self.path.join(self.working_directory, '.file-cache')
         if self.executables_directory is None:
             self.executables_directory = self.path.join(self.working_directory, 'bin')


devlib/target.py
```diff
@@ -244,8 +244,8 @@ def connect(self, timeout=None, check_boot_completed=True):
         if check_boot_completed:
             self.wait_boot_complete(timeout)
         self._resolve_paths()
-        self.execute('mkdir -p {}'.format(self.working_directory))
-        self.execute('mkdir -p {}'.format(self.executables_directory))
+        self.execute('mkdir -p {0} && chown {1} {0}'.format(self.working_directory, self.user), as_root=True)
```
setrofim: This breaks for unrooted devices. Leave the old method -- it's up to the user to specify a working directory location that is writable by `self.user`.

devlib/target.py
```diff
@@ -945,10 +945,7 @@ def capture_screen(self, filepath):
 
     def _resolve_paths(self):
         if self.working_directory is None:
-            if self.connected_as_root:
-                self.working_directory = '/root/devlib-target'
-            else:
-                self.working_directory = '/home/{}/devlib-target'.format(self.user)
+            self.working_directory = self.execute("awk -F: '/{}/ {{ print $6 }}' /etc/passwd".format(self.user)).strip()
```
setrofim: `awk` must be invoked via `self.busybox`.

devlib/target.py
v0lker: ok

devlib/target.py
v0lker: i do think we're ok relying on awk being present and we can't use busybox because it hasn't been installed yet.

i think there is a more straightforward solution, though: how about using the output of ``` `pwd` ``` or  ```echo $PWD``` ? the really nice thing about this is that it should fall back gracefully, even if someone didn't create ~user or authentication doesn't actually work via /etc/passwd.

will push the change accordingly, please let me know what you think.

devlib/target.py
```diff
@@ -945,10 +945,7 @@ def capture_screen(self, filepath):
 
     def _resolve_paths(self):
         if self.working_directory is None:
-            if self.connected_as_root:
-                self.working_directory = '/root/devlib-target'
-            else:
-                self.working_directory = '/home/{}/devlib-target'.format(self.user)
+            self.working_directory = self.execute("pwd".format(self.user)).strip()
```
marcbonnici: `.format(self.user)` isn't required here.

v0lker: oops. leftover from previous version. sorry about that.

marcbonnici:
Would it make sense to keep the sub-directory `devlib-target` inside of the home directory?

v0lker:
> Would it make sense to keep the sub-directory devlib-target inside of the home directory?

possibly - but i don't see how that has to do with this PR? from a quick glance, it seems, `{/tmp/,<external_storage>,</mnt/stateful_partition>}/devlib-target` is used as working_directory, if the home is not set.

am i misunderstanding what you're saying?

marcbonnici:
Sorry, I just meant that the previous behaviour was if the working directory was not explicitly set, it would use the `devlib-target` inside of the default home directory, however now the base home directory would be used, meaning anything that devlib deploys would be mixed along with any other files in the home directory rather then keeping them contained in a subdirectory.

So I was suggesting maybe doing something like:
`self.working_directory = self.path.join(self.execute("pwd").strip() , 'devlib-target')`

v0lker:
> So I was suggesting maybe doing something like:
> `self.working_directory = self.path.join(self.execute("pwd").strip() , 'devlib-target')`

i think you're absolutely right. sorry.

PR #320, opened 2018-09-20
Misc Fixes

marcbonnici:


devlib/host.py
```diff
@@ -55,7 +56,10 @@ def pull(self, source, dest, timeout=None, as_root=False): # pylint: disable=unu
             for each_source in iglob(source):
                 shutil.copy(each_source, dest)
         else:
-            shutil.copy(source, dest)
+            if os.path.isdir(source):
+                copy_tree(source, dest)
```
setrofim: Why not use `shutil.copytree` here?

devlib/__init__.py
```diff
@@ -56,7 +56,7 @@
 from devlib.utils.version import get_commit as __get_commit
 
 
-__version__ = '1.0.0'
+__version__ = '1.0.dev1'
```
setrofim: Please expand commit message to state the reason for this change.

marcbonnici: Updated.

devlib/host.py
marcbonnici: `shutil.copytree` does not support copying into an existing directory structure whereas the implementation from `distutils` does, therefore to try and mimic the behaviour of other targets as closely as possible I opted for this implementation.

setrofim: OK that makes sense. It's worth leaving a comment to that effect in the code.

marcbonnici: Updated.

PR #322, opened 2018-09-20
utils/version: Add a fallback encoding

valschneider:
In some environments (e.g. nosetest), sys.stdout.encoding can be None.
Add a fallback to handle these cases.


valschneider:
Hm that may have been a hasty PR, I'm also hitting the issue elsewhere. It's only happening with nosetests, investigation ongoing...

PR #324, opened 2018-09-21
ftrace: use correct trace-cmd binary in report function

joshuous:
target_binary should be used instead of host_binary

Signed-off-by: Joshua Choo <code@joshchoo.com>


joshuous:
Okay noted, thank you for pointing out my misunderstanding!

PR #330, opened 2018-10-29
target: Fix number_of_cpus() on targets with a coloured prompt

qperret:
On some Linux targets, the 'ls' command has a coloured output by
default. Since number_of_cpus() performs a regex on the output of that
command, the color codes around the name of files and directories can
cause issues. target.execute('ls /sys/devices/system/cpu') on my juno
will provide '\x1b[01;34mcpu1\x1b[0m' instead of 'cpu1', for example.
Because of this, number_of_cpus() can return 0.

Fix this with by forcing '--color=never' in ls. While at it, make sure
to raise a human-readable exception in case number_of_cpus() is null,
since this is always wrong, and can lead to pretty nasty and hard-to-debug
cases.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -140,10 +140,13 @@ class Target(object):
     def number_of_cpus(self):
         num_cpus = 0
         corere = re.compile(r'^\s*cpu\d+\s*$')
-        output = self.execute('ls /sys/devices/system/cpu')
+        output = self.execute('ls --color=never /sys/devices/system/cpu')
         for entry in output.split():
             if corere.match(entry):
                 num_cpus += 1
+
+        if not num_cpus:
+            raise TargetStableError('The number of CPUs cannot be null')
         return num_cpus
 
     @property


qperret:
I fixed the one place where I actually had an issue but I unfortunately suspect we could have more ... Not sure if there is a more general fix to that issue ? Maybe that 'ls --colour=never' thing should be a function on its own, so that we can use it where necessary ?

marcbonnici:
We have an existing method ``strip_bash_colors`` (https://github.com/ARM-software/devlib/blob/master/devlib/utils/misc.py#L516) that is supposed to dealing with this issue by default when performing an ``execute`` command.

Could I ask what connection you are using to your device to see why this is not happening?

qperret:
Ah, interesting ... I'm using a simple ssh connection (to a juno r0 with a debian userspace).

qperret:
Interestingly, git bisect tells me the following commit is the culprit: `511d4781646b ("exceptions: Classify transient exceptions")`. Adding/removing that commit from my branch toggles the issue repeatably. However, there is nothing obviously wrong with it AFAICT. I'll try to investigate some more ...

qperret:
OK got it, the new `will_succeed` argument introduced by the commit I mentioned above ends up overriding the default of `strip_colors`. I'll post a fix for that :)

PR #331, opened 2018-10-30
target: Ensure colours are stripped from target.execute()

qperret:
Commit 511d4781646b ("exceptions: Classify transient exceptions")
introduced a "will_succeed" argument in target.execute(). This argument
is then passed down to conn.execute(). However, since it is not
labelled, this argument happens to overwrite the "strip_colors" argument
of conn.execute because of its position in the list of parameters.

Fix this by labelling the arguments.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -362,9 +362,10 @@ class Target(object):
     # execution
 
     def execute(self, command, timeout=None, check_exit_code=True,
-                as_root=False, will_succeed=False):
-        return self.conn.execute(command, timeout, check_exit_code, as_root,
-                will_succeed)
+                as_root=False, strip_colors=True, will_succeed=False):
+        return self.conn.execute(command, timeout=timeout,
+                check_exit_code=check_exit_code, as_root=as_root,
+                strip_colors=strip_colors, will_succeed=will_succeed)
 
     def background(self, command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, as_root=False):
         return self.conn.background(command, stdout, stderr, as_root)


devlib/target.py
```diff
@@ -363,8 +363,9 @@ def get_directory(self, source_dir, dest, as_root=False):
 
     def execute(self, command, timeout=None, check_exit_code=True,
                 as_root=False, will_succeed=False):
```
douglas-raillard-arm: shouldn't Target.execute also provide strip_colors and forward it to the connection's `execute` ?

qperret: I don't really mind although I tend to prefer not adding code that has absolutely no users. Not a huge deal, though.

douglas-raillard-arm: From what I can see, Target.execute is just a stub and all *Connection.execute take the same parameter list. Avoiding slight inconsistencies like that avoid these issues in the first place

qperret: @setrofim , @marcbonnici : what do you guys prefer ? As long as this breakage gets resolved quickly I'm happy

marcbonnici: My vote would be to add in the extra parameter for consistency.

qperret: Alright let's do that then :smile:

qperret:
Since we want to keep the parameters of target.execute() and conn.execute() aligned, I followed a suggestion @douglas-raillard-arm made offline. Now target.execute() basically takes (*args, **kwargs) as arguments and passes that down to conn.execute(). That should prevent further misalignment between the two APIs.

qperret:
Fixed typos in commit message.

qperret:
Done.

PR #334, opened 2018-11-01
module/sched: Harden probe()

valschneider:
Running some tests on a VM I hit that unexpected scenario where the
required sched_domain file is there but then read_tree_values() fails
because there are no files to read, only directories.

This does a quick check that there's actual data to be read.

### DIFF
--- a/devlib/module/sched.py
+++ b/devlib/module/sched.py
@@ -195,7 +195,19 @@ class SchedProcFSData(SchedProcFSNode):
 
     @staticmethod
     def available(target):
-        return target.directory_exists(SchedProcFSData.sched_domain_root)
+        path = SchedProcFSData.sched_domain_root
+        cpus = target.list_directory(path) if target.file_exists(path) else []
+
+        if not cpus:
+            return False
+
+        # Even if we have a CPU entry, it can be empty (e.g. hotplugged out)
+        # Make sure some data is there
+        for cpu in cpus:
+            if target.file_exists(target.path.join(path, cpu, "domain0", "name")):
+                return True
+
+        return False
 
     def __init__(self, target, path=None):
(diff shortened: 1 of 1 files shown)


devlib/module/sched.py
```diff
@@ -195,7 +195,19 @@ class SchedProcFSData(SchedProcFSNode):
...
-        return target.directory_exists(SchedProcFSData.sched_domain_root)
+        path = SchedProcFSData.sched_domain_root
+        cpus = target.list_directory(path) if target.file_exists(path) else []
+
+        if not cpus:
+            return False
+
+        # Even if we have a CPU entry, it can be empty (e.g. hotplugged out)
+        # Make sure some data is there
+        for cpu in cpus:
```
valschneider: I changed that to check at least one CPU has some readable data, I don't know if that's deemed too heavy for probe().

devlib/module/sched.py
```diff
@@ -195,7 +195,20 @@ class SchedProcFSData(SchedProcFSNode):
...
+            return False
+
+        # Some systems can expose the sched_domain entry with some inner
+        # directories but there's no actual readable file. Have a quick
+        # look if that's the case or not
+        cpus = target.list_directory(path)
+        if not cpus:
+            return False
+
+        return target.file_exists(target.path.join(path, cpus[0], "domain0", "name"))
```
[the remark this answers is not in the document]
setrofim: ok, that makes sense.

devlib/module/sched.py
setrofim: This is fine, as this "short circuits" when it finds a populated node. So in the common case, this equivalent to just checking one CPU, but is more robust.

Issue #329, opened 2018-10-25
sudo malfunction not handled gracefully


v0lker:
on my root fs, `/usr/bin/sudo` wasn't setuid root which wasn't flagged up in a very helpful way

(i guess the output of utils/ssh.py: SshConnection::_execute_and_wait_for_prompt() is probably just one line (with the error message), missing the return value, and i think that's the reason why i'm getting a one-page backtrace stating that the tuple to be split didn't contain enough members (` output, exit_code_text, _ = raw_output.rsplit('\r\n', 2)`)).

i propose there should be a check at start of day and it should complain. something like
```
class SshConnection(object):
[...]
    def __init__(self,
[...]
        self.conn = ssh_get_shell(host, username, password, self.keyfile, port, timeout, False, None)                                                                
        atexit.register(self.close)
        try Exception as e:
            output = self.execute("echo hello" ,...)
            if output != "hello":
                raise TargetStableError("executing commands on the target does not seem to work {}".format(e))
[...]
```

i could send a PR if you like...

setrofim:
This sounds more like a device-specific hack than a fix. Would it make more sense for this to be in a target-specific adapter, rather than upstream?

v0lker:
no, not at all. i must have worded my proposal badly, let me try again.

current behaviour:
devlib throws a longish chain of exceptions with the last one being that a tuple doesn't have enough members.

problem: it is not obvious that it fails due to a misconfiguration of ssh/sudo/bash/...

proposed improvement: 'fail early, in an obvious manner'.
during establishment of ssh connection, it is checked whether commands can actually be executed as that user and an obvious error message is thrown (something like "you cannot execute commands via ssh, please check").

so, it is general and it would belong into `SshConnection`, the question is just whether it's useful or not - if you think that it isn't then that's fine by me.

setrofim:
I don't like the solution of executing an arbitrary echo on each and every connection creation (there may be multiple of these per session, as every thread will have its own connection, and the user can always create additional ones). Does this need to be tested for explicitly, or can this just be handled by examining the output in the exception handler on error?

v0lker:
good point - yes, certainly. i was just wondering whether you want this in the first place.

so, how about something along these lines:
```
class SshConnection[...]
    def execute[...]
        try:
            output, exit_code_text, _ = raw_output.rsplit (...)
        except ValueError as e:
            raise TargetStableError ("cannot parse ssh command reply - possible misconfiguration on target")
```

[...]

PR #337, opened 2018-11-07
utils/ssh.py: try to make failure to parse response more obvious

v0lker:
addresses #329...

### DIFF
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -205,9 +205,13 @@ class SshConnection(object):
         try:
             with self.lock:
                 _command = '({}); __devlib_ec=$?; echo; echo $__devlib_ec'.format(command)
-                raw_output = self._execute_and_wait_for_prompt(
-                    _command, timeout, as_root, strip_colors)
-                output, exit_code_text, _ = raw_output.rsplit('\r\n', 2)
+                full_output = self._execute_and_wait_for_prompt(_command, timeout, as_root, strip_colors)
+                split_output = full_output.rsplit('\r\n', 2)
+                try:
+                    output, exit_code_text, _ = split_output
+                except ValueError as e:
+                    raise TargetStableError(
+                        "cannot split reply (target misconfiguration?):\n'{}'".format(full_output))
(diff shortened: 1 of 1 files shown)


devlib/utils/ssh.py
```diff
@@ -205,9 +205,12 @@ def execute(self, command, timeout=None, check_exit_code=True,
...
                 _command = '({}); __devlib_ec=$?; echo; echo $__devlib_ec'.format(command)
-                raw_output = self._execute_and_wait_for_prompt(
-                    _command, timeout, as_root, strip_colors)
-                output, exit_code_text, _ = raw_output.rsplit('\r\n', 2)
+                split_output = self._execute_and_wait_for_prompt(
+                    _command, timeout, as_root, strip_colors).rsplit('\r\n', 2)
+                try:
+                    output, exit_code_text, _ = split_output
+                except ValueError as e:
+                    raise TargetStableError("cannot parse ssh command reply - target misconfiguration?")
```
setrofim: Maybe save unsplit output and report it here, as part of the error?

v0lker: yes, you're right that could be useful.  i considered that but decided not to, because of the effort involved in sanitising it.

it might contain unprintable characters and it might be very long, so, i think the correct approach would be to

`raise TargetStableError("cannot parse ssh reply (target misconfiguration: {}".format(limit_length(80, hexdump(split_output)))`

i shouldn't be logged separately, because it is used to determine certain features (`needs_root`, IIRC)

if you think it's worthwhile, i'll implement it that way - ?

setrofim: I would just filter out or encode unprintable characters, rather than produce hexdump, as the common case is that the output will contain text.

Do not limit the length at this stage -- report the whole thing; it's up to the code handling and/or reporting the error to perform formatting such as length limiting.

v0lker: in that case the code handling it can also do the conversion, i'd say...

setrofim: Yeah, I'm cool wit that.

PR #338, opened 2018-11-08
module/sched: Sched domain flag changes

valschneider:


devlib/module/sched.py
```diff
@@ -146,29 +162,56 @@ class SchedDomain(SchedProcFSNode):
...
-        "SD_ASYM_CPUCAPACITY"    : 0x0040,  # Groups have different max cpu capacities
-        "SD_SHARE_CPUCAPACITY"   : 0x0080,  # Domain members share cpu capacity
-        "SD_SHARE_POWERDOMAIN"   : 0x0100,  # Domain members share power domain
-        "SD_SHARE_PKG_RESOURCES" : 0x0200,  # Domain members share cpu pkg resources
-        "SD_SERIALIZE"           : 0x0400,  # Only a single load balancing instance
-        "SD_ASYM_PACKING"        : 0x0800,  # Place busy groups earlier in the domain
-        "SD_PREFER_SIBLING"      : 0x1000,  # Prefer to place tasks in a sibling domain
-        "SD_OVERLAP"             : 0x2000,  # sched_domains of this level overlap
-        "SD_NUMA"                : 0x4000,  # cross-node balancing
+        "SD_LOAD_BALANCE" : SchedDomainFlag(
```
douglas-raillard-arm: This dict can avoid repetition of names:
```
SD_FLAGS = {
flag.name: flag for flag in (
SchedDomainFlag(XXX),
SchedDomainFlag(YYY),
...
)}
```

devlib/module/sched.py
```diff
@@ -146,29 +162,56 @@ class SchedDomain(SchedProcFSNode):
...
 
     # Checked to be valid from v4.4
     SD_FLAGS_REF_PARTS = (4, 4, 0)
 
+    def __init__(self, nodes):
+        super(SchedDomain, self).__init__(nodes)
+
+        obj_flags = set()
+        for flag in self.SD_FLAGS.values():
+            if self.flags & flag.value == flag.value:
```
douglas-raillard-arm: self.flags does not seem to be defined yet here.

valschneider: It's built automagically from the parent __init__() when parsing the procfs.

douglas-raillard-arm: In that case, it's probably not a good idea to wildly change the type of something created by the mother class itself. There is no way a base class method will be happy with a set of SchedDomainFlag object if it expects an integer. Creating another attribute seems more appropriate, like `flags_obj` maybe ?

valschneider: The way I've wrote up `SchedProcFSNode` is that you still have access to the "raw" read_tree_file procs, but the object view of the procfs is free to use more sophisticated types.

For instance, the flags for a cpu/domain can be obtained via
```python
sd_info.procfs["cpu0"]["domain0"]["flags"] (int)
```
But a much handier version can be obtained via
```python
sd_info.cpus[0].domains[0].flags (set(SchedDomainFlag))
```

So the base classes can depend on the untouched raw procfs version.

valschneider:
> That will fail if for any reason another instance is used

Seeing how `flags` is rebuilt, there's no reason for that to happen.

> >>> "SD_ASYM_CPUCAPACITY" in domain.flags_str

The reason for having those flags defined in a class dict is to get autocompletion (I only really know 2 of those flags), and you get the `description` if you're unsure as to what the flag is doing. To me that makes more sense to check for inclusion using those pre-defined objects rather than pure strings.

douglas-raillard-arm:
If for example you serialize then deserialize an this `domain` object, this will not work anymore:
```
>>> SchedDomain.SD_FLAGS["SD_ASYM_CPUCAPACITY"] in domain.flags
```
This is quite dangerous as it would fail silently. The string approach fixes that without having to introduce real singletons. If real singletons are needed, it seems a better idea to rely on the enum34 module that should provide a correct implementation of it.

valschneider:
Hmm so I don't know how `enum34` would work for both Python2 & Python3.

Otherwise I could change the objectified `domain.flags` to be a `dict(SchedDomainFlag)`, so we can test by keystring and still have access to a full-fledged object.

setrofim:
There is also an `enum` implementation in WA (which could be moved into `devlib`) that could be use for this:

https://github.com/ARM-software/workload-automation/blob/master/tests/test_utils.py#L98

This will work on Python 2 and 3.

valschneider:
@setrofim I'll have a look, thanks for the pointer.

douglas-raillard-arm:
Looking at WA's enums, it looks like we would have to somehow implement a __reduce_ex__ method for it to make it possible to deserialize instances of the classes `Enum` created on the fly. Alternatively, it is possible to make the classes created with enum() available under a stable name, registered as the classe's `__name__`/`__qualname__` and `__module__`, like what `collections.namedtuple` provide.
The singleton issue would be addressed by using equality comparison, since it WA's `level` implements `__eq__` and `__hash__`.

valschneider:
@douglas-raillard-arm I went the enum34 way

devlib/module/sched.py
```diff
@@ -138,52 +139,74 @@ def __init__(self, nodes):
...
-    SD_ASYM_CPUCAPACITY    = 0x0040  # Groups have different max cpu capacities
-    SD_SHARE_CPUCAPACITY   = 0x0080  # Domain members share cpu capacity
-    SD_SHARE_POWERDOMAIN   = 0x0100  # Domain members share power domain
-    SD_SHARE_PKG_RESOURCES = 0x0200  # Domain members share cpu pkg resources
-    SD_SERIALIZE           = 0x0400  # Only a single load balancing instance
-    SD_ASYM_PACKING        = 0x0800  # Place busy groups earlier in the domain
-    SD_PREFER_SIBLING      = 0x1000  # Prefer to place tasks in a sibling domain
-    SD_OVERLAP             = 0x2000  # sched_domains of this level overlap
-    SD_NUMA                = 0x4000  # cross-node balancing
+    SD_LOAD_BALANCE =        0x0001, "Do load balacing on this domain"
```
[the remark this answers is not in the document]
valschneider: Fixed

marcbonnici:
@douglas-raillard-arm  Are you happy with the changes to the PR?

douglas-raillard-arm:
LGTM

PR #339, opened 2018-11-12
[RFC] Fix quoting

douglas-raillard-arm:
This PR address some quoting issues in various places where a string containing a command is fed to a shell, directly or indirectly e.g. through `su`.

The third commit "escaping: Use pipes.quote instead of escape_*" unifies escaping in devlib so that the strings provided to the library are used as is, without further interpolation. That could break code relying on referencing environment variable in provided strings, and code that manually escapes strings in order to work around that. It should be noted that any calling code can make use of `os.path.expandvars` to interpolate env var, and that devlib seems to use both single and double quoting in various places, without clear documented consistency.

This PR was tested on SSH connections for a LinuxTarget only, so it would benefit from broader testing and/or close code inspection.

The original issue is Target.push/pull, that cannot deal correctly with paths containing `(` and `)`:
```
  File "/home/dourai01/Data/Git/lisa/py3k/modules_root/devlib/target.py", line 313, in push
    self.conn.push(source, dest, timeout=timeout)
  File "/home/dourai01/Data/Git/lisa/py3k/modules_root/devlib/utils/ssh.py", line 191, in push
    return self._scp(source, dest, timeout)
  File "/home/dourai01/Data/Git/lisa/py3k/modules_root/devlib/utils/ssh.py", line 321, in _scp
    command_redacted, e.output))
devlib.exception.HostError: Failed to copy file with '/usr/bin/scp -r -i /tmp/id_rsa  /home/dourai01/Data/Git/lisa/py3k/results/20181112_11:33:24_02ebf5b9de5d48c7a296d1f773596551/PELTTaskTest.test_load_avg_range/PELTTaskTest:test_load_avg_range(allowed_error_pct=ForcedType)/e95235eb4fe34c21837516f51b67fd5b/lisa.env.TestEnv/1/rta_calib-20181112_113350.106580/rta_calib_cpu0.json "root"@"pwrsft-hikey620-1":"/root/devlib-target/rta_calib_cpu0.json"'. Output:
```

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -26,6 +26,7 @@ import threading
 import xml.dom.minidom
 import copy
 from collections import namedtuple, defaultdict
+from pipes import quote
 
 from devlib.host import LocalConnection, PACKAGE_BIN_DIRECTORY
 from devlib.module import get_module
@@ -36,7 +37,7 @@ from devlib.exception import (DevlibTransientError, TargetStableError,
 from devlib.utils.ssh import SshConnection
 from devlib.utils.android import AdbConnection, AndroidProperties, LogcatMonitor, adb_command, adb_disconnect, INTENT_FLAGS
 from devlib.utils.misc import memoized, isiterable, convert_new_lines
-from devlib.utils.misc import commonprefix, escape_double_quotes, merge_lists
+from devlib.utils.misc import commonprefix, merge_lists
 from devlib.utils.misc import ABI_MAP, get_cpu_name, ranges_to_list
 from devlib.utils.types import integer, boolean, bitmask, identifier, caseless_string, bytes_regex
 
@@ -555,7 +556,7 @@ class Target(object):
         raise IOError('No usable temporary filename found')
 
     def remove(self, path, as_root=False):
-        self.execute('rm -rf "{}"'.format(escape_double_quotes(path)), as_root=as_root)
+        self.execute('rm -rf {}'.format(quote(path)), as_root=as_root)
 
     # misc
     def core_cpus(self, core):
@@ -891,7 +892,7 @@ class LinuxTarget(Target):
         pass
 
     def kick_off(self, command, as_root=False):
-        command = 'sh -c "{}" 1>/dev/null 2>/dev/null &'.format(escape_double_quotes(command))
+        command = 'sh -c {} 1>/dev/null 2>/dev/null &'.format(quote(command))
         return self.conn.execute(command, as_root=as_root)
 
     def get_pids_of(self, process_name):
@@ -925,7 +926,7 @@ class LinuxTarget(Target):
             return filtered_result
 
     def list_directory(self, path, as_root=False):
-        contents = self.execute('ls -1 "{}"'.format(escape_double_quotes(path)), as_root=as_root)
+        contents = self.execute('ls -1 {}'.format(quote(path)), as_root=as_root)
         return [x.strip() for x in contents.split('\n') if x.strip()]
 
     def install(self, filepath, timeout=None, with_name=None):  # pylint: disable=W0221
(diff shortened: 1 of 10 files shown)


devlib/utils/ssh.py
```diff
@@ -264,7 +259,7 @@ def _execute_and_wait_for_prompt(self, command, timeout=None, as_root=False, str
             # As we're already root, there is no need to use sudo.
             as_root = False
         if as_root:
-            command = self.sudo_cmd.format(escape_single_quotes(command))
+            command = self.sudo_cmd.format(quote(command))
```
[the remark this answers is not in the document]
douglas-raillard-arm: Maybe we could issue a `warnings.warn()` in the constructor if `sudo_cmd` contains the string `"{}"` or `'{}'`, and replace it by a plain `{}` ? AFAIK `'{}'`/`"{}"` cannot have a different meaning than the placeholder for the command, and we know by construction that we will escape it correctly, so it should at worst lead to less evaluation from the shell (therefore no security issue).

devlib/target.py
```diff
@@ -1359,9 +1361,9 @@ def broadcast_media_mounted(self, dirpath, as_root=False):
         """
```
marcbonnici: Since this method is being quoted it is probably a good idea to quote the file path in the above `broadcast_media_scan_file` as well.

devlib/utils/ssh.py
```diff
@@ -308,12 +303,11 @@ def _scp(self, source, dest, timeout=30):
         # only specify -P for scp if the port is *not* the default.
         port_string = '-P {}'.format(self.port) if (self.port and self.port != 22) else ''
         keyfile_string = '-i {}'.format(self.keyfile) if self.keyfile else ''
-        command = '{} -r {} {} {} {}'.format(scp, keyfile_string, port_string, source, dest)
+        command = '{} -r {} {} {} {}'.format(scp, quote(keyfile_string), port_string, quote(source), quote(dest))
```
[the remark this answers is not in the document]
marcbonnici: I think so and I agree for passwords (unless we could detect something mandatory was missing early ourselves) unlike here where it is optional so it will depend on the situation as to the correct approach.

douglas-raillard-arm:
The PR has now been lightly tested on an Android target (probably not much more than `target.push()` and `target.pull()` however, since it was only a run of the LISA synthetic tests)

devlib/target.py
```diff
@@ -1352,16 +1354,16 @@ def broadcast_media_scan_file(self, filepath):
...
         Force a re-index of the mediaserver cache for the specified file.
         """
-        command = 'am broadcast -a android.intent.action.MEDIA_SCANNER_SCAN_FILE -d file://'
-        self.execute(command + filepath)
+        command = 'am broadcast -a android.intent.action.MEDIA_SCANNER_SCAN_FILE -d {}'
+        self.execute(command.format(quote('file://' + filepath))
```
marcbonnici: There is a missing `)` at the end of this line causing a syntax error.

devlib/utils/misc.py
```diff
@@ -416,24 +417,50 @@ def convert_new_lines(text):
     """ Convert new lines to a common format.  """
     return text.replace('\r\n', '\n').replace('\r', '\n')
 
+def sanitize_cmd_template(cmd):
+    msg = (
+        '''Quoted placeholder should not be used, as it will result in quoting the text twice. {} should be used instead of '{}' or "{}".'''
+    )
+    for unwanted in ('"{}"', "'{}'"):
+        if unwanted in cmd:
+            warnings.warn(msg, stacklevel=2)
```
marcbonnici: Would it be worth adding the cmd itself to the warning to make it a bit more obvious what is causing it?

devlib/utils/misc.py
```diff
@@ -416,24 +417,50 @@ def convert_new_lines(text):
     """ Convert new lines to a common format.  """
     return text.replace('\r\n', '\n').replace('\r', '\n')
 
+def sanitize_cmd_template(cmd):
+    msg = (
+        '''Quoted placeholder should not be used, as it will result in quoting the text twice. {} should be used instead of '{}' or "{}" in the template: '''
```
marcbonnici: Is the message missing the command placeholder?

douglas-raillard-arm: It somehow got lost, I'll concatenate the strings instead, to avoid double curly braces in the message string so it can be grepped more easily

PR #340, opened 2018-11-13
module/cpufreq: Add userspace special case for use_governor()

valschneider:
Frequencies are not considered as governor tunables by
list_governor_tunables(), so add a special case to restore userspace
frequency when using use_governor().

While at it, reword the internal "governor" variable to not clash
with the parameter of the same name.

I was not sure if that should be handled in the tunables themselves (i.e. special-case `userspace` in `list_governor_tunables` & co) or if it was fine to special case them in the context manager. I went with this implementation to get the discussion going.


valschneider:
Wrong command! Forcepushed the wrong branch here, hang on...

valschneider:
...Fixed

PR #351, opened 2018-12-03
ftrace: Turn FtraceCollector into a context manager

douglas-raillard-arm:
This allows using an FtraceCollector object as a context manager, so
tracing will stop even if the devlib application is interrupted, or if
an exception is raised.

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -323,17 +323,14 @@ You can collected traces (currently, just ftrace) using
    # the buffer size to be used.
    trace = FtraceCollector(t, events=['power*'], buffer_size=40000)
 
-   # clear ftrace buffer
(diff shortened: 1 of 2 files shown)


valschneider:
Would it make more sense to let `TraceCollector` be used as a context manager in general, rather than just `FtraceCollector`?

douglas-raillard-arm:
Right, I haven't considered that, I'll respin the PR

doc/overview.rst
```diff
@@ -326,14 +326,13 @@ You can collected traces (currently, just ftrace) using
    # clear ftrace buffer
    trace.reset()
```
derkling: Is not that already part of the ctx manager ?

PR #352, opened 2018-12-06
cgroups: allow to set dot-named attributes

derkling:
Certain controller, especially on CGroups v2, uses attributes whith a
"dot" in their name, e.g. "cpus.allowed". Unfortunately, those are not
valid function parameter names in Python.

Let's fix this by allowing to define those attributes as (key, value)
tuples. Thus, for example, it will be possible use the simplest call for
each attribute we want to configure:

   cg.set(cpus="0-5", ("cpus.allowed", "1,2"))

Suggested-by: Douglas Raillard <douglas.raillard@arm.com>

### DIFF
--- a/devlib/module/cgroups.py
+++ b/devlib/module/cgroups.py
@@ -302,7 +302,8 @@ class CGroup(object):
 
         return conf
 
-    def set(self, **attrs):
+    def set(self, *args, **attrs):
+        attrs.update(args)
         for idx in attrs:
             if isiterable(attrs[idx]):
                 attrs[idx] = list_to_ranges(attrs[idx])


douglas-raillard-arm:
The example in the commit message needs to have the positional and named parameters swapped, otherwise it will raise this kind of exception: `SyntaxError: positional argument follows keyword argument`

setrofim:
how is the new syntax substantially different from simply doing `cg.set(**{"cpus.allowed": "1,2"})`, which you could already do without this change?

douglas-raillard-arm:
Looks like they both do the same (slightly less syntactic overhead for the tuple), and since the parameters to `cg.set()` are already all "reserved" for Cgroups attributes, it shouldn't restrain API evolution anyway. It also looks like Python is accepting non-valid identifiers as kwargs keys, as long as they are strings so it is indeed enough for all the use cases.

setrofim:
I would argue the syntactic overhead is lower for the dict, as you don't have to parenthesize each set of argument-value pairs, or remember the "weird" ordering (the mistake made in the commit message)

derkling:
So, the trick proposed by @setrofim is very "pythonic"... I'm wondering how many users could have come up with that syntax if not at least suggested.
However, the ordering propblem is not there also for what you propose?
For example, I'me getting:
```python
    f(**{'cpu.allowed': 1}, cpus='2')
                          ^
SyntaxError: invalid syntax
```
While this is perfectly fine:
```python
f(cpus='2', **{'cpu.allowed': 1})
```

So, I would say that in any case we should check that we don't mix the syntax, isn't it ?

douglas-raillard-arm:
> However, the ordering problem is not there also for what you propose?

Looks like Python 3.4 fails at parsing that, but 3.7 is perfectly happy with it

derkling:
Well... we cannot keep following the (crazy) Python minor version changes...  :/
Should not 3.4 considered a "stable" version of the language ?

douglas-raillard-arm:
I was just pointing out that the "issue" in the Python syntax has been fixed in 3.7, so long term speaking the requirement on the order of keyword arguments and the dict unpacking vanishes without us having to do anything.

setrofim:
> I'm wondering how many users could have come up with that syntax.

Yes, I agree -- a note should be added to the documentation to highlight this, as it may not be obvious to everyone.

> So, I would say that in any case we should check that we don't mix the syntax, isn't it ?

Yes, correct; however, `kwargs` expansions always go at the end, and in this case the correct ordering is a lot more intuitive  (to my mind, at least -- this is subjective).

My main issue with this pull is that there already exists a way of dealing with the situation this is designed to handle, and the new method isn't really an improvement on the existing one. To me (again, there is a degree of subjectivity here), this adds little to to the code aside from churn and extra complexity -- now the use needs to understand both the dict expansion _and_ the tuple handling, and the correct ordering for each.

However, if you do feel that the tuple method is a lot more obvious and  would strongly prefer it to the dict expansion, I am not dead-set again this being merged.

derkling:
No, I'm not pushing for a merging this. I like better the dictionary proposal, we just have to properly document it. It should be part of the API documentation which unfortunately it's still missing for this module (my bad).

However, I've already in my todo list the task to update the cgroup module to properly support CGroups v2 and fix some other small issues it has. I'll take that chance to also add the missing documentation.

Meanwhile... let's close this PR ;)

PR #353, opened 2018-12-13
module/cgroups: Execute command as root when launching cmd in cgroup

marcbonnici:
On some devices root is required to configure cgroups therefore run the
command as root if available.

### DIFF
--- a/devlib/module/cgroups.py
+++ b/devlib/module/cgroups.py
@@ -432,16 +432,20 @@ class CgroupsModule(Module):
                 .format(self.cgroup_root, self.target.shutils,
                         cgroup, cmdline)
 
-    def run_into(self, cgroup, cmdline):
+    def run_into(self, cgroup, cmdline, as_root=None):
         """
         Run the specified command into the specified CGroup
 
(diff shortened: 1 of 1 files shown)


devlib/module/cgroups.py
```diff
@@ -441,7 +441,7 @@ def run_into(self, cgroup, cmdline):
         :returns: Output of command.
         """
         cmd = self.run_into_cmd(cgroup, cmdline)
-        raw_output = self.target.execute(cmd)
+        raw_output = self.target.execute(cmd, as_root=self.target.is_rooted)
```
derkling: What about exposing this via the API:
```python
def run_into(self, cgroup, cmdline, as_root=self.target.is_rooted):
```

Maybe there could be use-cases in which we want to test that a user cannot spawn tasks into a cgroup...

marcbonnici: I've updated this to allow overriding the behaviour.

setrofim: Should update the docstring to document the new parameter.

marcbonnici: Updated.

PR #356, opened 2018-12-21
cgroups.py: strip slashes, don't drop first char

v0lker:
this makes it cope with both, names starting with or without slashes.

currently, if you omit the leading slash, it will just remove the first character of the directory instead, which is annoying.

the trailing slash isn't needed, if i'm wrong could lstrip() instead...

### DIFF
--- a/devlib/module/cgroups.py
+++ b/devlib/module/cgroups.py
@@ -262,8 +262,9 @@ class CGroup(object):
 
         # Control cgroup path
         self.directory = controller.mount_point
+
         if name != '/':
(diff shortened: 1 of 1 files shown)


derkling:
I don't like client code where we can have `group_name` instead of `/group_name`, and I find the second version much more clear in defining that "group_name" is a top-level control group. 

Unfortunately I was late for this review, but I would have preferred a semantics where we force all cgroup names to start with a `/` just to have more consistent naming. The original code probably failed just on checking this requirement and raising an exception in case the specified cgroup name was not starting with `/`... @v0lker could you please send another fix where we enforce this behaviour?

PR #361, opened 2019-01-24
utils/ssh: Implement work around for issue in pexpect

marcbonnici:
As detailed in https://github.com/pexpect/pexpect/issues/552 when
sending a command with exactly the length of the prompt less than the
window width, an extra prompt is introduced into the returned data
causing the matching to fail and only return part of the commands
output. To workaround this check the length of the command to be
submitted and if of the specific length add an additional whitespace
character to the end of the command to prevent this behaviour.

### DIFF
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -41,7 +41,8 @@ from pexpect import EOF, TIMEOUT, spawn
 # pylint: disable=redefined-builtin,wrong-import-position
 from devlib.exception import (HostError, TargetStableError, TargetNotRespondingError,
                               TimeoutError, TargetTransientError)
-from devlib.utils.misc import which, strip_bash_colors, check_output, sanitize_cmd_template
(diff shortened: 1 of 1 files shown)


devlib/utils/ssh.py
```diff
@@ -321,6 +322,21 @@ def _scp(self, source, dest, timeout=30):
...
+            command += ' '
+        self.conn.sendline(command)
+
+    @memoized
+    def _get_prompt_length(self):
+        self.conn.sendline()
+        self.conn.prompt()
+        return len(self.conn.after)
+
+    @memoized
```
setrofim: This should not be memoized, as window size might change during the run  (e.g. terminal pane resized in tmux).

marcbonnici: Discussed offline and decided to leave as is to ensure this is as quick as possible and should not cause issues.

PR #362, opened 2019-01-25
target: add MountPoints abstraction

derkling:
This adds the required support to query the mount points available on a
target. The new class adds a set of utility methods to easily identify
mounted file systems and their attributes.


setrofim:
How is this different from the existing `target.list_file_systems()`?

PR #366, opened 2019-01-31
target: Speed-up read_tree_values()

qperret:
Since target.read_tree_values() has been modified to use tar as a way to
fetch the content of files from the target, it is more robust, but also
much slower. Since that level of robustness is in practice required only
for very specific use-cased, re-introduce the old way of doing read_tree
using find and grep.

read_tree_values() gains a new parameter to specify how files should be
read from the target, with or without tar. It defaults to the old way
of doing things.

### DIFF
--- a/devlib/bin/scripts/shutils.in
+++ b/devlib/bin/scripts/shutils.in
@@ -255,6 +255,28 @@ sched_get_kernel_attributes() {
 # Misc
 ################################################################################
 
+read_tree_values() {
+    BASEPATH=$1
+    MAXDEPTH=$2
+
+    if [ ! -e $BASEPATH ]; then
+        echo "ERROR: $BASEPATH does not exist"
+        exit 1
+    fi
+
+    PATHS=$($BUSYBOX find $BASEPATH -follow -maxdepth $MAXDEPTH)
+    i=0
+    for path in $PATHS; do
+        i=$(expr $i + 1)
+        if [ $i -gt 1 ]; then
+            break;
+        fi
+    done
+    if [ $i -gt 1 ]; then
+        $BUSYBOX grep -s '' $PATHS
+    fi
+}
+
 read_tree_tgz_b64() {
     BASEPATH=$1
     MAXDEPTH=$2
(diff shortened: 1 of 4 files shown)


douglas-raillard-arm:
Shouldn't the code that works all the time be the default, and have an option to speed it up when it is known to not break ? That would avoid ending up with user code that seems to work on a machine and then starts failing on another one just because an "annoying" file shows up there in sysfs.

If tar=False becomes the default, the `hwmon` module should probably make use of `tar=True` since it is known to break on some hosts with the previous `read_tree_values` implementation.

qperret:
It feels to me that the significant majority of cases work just fine with the old grep-based implementation. So it probably makes sense to have the default that suits most users.

I'll add a commit to fix hwmon ;)

douglas-raillard-arm:
I don't know about how WA uses devlib, but you could argue that the significant majority of cases in LISA work just fine with the functionally correct tar-based implementation. Next user tripping on that (maybe through a devlib module, maybe via 3rd party code) is likely to unnecessarily loose time on an issue that has been identified, debugged and solved in backward compatible way, but was just not enabled by default.

qperret:
Yeah but I also care about keeping Lisa fast, and that tar-based implementation made it worse. And the option is documented so users should read the doc IMO :).
I added the fix for hwmon.

setrofim:
It can be argued that it is preferable to have an overt failure than an inexplicable slow down. In the former case, it is at least clear than there is a problem, in the latter the user might not even be aware that the issue can be rectified.

In either case the docs should be updated to describe this issue and why ``tar`` flag even exists.

qperret:
Ah, right, the docstring isn't enough ... I'll go write something in doc/

douglas-raillard-arm:
After a bit of investigations, not handling MAXDEPTH in the shell script (which forces to have a loop+find) and just copying everything with `cp -R` restores normal speed. It is possible to handle that on the Python side if necessary. That won't work if the tree contains really big files deep in the hierarchy, but for /sys/devices/system/cpu it does the job well and simplifies `shutils.in`.

qperret:
I added a paragraph to the doc of read_tree_values(). Let me know if that works for you.

PR #376, opened 2019-03-14
trace: Add DmesgCollector

douglas-raillard-arm:
Allows collecting dmesg output and parses it for easy filtering.

### DIFF
--- a/devlib/trace/dmesg.py
+++ b/devlib/trace/dmesg.py
@@ -1,0 +1,187 @@
+#    Copyright 2019 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from __future__ import division
+import re
+from itertools import takewhile
+from datetime import timedelta
+
+from devlib.trace import TraceCollector
+
+
+class KernelLogEntry(object):
+    """
+    Entry of the kernel ring buffer.
+
+    :param facility: facility the entry comes from
+    :type facility: str
+
+    :param level: log level
+    :type level: str
+
+    :param timestamp: Timestamp of the entry
+    :type timestamp: datetime.timedelta
+
+    :param msg: Content of the entry
+    :type msg: str
+    """
+
+    _TIMESTAMP_MSG_REGEX = re.compile(r'\[(.*?)\] (.*)')
+
+    def __init__(self, facility, level, timestamp, msg):
+        self.facility = facility
+        self.level = level
+        self.timestamp = timestamp
+        self.msg = msg
+
+    @classmethod
+    def from_str(cls, line):
+        """
+        Parses a "dmesg --decode" output line, formatted as following:
+        kern  :err   : [3618282.310743] nouveau 0000:01:00.0: systemd-logind[988]: nv50cal_space: -16
+
+        Or the more basic output:
+        [3618282.310743] nouveau 0000:01:00.0: systemd-logind[988]: nv50cal_space: -16
+
+        """
+
+        def parse_timestamp_msg(line):
+            match = cls._TIMESTAMP_MSG_REGEX.match(line.strip())
+            if not match:
+                raise ValueError('dmesg entry format not recognized: {}'.format(line))
+            timestamp, msg = match.groups()
+            return timestamp, msg
+
+        # If we can parse the timestamp directly, that is a basic line
+        try:
+            timestamp, msg = parse_timestamp_msg(line)
+        except ValueError:
+            facility, level, remainder = line.split(':', 2)
+            timestamp, msg = parse_timestamp_msg(remainder)
+            facility = facility.strip()
+            level = level.strip()
+        else:
+            facility = None
+            level = None
+
+        return cls(
+            facility=facility,
+            level=level,
+            timestamp=timedelta(seconds=float(timestamp.strip())),
+            msg=msg.strip(),
+        )
+
+    def __str__(self):
+        if self.facility and self.level:
+            prefix = '{facility}:{level}:'.format(
+                facility=self.facility,
+                level=self.level,
+            )
+        else:
+            prefix = ''
+
+        return '{prefix}[{timestamp}] {msg}'.format(
+            timestamp=self.timestamp.total_seconds(),
+            msg=self.msg,
+            prefix=prefix,
+        )
+
+
+class DmesgCollector(TraceCollector):
+    """
+    Dmesg output collector.
+
+    :param level: Minimum log level to enable. All levels that are more
+        critical will be collected as well.
+    :type level: str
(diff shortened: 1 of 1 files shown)


devlib/trace/dmesg.py
```diff
@@ -0,0 +1,185 @@
...
+        Parses a "dmesg --decode" output line, formatted as following:
+        kern  :err   : [3618282.310743] nouveau 0000:01:00.0: systemd-logind[988]: nv50cal_space: -16
+
+        Or the more basic output:
+        [3618282.310743] nouveau 0000:01:00.0: systemd-logind[988]: nv50cal_space: -16
+
+        """
+
+        def parse_ts_msg(line):
+            ts_msg_regex = r'\[(.*?)\] (.*)'
```
marcbonnici: Pre-compile the regex and reuse instead of compiling for each line?

devlib/trace/dmesg.py
```diff
@@ -0,0 +1,185 @@
...
+                raise ValueError('dmesg entry format not recognized: {}'.format(line))
+            ts, msg = match.groups()
+            return ts, msg
+
+        # If we can parse the timestamp directly, that is a basic line
+        try:
+            ts, msg = parse_ts_msg(line)
+        except ValueError:
+            facility, level, remainder = line.split(':', 2)
+            ts, msg = parse_ts_msg(remainder)
```
marcbonnici: Wrap this in a try catch as well in case the second parsing attempt also fails?

devlib/trace/dmesg.py
```diff
@@ -0,0 +1,185 @@
...
+        "alert",        # action must be taken immediately
+        "crit",         # critical conditions
+        "err",          # error conditions
+        "warn",         # warning conditions
+        "notice",       # normal but significant condition
+        "info",         # informational
+        "debug",        # debug-level messages
+    ]
+
+    def __init__(self, target, level=LOG_LEVELS[-1], facility='kern'):
```
valschneider: Could we have a check for that facility being available (other than the `dmesg` exit code check) ?

I don't know how much you can rely on the output of `dmesg -h` but you could parse facilities (and log levels BTW) from there. Sort of something like this (but as a static/classmethod maybe) https://github.com/ARM-software/devlib/blob/master/devlib/trace/systrace.py#L61

devlib/trace/dmesg.py
douglas-raillard-arm: If this one fails, that means the parser is broken so we want to know about it to fix it. The 1st failure is somehow expected and should always detect correctly what it is supposed to detect (i.e. something before the timestamp `[42.4242]` or not).

devlib/trace/dmesg.py
douglas-raillard-arm: Here is the busybox's dmesg help :-)

```
# dmesg -h
dmesg: invalid option -- 'h'
BusyBox v1.29.2 (2018-12-18 16:55:05 GMT) multi-call binary.

Usage: dmesg [-c] [-n LEVEL] [-s SIZE]

Print or control the kernel ring buffer

	-c		Clear ring buffer after printing
	-n LEVEL	Set console logging level
	-s SIZE		Buffer size
	-r		Print raw message buffer
```

Note that there isn't even a `--help` option, so we can safely forget about any "introspection" capabilities being always available.

There is also relatively little value in practice, since 99% of devlib users will probably restrict that to `kern`. I had to get the list of log levels for another reason: util-linux `dmesg` wants an explicit list of all levels to show, i.e. it will not show "levels above X", it will show "level X". Since we usually want the "levels above X" behavior, I had to get the ordered list of levels.

valschneider: Fair enough!

devlib/trace/dmesg.py
```diff
@@ -0,0 +1,187 @@
...
+            prefix = ''
+
+        return '{prefix}[{timestamp}] {msg}'.format(
+            timestamp=self.timestamp.total_seconds(),
+            msg=self.msg,
+            prefix=prefix,
+        )
+
+
+class DmesgCollector(TraceCollector):
```
qais-yousef: Not your problem, but I find this hierarchy confusing. Could the base class be named something different like BufferCollector? I appreciate this is probably outside the scope of this patch by the way. But dmesg and tracing has no relationship at all.

devlib/trace/dmesg.py
```diff
@@ -0,0 +1,187 @@
...
+            cmd = 'dmesg'
+        else:
+            cmd = 'dmesg --facility={facility} --force-prefix --decode --level={levels}'.format(
+                levels=','.join(levels_list),
+                facility=self.facility,
+            )
+
+        self.dmesg_out = self.target.execute(cmd)
+
+    def get_trace(self, outfile):
```
qais-yousef: Nit: get_log() or get_buffer() makes more sense here if it can be changed. I appreciate this is overriding a similarly named method though. So again out of the scope of this patch probably.

valschneider: `get_trace()` is part of the `TraceCollector` API (https://github.com/ARM-software/devlib/blob/master/devlib/trace/__init__.py), so we have to provide it.

In the end dmesg is a trace itself (you get events at given timestamps), so it's not too outlandish if you look at it the right way.

devlib/trace/dmesg.py
valschneider: Devlib's notion of tracing != Ftrace. Here it's just something you can `start`, `stop`, and get a log/trace/item/`<find name you like>` containing **something** that happening during this `start` & `stop`.

devlib/trace/dmesg.py
qais-yousef: Yes I know. I did say that in my comment that I appreciate it's overriding a similarly named function and that it's probably outside the scope of this patch.

>n the end dmesg is a trace itself (you get events at given timestamps), so it's not too outlandish if you look at it the right way.

I think you stared at these stuff for too long. A dmesg is not a trace. Not even remotely.

devlib/trace/dmesg.py
qais-yousef: Tracing doesn't have to be ftrace. But tracing implies a step by step break down of something. Again it seems you stared at this code for too long.

Anyways. This comment was directed towards the maintainers if they'd like to consider renaming following Douglas suggestion.

devlib/trace/dmesg.py
douglas-raillard-arm: I guess it depends on the usage of printk: sometimes the order does not hugely matter and it's just a way to log stuff to show to the user (some boot-time messages). Sometimes printk can be treated as timestamped events, like when enabling cpufreq debugging, in which case it is very much like a basic trace system (and I intend to mix that with other actual ftrace events on a timeline, so they can definitely be treated as equivalent).

douglas-raillard-arm: In any case, I agree that `TraceCollector` is not the best name for the collector's base. Something agnostic like `CollectorBase` may be more appropriate, but that would mean rename + alias on the old name.

setrofim:
> however for now I'm happy to merge this PR as it is and update it later if required.

Agreed.  The naming issue is orthogonal to the particulars of this PR, and is worth putting some thought into to get right; there is no point in blocking this PR on that. 

I agree that "trace" is not an appropriate term for everything that's currently being collected via the `TraceCollector` API. The intention for this is to provide an interface for collection of everything that is not purely numeric measurements.

qais-yousef:
In case my comment was seeing as an objection. I'm okay with this change. As I mentioned my comments are probably outside the scope of this patch, so certainly shouldn't block this PR :)

PR #377, opened 2019-03-18
Scheduler module fixes

valschneider:


valschneider:
@credp This should fix the issues you've seen.

Issue #378, opened 2019-03-19
Rename TraceCollector


setrofim:
As discussed here: https://github.com/ARM-software/devlib/pull/376, "trace" is not an appropriate term for everything that is being collected by `TraceCollector` implementation. Thus, the name is misleading and should be changed. `CollectorBase`, suggested in the same thread, seems reasonable. This will also require corresponding changes to the module name.

marcbonnici:
Implemented with https://github.com/ARM-software/devlib/pull/454

PR #382, opened 2019-04-25
Add support for 'perf record and 'perf report' to devlib

markbellamy:


devlib/trace/perf.py
```diff
@@ -82,10 +86,20 @@ def __init__(self, target,
...
+            self.report_optionstrings = report_optionstring
+        else:
+            self.report_optionstrings = [report_optionstring]
         if not self.labels:
             self.labels = ['perf_{}'.format(i) for i in range(len(self.optionstrings))]
         if len(self.labels) != len(self.optionstrings):
             raise ValueError('The number of labels must match the number of optstrings provided for perf.')
+        if len(self.optionstrings) != len(self.report_optionstrings):
+            raise ValueError('The number of report_optionstrings must match the number of optionstrings provided for perf.')
+        if perf_mode == 'stat' or perf_mode == 'record':
```
setrofim: This, along with the `elif` clause can be converted to just

```
if perf_mode not in ['stat', 'record']:
    raise ValueError('Invalid perf_mode: must be either "stat" or "record"')
self.perf_mode = perf_mode
```

devlib/trace/perf.py
```diff
@@ -131,7 +163,25 @@ def _get_target_outfile(self, label):
...
+        return self.target.get_workpath('{}.rpt'.format(label))
+
+    def _get_target_tracefile(self, label):
+        return self.target.get_workpath('{}.trace'.format(label))
+
+    def _build_perf_report_command(self, reportoptions, label):
+        reportoptions_string = reportoptions
+        command = PERF_REPORT_COMMAND_TEMPLATE.format(self.binary,
+                                                      reportoptions_string,
+                                                      self._get_target_perfdatafile(),
```
marcbonnici: When testing this on my linux device the `perf.data` file is generated in the default login location due to the automatic generation rather than the workpath that this expects. If there is no way to control where this file is generated could I suggest explicitly changing to the desired output folder before running the record command to ensure that the output file is in the correct location?

devlib/trace/perf.py
```diff
@@ -131,7 +163,25 @@ def _get_target_outfile(self, label):
...
+                                                      self._get_target_reportfile(label))
```
marcbonnici: And when testing this on my android device the fails with error code 244 caused by `magic/endian check failed`. Have you seen this on your end?

devlib/trace/perf.py
markbellamy: On Android it seems to include a `cd` to the output folder as part of the command execution:
`2019-04-29 19:23:04,804 DEBUG       android:         adb -s 10.1.16.10:5555 shell echo 'cd sdcard/devlib-target && /data/local/tmp/bin/busybox nohup /data/local/tmp/bin/perf record -a -e migrations -e cs -e r11 sleep 1000 > /sdcard/devlib-target/perf_0.out 2>&1  &' | su`

I'd assume that it is the same on Linux but haven't confirmed. I can look at running on Linux target but will take me a little time. Do you happen to have the run.log file from your testing to confirm the command used on Linux?

devlib/trace/perf.py
markbellamy: I have seen error code 244 on a few occasions but haven't tracked it down. It has been working reliable for a while for me on the Android device I am using. I was guessing it was related to the `perf report` command being '<command>' | su rather than '<command> | su' used for `perf record` but that just a guess. I'll try it on a few devices to see if it reliable reproduces or not.

devlib/trace/perf.py
```diff
@@ -108,7 +124,23 @@ def stop(self):
...
+                target_file = self._get_target_reportfile(label)
+                host_relpath = os.path.basename(target_file)
+                host_file = _f(os.path.join(outdir, host_relpath))
+                self.target.pull(target_file, host_file)
+                # .trace
+                target_file = self._get_target_tracefile(label)
+                host_relpath = os.path.basename(target_file)
+                host_file = _f(os.path.join(outdir, host_relpath))
+                target_file = self._get_target_perfdatafile()
+                self.target.pull(target_file, host_file)
```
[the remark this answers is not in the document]
markbellamy: The file never exists on the device because a mixture of perf behavior and how it is executed by WA. WA executes perf to run sleep and then kills it after a certain time. This seems to break the perf -o option and perf just outputs to a default place/file name ({current directory}/perf.data). So the -o option isn't used to run perf record and we do a rename of perf.data. Seemed easiest just to do it as part of the copy rather than directly on the device.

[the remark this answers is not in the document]
pietos01: > [..] then kills it after a certain time. This seems to break the perf -o option [...]

Interesting, I didn't know that. It's probably because `perf` has buffered some data and, as `.kill()` calls `kill` which sends `SIGKILL`, `perf` can't handle the signal and, therefore, can't write back what it's holding to the file. Sending a signal it's allowed to handle would probably solve that; this is something to keep in mind. Thanks

devlib/trace/perf.py
```diff
@@ -69,6 +71,8 @@ def __init__(self, target,
                  events=None,
                  optionstring=None,
                  labels=None,
+                 perf_mode='stat',
+                 report_optionstring=None,
```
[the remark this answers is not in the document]
setrofim: Could you elaborate on what you mean (maybe raise a separate issue, if inappropriate for this PR)?

devlib/trace/perf.py
```diff
@@ -23,7 +24,8 @@
 from devlib.utils.misc import ensure_file_directory_exists as _f
 
 
-PERF_COMMAND_TEMPLATE = '{} stat {} {} sleep 1000 > {} 2>&1 '
+PERF_COMMAND_TEMPLATE = '{} {} {} {} sleep 1000 > {} 2>&1 '
```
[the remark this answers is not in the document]
setrofim: Basically, this is to save having to pull two files instead of one. Given that the output is generally parsed with a line parser looking for specific patterns, mixing in `stderr` lines usually don't impact things. The question is -- is there specific benefit to splitting `stdout` and `stderr` in this case?

>If we don't want that, isn't > {outfile} 2>&1 simply &> {outfile} or is there a subtle difference?

The `bash` man page specifically states they're semantically equivalent. I just end to write `2>&1` out if habit and because I keep forgetting that `&>` exists.

> Second question (tiny detail, out of curiosity), what is the final space in the string for?

Erm. Aesthetics? A subtle reference to the TBS [animated show](https://en.wikipedia.org/wiki/Final_Space)? \</joke> Seriously, I have no idea. It might be a hold-over form an earlier WA2 implementation where it might have mattered. But I don't see it serving a purpose in the current code. Well spotted.

devlib/trace/perf.py
```diff
@@ -108,7 +123,23 @@ def stop(self):
 
     # pylint: disable=arguments-differ
     def get_trace(self, outdir):
-        for label in self.labels:
+        for label, report_opt in zip(self.labels, self.report_optionstrings):
```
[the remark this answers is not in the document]
setrofim: Those are all excellent points. The "trace" interface hasn't been particularly well thought out and probably needs some serious though put into redesigning it. (And as has been pointed out elsewhere, "trace" is not a good name for what is being handled by this interface at the moment (see https://github.com/ARM-software/devlib/issues/378.)

This a much bigger issue that this PR, and I don't think it should block it. There is already precedent for  directories being passed to `get_trace` (e.g. `ftrace` implementation expects `outfile` to be either a directory or a file). I agree that the name `outfile` is confusing, and you're right, it's not documented -- that's on me.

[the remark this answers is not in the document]
setrofim: I have raised an issue to track this: https://github.com/ARM-software/devlib/issues/383. If you have any suggestions for changes to the collector API, please feel free to add them to the issue.

devlib/trace/perf.py
markbellamy: I've tested on a few Android devices and see error code 244 or just failures on some but not others. It seems to be consistent behavior on the devices, so a device either works or doesn't.

The error message seems to be related to some check of the perf.data file and likely finding an corruption issue with perf.data. So I think this is a result of the approach WA uses for running (aka killing) perf and some platforms behavior writing to the perf.data file to storage when killed. 

Apart from a re-write of perf support in WA I think the only thing that could be done here is to try identify the behavior that causes the file corruption and see if we can override it from the device command prompt (e.g. setprop ....)

devlib/trace/perf.py
marcbonnici: Ah yes I see the same behaviour on my android device however on linux it produces: 
```DEBUG           ssh:         (sh -c '/usr/bin/perf record -a -e migrations -e cs sleep 1000 > /home/root/devlib-target/perf_0.out 2>&1 ' 1>/dev/null 2>/dev/null &); __devlib_ec=$?; echo; echo $__devlib_ec```

For some reason the `kick_off` implementation on linux diffs to androids and does not have the initial `cd` command. I'm not sure why this would be the case and it seems like something that should be corrected to be consistent.

devlib/trace/perf.py
marcbonnici: That's interesting, I've tried running this again with the latest version and no longer see this problem on my device. 
And I apologies however I think I may have made an error before. In order to stop the perf instrument we're actually just killing the `sleep` command that perf is monitoring rather than killing `perf` itself in order to stop the measurements. The killing of `perf` only happens when the instrument is reset to ensure there are no extra processes running before running the workload and in the teardown stage. So I'm not sure this would explain this behaviour. 
I'm instead wondering if it could be down to the speed of the device and if the data hasn't finished being written to disk before we try to process it? Would it be possible to try adding a delay in between the two and then test this on a device that has this issue to see if that could be the case?

pietos01: I hadn't notice @marcbonnici had replied to this: could we try rebasing on #384 to see if the error message still appears?

markbellamy: I've tried #384  and sleep(10) and still seeing exit code 244.

MABonnici: Hmm... I'm wondering why I received this initially and haven't been able to reproduce it since.  
Would you be able to run the steps that WA is performing manually to determine if it is something that WA is doing to cause this or whether it's perf on the device itself that is having issues?

marcbonnici: Hmm... I'm wondering why I received this initially and haven't been able to reproduce it since.
Would you be able to run the steps that WA is performing manually to determine if it is something that WA is doing to cause this or whether it's perf on the device itself that is having issues?

pietos01: Just in case, could you give #386 a try?

devlib/trace/perf.py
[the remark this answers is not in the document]
pietos01: This might be the cause of the permission errors: `perf record` seems to generate a `perf.data` file with permissions `600` (owned by `root` as it's running through `su`) so that `self.target.pull` (which ends up doing an `adb pull` which is seen, from the target filesystem, as an access from a non-`root` user) would fail. The solution might be to use `self.target.pull(..., as_root=True)`?

PR #385, opened 2019-05-09
Exit context manager

douglas-raillard-arm:
Allow reverting a write to path by using `Target.write_value` as returning a context manager. This is useful to have more guarantees that the target is left in a clean state after the code has executed, even if some exceptions were raised (including user interruptions with `KeyboardInterrupt`).

This is 100% backward compatible, as `undo=False` by default.

### DIFF
--- a/doc/target.rst
+++ b/doc/target.rst
@@ -346,6 +346,18 @@ Target
        some sysfs entries silently failing to set the written value without
        returning an error code.
 
+.. method:: Target.revertable_write_value(path, value [, verify])
+
+   Same as :meth:`Target.write_value`, but as a context manager that will write
+   back the previous value on exit.
+
+.. method:: Target.batch_revertable_write_value(kwargs_list)
+
+   Calls :meth:`Target.revertable_write_value` with all the keyword arguments
+   dictionary given in the list. This is a convenience method to update
+   multiple files at once, leaving them in their original state on exit. If one
+   write fails, all the already-performed writes will be reverted as well.
+
 .. method:: Target.read_tree_values(path, depth=1, dictcls=dict, [, tar [, decode_unicode [, strip_null_char ]]]):
 
    Read values of all sysfs (or similar) file nodes under ``path``, traversing
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -29,6 +29,7 @@ import threading
 import xml.dom.minidom
 import copy
 from collections import namedtuple, defaultdict
+from contextlib import contextmanager
 from pipes import quote
 from past.builtins import long
 from past.types import basestring
@@ -51,6 +52,7 @@ from devlib.utils.android import AdbConnection, AndroidProperties, LogcatMonitor
 from devlib.utils.misc import memoized, isiterable, convert_new_lines
 from devlib.utils.misc import commonprefix, merge_lists
 from devlib.utils.misc import ABI_MAP, get_cpu_name, ranges_to_list
+from devlib.utils.misc import batch_contextmanager
 from devlib.utils.types import integer, boolean, bitmask, identifier, caseless_string, bytes_regex
 
 
@@ -69,7 +71,6 @@ GOOGLE_DNS_SERVER_ADDRESS = '8.8.8.8'
 
 
 installed_package_info = namedtuple('installed_package_info', 'apk_path package')
-
 
 class Target(object):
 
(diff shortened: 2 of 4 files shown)


douglas-raillard-arm:
@valschneider Any thoughts on that design pattern ?

setrofim:
I like the idea, however I'm not sure putting this into `write_value` is the best move. E.g. if you want to write and restore complex state, you might end up nesting a lot of contexts. Also, to me at least, the notion of adding a  parameter to conditionally turn a method into a context manager seems counterintuitive.

As an alternative, how about adding a separate `create_revertable_state()` (or some better name) method that takes a `dict` of `{path: value}`, and then writes and reverts them all at once (at least, as far as the calling code is concerned)?

douglas-raillard-arm:
> E.g. if you want to write and restore complex state, you might end up nesting a lot of contexts
Most of the time, these complex state also have their own dedicated context managers (like cpufreq in module), and they can be all specified on one line (separated by commas) if needed instead of nesting, which sometimes adds a bit of readability.

If many files need to be written to at the same time, the user should use the standard library's `contextlib.ExitStack`, which allows stacking together any sequence of context managers:
https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack
If we still want to have that convenience function, it seems cleaner to let `write_value` handle its piece, and then just stick `ExitStack` and `write_value` inside `create_revertable_state()`. It's almost always a better pattern to deal with one item at a time and then use some generic combinators rather than one "big" function that tries to deal with a whole bunch of items at once, whenever possible. In this case, it allows having a different `verify` value for each file, which becomes either impossible with `create_revertable_state()`, or adds complexity to the API (list of kwargs to pass to `write_value`), and if I need to manipulate one file only, I end up calling the "batch" function with just one item, which is quite clunky.

(Python 2 users can 1) hurry up and convert to Python3, EOL is coming soon and 2) use https://pypi.org/project/contextlib2/, which is a backport of Python 3's `contextlib` module)

> Also, to me at least, the notion of adding a parameter to conditionally turn a method into a context manager seems counterintuitive.
We could do without `undo` but that would limit usage of `exit_action_cm` to functions that did not return anything useful before, to avoid breaking any API. If we do it this way, `write_value` will behave like `open()`: used along `with` it "undoes" the action, used standalone it does not undo anything. Maybe it's the best way to go. 

note: In that case, `exit_action_cm` would mirror `contextlib.contextmanager`, but will always run the first part (above `yield`) so it can be used outside `with`. Now that I think about it, it's much nicer than the `lambda` wart currently needed.

setrofim:
OK, I take you point regarding `verify` -- I had not considered that.

>We could do without undo but that would limit usage of exit_action_cm to functions that did not return anything useful before, to avoid breaking any API. If we do it this way, write_value will behave like open(): used along with it "undoes" the action, used standalone it does not undo anything. Maybe it's the best way to go.

Yes, I would prefer this solution.

I still think a convenience function to handle multiple values as a single "state"  `dict` would be useful.

setrofim:
> EDIT: That would mean reading the original value everytime, which would slow it down and make it fail on some write-only files. The only way out is either undo arg, or a separate function. That said, it is orthogonal to having that function handle one or multiple files as discussed above.

I would prefer a separate function/method then.

douglas-raillard-arm:
@setrofim updated the PR as following:
* Added `revertable_write_value()`, which is a `with`-friendly version of `write_value()`
* Added `batch_contextmanager()` decorator, based on `ExitStack`. To support that in Python2, added `contextlib2` dependency in setup.py.
* Added `batch_revertable_write_value()`, implemented using `batch_contextmanager()`, which takes a list of `revertable_write_value()` kwargs.

If we don't want to depend on `contextlib2`, it should relatively easy to implement a reduced `ExitStack` with the subset of its API we need (namely the `enter_context()` method).

doc/target.rst
```diff
@@ -346,6 +346,18 @@ Target
        some sysfs entries silently failing to set the written value without
        returning an error code.
 
+.. method:: Target.revertable_write_value(path, value [, verify])
+
+   Same as :meth:`Target.write_value`, but as a context manager that will write
+   back the previous value on exit.
+
+.. method:: Target.revertable_write_value(kwargs_list)
```
marcbonnici: This should be `Target.batch_revertable_write_value(kwargs_list)`

valschneider:
> @valschneider Any thoughts on that design pattern ?

Took me a while to get there, but I quite like it. It's similar in concept to https://github.com/ARM-software/devlib/blob/master/devlib/module/cpufreq.py#L101.

It would be nice to be able to use these new methods in the above, but we need to call `set_governor()` which uses `write_value()`, so I think we'll be stuck with a separate implementation here.

douglas-raillard-arm:
> It would be nice to be able to use these new methods in the above, but we need to call set_governor() which uses write_value(), so I think we'll be stuck with a separate implementation here.

If we want to easily add this kind of capabilities, it could be possible to have an API like that:
```
with target.enter_revertable_writes() as revertable_target:
    revertable_target.write_value(foo, bar)
    ...
# All the `write_value()` that happened inside the block are undone here
```

`revertable_target` here is just a proxy for the real target, that wraps `write_value()` to maintain an `ExitStack` which is used in `__exit__`.

douglas-raillard-arm:
@setrofim Fixed the typo in the documentation of `batch_revertable_write()`

I'll remove the RFC status as I consider it ready for merge on my side.

Issue #386, opened 2019-05-13
Android tagets: Piped su call vs --command option


PiCT0:
This is with (Google Pixel 2):
```
walleye:/ # su --version
15.3:MAGISKSU (topjohnwu)
```
and (host):
```
$ adb --version
Android Debug Bridge version 1.0.40
Version 28.0.2-5303910
```
---
The way `devlib.AndroidTarget.execute` works relies on `su` and there seems to be a difference in behaviour between a call to that method and actually running a command inside of `su` inside of an `adb shell`.

I am unsure about the reason but, as an example (the device is connected through `adb` and there exists a `/sdcard/devlib-target/perf.data` file on the device):
```Python
Python 3.6.7 (default, Oct 22 2018, 11:32:17)
>>> import devlib
>>> t = devlib.AndroidTarget()
>>> print(t.execute('cd /sdcard/devlib-target && /data/local/tmp/bin/perf report', as_root=True))
incompatible file format (rerun with -v to learn more)Warning:
Kernel address maps (/proc/{kallsyms,modules}) were restricted.

Check /proc/sys/kernel/kptr_restrict before running 'perf record'.

As no suitable kallsyms nor vmlinux was found, kernel samples
can't be resolved.

Samples in kernel modules can't be resolved as well.

Error:
The - file has no samples!
# ========
# captured on: Mon May 13 18:35:09 2019
# ========
#
```
while
```
$ adb shell
walleye:/ $ su
walleye:/ # cd /sdcard/devlib-target
walleye:/sdcard/devlib-target # /data/local/tmp/bin/perf report
```
gives the expected result.

In fact, going through `adb shell` with oneliners:
```
$ adb shell "echo 'cd /sdcard/devlib-target && /data/local/tmp/bin/perf report' | su"
incompatible file format (rerun with -v to learn more)Warning:
Kernel address maps (/proc/{kallsyms,modules}) were restricted.

Check /proc/sys/kernel/kptr_restrict before running 'perf record'.

As no suitable kallsyms nor vmlinux was found, kernel samples
can't be resolved.

Samples in kernel modules can't be resolved as well.

Error:
The - file has no samples!
# ========
# captured on: Mon May 13 18:22:58 2019
# ========
#
```
while the following works:
```
$ adb shell "su -c 'cd /sdcard/devlib-target && /data/local/tmp/bin/perf report'"
```

[...]

PR #387, opened 2019-05-14
Improve adb_shell and move to 'su -c'

pietos01:
Slight rework of `adb_shell` to improve readability and stability but, most importantly, to address #386.

Please note that this requires modern Py3 and should be merged post-WA3.1 (preferred) or could be modified to be Py2.7-compatible.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -431,7 +431,7 @@ def adb_shell(device, command, timeout=None, check_exit_code=False,
     if device is not None:
         parts += ['-s', device]
     parts += ['shell',
-              command if not as_root else 'echo {} | su'.format(quote(command))]
+              command if not as_root else 'su -c {}'.format(quote(command))]
 
     logger.debug(' '.join(quote(part) for part in parts))
     # On older combinations of ADB/Android versions, the adb host command always
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -31,7 +31,10 @@ import pexpect
 import xml.etree.ElementTree
 import zipfile
 
-from pipes import quote
+try:
+    from shlex import quote
+except ImportError:
+    from pipes import quote
 
 from devlib.exception import TargetTransientError, TargetStableError, HostError
 from devlib.utils.misc import check_output, which, ABI_MAP
@@ -422,23 +425,23 @@ def _ping(device):
 def adb_shell(device, command, timeout=None, check_exit_code=False,
               as_root=False, adb_server=None):  # NOQA
     _check_env()
-    if as_root:
-        command = 'echo {} | su'.format(quote(command))
-    device_part = []
-    if adb_server:
-        device_part = ['-H', adb_server]
-    device_part += ['-s', device] if device else []
+    parts = ['adb']
+    if adb_server is not None:
+        parts += ['-H', adb_server]
+    if device is not None:
+        parts += ['-s', device]
+    parts += ['shell',
+              command if not as_root else 'echo {} | su'.format(quote(command))]
 
+    logger.debug(' '.join(quote(part) for part in parts))
     # On older combinations of ADB/Android versions, the adb host command always
     # exits with 0 if it was able to run the command on the target, even if the
     # command failed (https://code.google.com/p/android/issues/detail?id=3254).
     # Homogenise this behaviour by running the command then echoing the exit
     # code.
-    adb_shell_command = '({}); echo \"\n$?\"'.format(command)
-    actual_command = ['adb'] + device_part + ['shell', adb_shell_command]
-    logger.debug('adb {} shell {}'.format(' '.join(device_part), command))
+    parts[-1] += ' ; echo "\n$?"'
     try:
-        raw_output, _ = check_output(actual_command, timeout, shell=False, combined_output=True)
+        raw_output, _ = check_output(parts, timeout, shell=False, combined_output=True)
     except subprocess.CalledProcessError as e:
         raise TargetStableError(str(e))
 


devlib/utils/android.py
```diff
@@ -31,7 +31,7 @@
 import xml.etree.ElementTree
 import zipfile
 
-from pipes import quote
+from shlex import quote
```
pietos01: `pipes.quote` is hidden and shouldn't be used; [`shlex.quote`](https://docs.python.org/3/library/shlex.html?highlight=shlex#shlex.quote) was added in Py3.3 for this purpose.

devlib/utils/android.py
```diff
@@ -422,23 +422,23 @@ def _ping(device):
...
-        device_part = ['-H', adb_server]
-    device_part += ['-s', device] if device else []
-
+    parts = ['adb']
+    if adb_server is not None:
+        parts += ['-H', adb_server]
+    if device is not None:
+        parts += ['-s', device]
+    parts += ['shell',
+              command if not as_root else f'su -c {quote(command)}']
```
pietos01: f-strings are introduced in Py3.6.

devlib/utils/android.py
```diff
@@ -422,23 +422,23 @@ def _ping(device):
...
     # command failed (https://code.google.com/p/android/issues/detail?id=3254).
     # Homogenise this behaviour by running the command then echoing the exit
     # code.
-    adb_shell_command = '({}); echo \"\n$?\"'.format(command)
-    actual_command = ['adb'] + device_part + ['shell', adb_shell_command]
-    logger.debug('adb {} shell {}'.format(' '.join(device_part), command))
+    parts[-1] += r' ; echo "\n$?"'
     try:
-        raw_output, _ = check_output(actual_command, timeout, shell=False, combined_output=True)
+        raw_output, _ = check_output(parts, timeout, shell=False, combined_output=True)
```
pietos01: About `check_output`: `timeout` was added to most `subprocess` functions, including [`subprocess.check_output`](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) in Py3.3. Furthermore, the `universal_newlines` (or the alias `text` in Py3.7) exists to avoid having to do what follows with the
```Python
raw_output.replace('\r\n', '\n').replace('\r', '\n')
```

devlib/utils/android.py
```diff
@@ -422,23 +422,23 @@ def _ping(device):
...
-
+    parts = ['adb']
+    if adb_server is not None:
+        parts += ['-H', adb_server]
+    if device is not None:
+        parts += ['-s', device]
+    parts += ['shell',
+              command if not as_root else f'su -c {quote(command)}']
+
+    logger.debug(' '.join(quote(part) for part in parts))
```
pietos01: Quoted logs!

devlib/utils/android.py
```diff
@@ -422,23 +422,23 @@ def _ping(device):
 def adb_shell(device, command, timeout=None, check_exit_code=False,
               as_root=False, adb_server=None):  # NOQA
     _check_env()
-    if as_root:
-        command = 'echo {} | su'.format(quote(command))
-    device_part = []
-    if adb_server:
```
pietos01: Truthiness doesn't seem consistent with the variable definition.

devlib/utils/android.py
setrofim: Note: we're currently still supporting Python 2.7;  please add an import guard with fallback to importing from `pipes`.

devlib/utils/android.py
setrofim: We're still aiming to be compatible with Python 2.7.

pietos01:
From #386 ([comment](https://github.com/ARM-software/devlib/issues/386#issuecomment-492173869)):
> I can't think of a reason other than that for the pipe, and would be happy to make the switching to using '-c' . Though it would be have to be thoroughly tested with WA/LISA before merging.

How should this testing be done?

setrofim:
We can handle the testing for WA. For LISA,  I believe there is a way to get it to point to an alternative branch/repo for devlib  -- so just switch it over and make sure that everything still works. If it's been used for a while and nothing breaks, then we're probably good. Obviously, if LISA  had it's own test suite, make sure that passes too.

valschneider:
I'll pick that up and feed it through our device tests. Will update you folks here.

@douglas-raillard-arm FYI.

devlib/utils/android.py
pietos01: Indeed, under the hood of CPython, [`pipes.quote` is `shlex.quote`](https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Lib/pipes.py#L65) but [it's messy even for them](https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Lib/pipes.py#L64). Using `shlex.quote` is incontestably the cleanest way (as that's what the Standard says) to get the `quote` function (in Py3.3+) but I agree with you that using `pipes.quote` _works_ (until it doesn't?).

devlib/utils/android.py
```diff
@@ -422,23 +422,23 @@ def _ping(device):
...
-        device_part = ['-H', adb_server]
-    device_part += ['-s', device] if device else []
-
+    parts = ['adb']
+    if adb_server is not None:
+        parts += ['-H', adb_server]
+    if device is not None:
+        parts += ['-s', device]
+    parts += ['shell',
+              command if not as_root else f'echo {quote(command)} | su']
```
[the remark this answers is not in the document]
pietos01: I was unable to find documentation stating which (sub-)version of Python 3 devlib was supporting (I might have missed a document or webpage?) so I assumed Py3.6.

You're right that devlib is still supporting Py3.5 but, now that the projects (incl. WA) are slowly moving away from Python 2, it is necessary to define a "minimum" version of Python 3 to be supported. I though Py3.6 made sense as it's fairly common, relatively stable ([2.5 years old; bug-fix support already ended at the end of last year](https://www.python.org/downloads/release/python-360/)) and provides some valuable new features.

The f-strings have been removed from this PR but it is still necessary to define which version of Python 3 is/will be used by the projects; @setrofim & @marcbonnici ?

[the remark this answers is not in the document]
marcbonnici: So for this release we will continue to be targeting older version of 3 as well as 2.7. However as @pietos01 mentioned for both WA and devlib, python 2.7 is now deprecated and support will be dropped in the next release. At the point we are enforcing a switch to python 3 we were also thinking of setting Python 3.6 as our new required version as this appears to have a large adoption and introduces a number of worthwhile changes.
I realise this has not been communicated properly so we will need to formalise this information and  get feedback on our plans. @douglas-raillard-arm Do you foresee any issues with this?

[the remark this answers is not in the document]
douglas-raillard-arm: No worries, it eventually reaches involved people :)

So, in the current LISA landscape:
* Travis' latest GNU/Linux environment is Ubuntu 16,04 xenial which ships with 3.5. Getting higher versions may be possible via some PPA. Travis can also install multiple version of python if we tell it it's a python project, but we can't use that since it also comes with its own (non-optional) virtualenv which conflicts with the one from LISA. (we could configure LISA to cooperate, but we want to check end-to-end install as well in a way that matches what users will do).
* Some LAVA containers/Jenkins environment we use is using Ubuntu Xenial as well. That may be updated, but I would need to have a chat with the infra guys before committing on that. @ionela-voinescu any comment on that ?
* Some people in our team may be using Ubuntu Xenial, I assume this can be extended to other teams, since it will be supported until 2021/2024. 

All in all, we might be able to require Python >=3.6 but that means asking people to use PPA. Do you have some kind of informal list of things that are desirable in 3.6 and not in 3.5 ? (formatted literals are definitely desirable :-)  but may not be big enough to warrant dropping native xenial support :-( ).

devlib/utils/android.py
douglas-raillard-arm: I guess I was just lazy and did not introduce an except block for something that appeared to work in Python 3 as well, using shlex when available is definitely less confusing :)

devlib/utils/android.py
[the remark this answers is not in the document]
marcbonnici: Ok thank you very much for the information, that's definitely something we'll need to take into account then. 
We don't have such a list yet so we will need to evaluate this. In order to collate this information I've opened the issue https://github.com/ARM-software/devlib/issues/389 to discuss this further.

marcbonnici:
@valschneider Just wondering if you'd had a chance to run this through your tests?

valschneider:
I think this got somewhat lost in the mayhem caused by our Travis issues, let me check with @douglas-raillard-arm.

marcbonnici:
@pietos01 Would you be able to squash the relevant commits and update the commit messages to reflect it is no longer Python3 only?
We'd like to merge this shortly once @valschneider's tests are done.

PR #388, opened 2019-05-15
Add generic perf trace instrument

pietos01:
As discussed in #382, this PR modifies the `perf` trace instrument so that it accepts any `perf` command. It implements the back-end of ARM-software/workload-automation#979 and requires fix #387 to avoid #386. It is arguably simple as it receives the commands to run as strings (prepending the path to `perf`) and runs them "at the right time".

Based on the  subcommands `perf` currently has, it looks like these can be placed in 3 categories:
- The "measuring" commands (`commands`): used to trace/measure some system properties (`stat`, `record`, `mem`, `sched`, `kmem`, `kvm`, `c2c`, ...);
- The "reporting" commands (`post_commands`): used to process a trace file from a "measuring" command (`report`, `script`, `timechart`, `annotate, `c2c report`, ...);
- The "information" and "configuration" commands (`pre_commands`): used to configure the system or get general information (`list`, `config`, `buildid-list`, `help`, ...)

Therefore, we take 3 dictionaries of commands that are run in order (`pre_`, then `commands`, then `post_`) in accord with these definitions. I'm still not perfectly sure about "when" in the WA/devlib process these are to be run, though. Please check that.

The reason why we have dictionaries for commands (instead of lists) is that the keys can act as a label which seemed to be a desired feature of the previous implementation. As we have multiple dictionaries, a same label can appear multiple times (_i.e._ as a key in more than a single dictionary). In that case, it makes sense to "group" these commands in their own "execution lane" _e.g._ a `record` and its corresponding `report`. Maybe, these labels could be used as directory names on the target and then on the host, once having been extracted(?) For example, commands within the same label would "see" the same files (as they would be in the same subfolder) while commands with different labels wouldn't and would therefore be able to re-use the same filenames. A similar approach with respect to the filename would be to prepend/inject/append the label to the filename.

The `Command` class is implemented to be used by the corresponding WA `perf` instrument when generating the `perf` commands passed to the devlib `PerfCollector`. I thought it would make sense to have that class in devlib instead of WA. See the front-end PR for a description of the constructor parameters.

The `get_trace` hasn't been implemented yet. If we're going with a function that extracts one `outfile` at a time, the caller (WA) will have to keep track of which files are generated by which commands (_e.g._ the `stdout`, `stderr`, `perf.data`, ...) which is hard to do. Another approach might be to extract a pull a complete folder from the target (_e.g._ `/sdcard/devlib-target/perf`) in which `perf` would have run. This decision is also influenced by whether labels are used as subfolders or not.

Parsing of the output(s) is the responsibility of WA and won't be done here.

Please note that the proposed implementation is *not* an in-place replacement for the previous implementation of `PerfCollector`. Therefore, the previous implementation has been moved to `perf_stat`, should be deprecated (not sure what the best way to do that is) and should eventually be removed.

### TODO:

- [x] Check the location of execution of `pre_`, `commands`, `post_` in the WA/devlib flow;
- [x] ~~Deprecate `PerfStatCollector`~~
- [x] Decide how labels are used on the device
- [x] Document `PerfCollector` (what is expected?)
- [x] Trace extraction (per file? ...)

### DIFF
--- a/devlib/trace/perf.py
+++ b/devlib/trace/perf.py
@@ -46,12 +46,6 @@ class PerfCommandDict(collections.OrderedDict):
             if 'stat'in parameters['command']:
                 self._stat_command_labels.add(label)
 
-    def stat_commands(self):
-        return {label: self[label] for label in self._stat_command_labels}
-
-    def as_strings(self):
-        return {label: str(cmd) for label, cmd in self.items()}
-
 
 class PerfCollector(TraceCollector):
     """Perf is a Linux profiling tool based on performance counters.
--- a/devlib/trace/perf.py
+++ b/devlib/trace/perf.py
@@ -109,17 +109,21 @@ class PerfCollector(TraceCollector):
         for label, command in self.post_commands.items():
             self.execute(str(command), label)
 
+    def _target_runnable_command(self, command, label=None):
+        cmd = '{} {}'.format(self.binary, command)
+        if label is None:
+            return cmd
+        directory = quote(self.working_directory(label))
+        cwd = 'mkdir -p {0} && cd {0}'.format(directory)
+        return '{cwd} && {cmd}'.format(cwd=cwd, cmd=cmd)
+
     def kick_off(self, command, label=None):
-        directory = quote(self.working_directory(label or 'default'))
-        return self.target.kick_off('mkdir -p {0} && cd {0} && {1} {2}'
-                                    .format(directory, self.binary, command),
-                                    as_root=self.target.is_rooted)
+        cmd = self._target_runnable_command(command, label)
+        return self.target.kick_off(cmd, as_root=self.target.is_rooted)
 
     def execute(self, command, label=None):
-        directory = quote(self.working_directory(label or 'default'))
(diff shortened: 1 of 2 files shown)


devlib/trace/perf.py
```diff
@@ -11,131 +11,128 @@
...
 
+from devlib.utils.cli import Command
 from devlib.host import PACKAGE_BIN_DIRECTORY
 from devlib.trace import TraceCollector
-from devlib.utils.misc import ensure_file_directory_exists as _f
 
+if sys.version_info >= (3, 0):
+    from shlex import quote
+else:
+    from pipes import quote
```
marcbonnici: This implies that compatibility with Python2.7 is attempted to be retained however is not present elsewhere in this PR. Is this aimed to be merged before the next release aka supporting Python 2.7 or to be held until after support has been dropped?

pietos01:
See the [comment](https://github.com/ARM-software/workload-automation/pull/979#issuecomment-509550190) on ARM-software/workload-automation#979.

PR #393, opened 2019-06-05
module/thermal: Gracefully handle unexpected sysfs names

douglas-raillard-arm:
Instead of raising an exception, log an error and carry on.


devlib/module/thermal.py
```diff
@@ -88,6 +88,9 @@ def __init__(self, target):
 
         for entry in target.list_directory(self.thermal_root):
             re_match = re.match('^(thermal_zone|cooling_device)([0-9]+)', entry)
+            if not re_match:
+                logging.warn('unknown thermal entry: %s', entry)
```
setrofim: Use `self.logger.warning` instead of `logging.warn`

douglas-raillard-arm:
@setrofim  Updated with `self.logger.warning`

PR #397, opened 2019-07-06
module/sched: Return the correct maximum capacity

joshuous:
The existing behaviour assumes that the cap_states file contains a list
of capacity|cost pairs, and attempts to return the maximum capacity by
selecting the value at the second last index of the list.

This assumption fails on some newer Qualcomm kernels where the
cap_states file contains a list of capacity|frequency|cost triplets.
Consequently, the maximum frequency would be erroneously returned
instead of the maximum capacity.

Fix the problem by dynamically calculating the index of the maximum
capacity by dividing the number of entries in cap_states by the value in
nr_cap_states.

---

For example, on a Snapdragon 855 device:

/proc/sys/kernel/sched_domain/cpu0/domain0/group0/energy/cap_states
        54 entries:

        CAP     FREQ    COST
        --------------------
        65	300000	12
        87	403200	17
        104	480000	21
        125	576000	27
        141	652800	31
        162	748800	37
        179     825600	42
        195	902400	47
        212	979200	52
        228	1056000	57
        245	1132800	62
        266	1228800	70
        286	1324800 78
        307	1420800	89
        328	1516800	103
        348	1612800	122
        365	1689600	141
        381	1766400	160

/proc/sys/kernel/sched_domain/cpu0/domain0/group0/energy/nr_cap_states
        18

Max capacity = 381 (third-last index)

### DIFF
--- a/devlib/module/sched.py
+++ b/devlib/module/sched.py
@@ -415,7 +415,10 @@ class SchedModule(Module):
             sd = SchedProcFSData(self.target, cpu)
 
         cap_states = sd.domains[0].groups[0].energy.cap_states
-        return int(cap_states.split('\t')[-2])
+        cap_states_list = cap_states.split('\t')
+        num_cap_states = sd.domains[0].groups[0].energy.nr_cap_states
+        max_cap_index = -1 * int(len(cap_states_list) / num_cap_states)
+        return int(cap_states_list[max_cap_index])
 
     @memoized
     def get_dmips_capacity(self, cpu):


valschneider:
Looks good to me, I'll give it a quick test later on today on two different devices just to make sure.

joshuous:
I've been testing on the latest Lisa. The unscaled_utilization() method in base.py incorrectly uses the max freq value without this patch. Consequently incorrect values (outside the range of 0...100) will be set for for start_pct and end_pct in eas_behaviour.py, which causes several tests to break.

valschneider:
> I've been testing on the latest Lisa. The unscaled_utilization() method in base.py incorrectly uses the max freq value without this patch. Consequently incorrect values (outside the range of 0...100) will be set for for start_pct and end_pct in eas_behaviour.py, which causes several tests to break.

Yeah it's definitely broken there - I meant that we fixed it for the full energy model scrub (which isn't used to the rtapp profile generation of those tests), and we did it in a similar way to this fix.

joshuous:
Ah yes understood

valschneider:
Seems to be working fine, thanks again for the PR. While validating this I found some very stupid lines in that module written by yours truly... I'll be sending another PR sometime soon.

joshuous:
Gladly! :)

Issue #406, opened 2019-07-29
AndroidTarget: adb_root() needed before module loading


douglas-raillard-arm:
the `AndroidTarget.adb_root()` method can only be called after the target object is created. However, it may be needed for it to be called before attempting to load the devlib modules, since some of them need root permissions.

Adding an option in AndroidTarget connection settings or constructor to enable rooting would solve that, or allow loading the modules manually, so it could be done after `connect()`. Maybe "user='root'" connection setting could be used as a hint as it seems unused ?

PR #413, opened 2019-07-30
Allow for restarting adb as root upon connection. 

marcbonnici:
As per https://github.com/ARM-software/devlib/issues/406 currently in order to enable adbd as root, a connection to the target already needs to exist however there are use case that we want to perform initialising as root as well.

- Add a utility function to restarted adbd in rooted and unrooted mode.
- Add the `adb_as_root` connection parameter to `AdbConnection`.

Note: This will restart adbd as root before attempting to connect to the target and will also restart adbd unrooted upon the last disconnection from the target.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -113,15 +113,15 @@ class Target(object):
         return self._connected_as_root
 
     @property
-    @memoized
     def is_rooted(self):
-        if self.connected_as_root:
-            return True
-        try:
-            self.execute('ls /', timeout=5, as_root=True)
-            return True
-        except (TargetStableError, TimeoutError):
-            return False
+        if self._is_rooted is None:
+            try:
+                self.execute('ls /', timeout=5, as_root=True)
+                self._is_rooted = True
+            except (TargetStableError, TimeoutError):
+                self._is_rooted = False
+
+        return self._is_rooted or self.connected_as_root
 
     @property
     @memoized
@@ -214,6 +214,7 @@ class Target(object):
                  is_container=False
                  ):
         self._connected_as_root = None
+        self._is_rooted = None
         self.connection_settings = connection_settings or {}
         # Set self.platform: either it's given directly (by platform argument)
         # or it's given in the connection_settings argument
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1481,14 +1481,11 @@ class AndroidTarget(Target):
         adb_command(self.adb_name, 'reboot-bootloader', timeout)
 
     def adb_root(self, enable=True, force=False):
-        if enable:
-            if self._connected_as_root and not force:
-                return
-            adb_command(self.adb_name, 'root', timeout=30)
-            self._connected_as_root = True
+        if not isinstance(self.conn, AdbConnection):
+            raise TargetStableError('Cannot enable adb root without adb connection')
+        if enable and self.connected_as_root and not force:
             return
-        adb_command(self.adb_name, 'unroot', timeout=30)
-        self._connected_as_root = False
+        self.conn.adb_root(self.adb_name, enable=enable)
 
     def is_screen_on(self):
         output = self.execute('dumpsys power')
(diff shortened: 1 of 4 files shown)


douglas-raillard-arm:
When working on #414 two things occurred to me:
* `AndroidTarget.adb_root()` records the fact that the target has been rooted or not in `_connected_as_root` attribute. AFAICS that attribute would now "lie" since that is now up to the connection to decide what to do. That could be solved easily by making `_connected_as_root` a property that delegates the knowledge to the connection object.

* The fact that the target is rooted is fundamentally a target property AFAIK. There is at most one adb server per device, and there is a 1-to-1 mapping between actual devices and devlib.Target instances. OTH, `Target` seems to somehow support multithreading by opening a separate connection in each thread using it, so we might end up with multiple connection objects. That means that once moved to the connection object, `_connected_as_root` would need to be kept in sync by `Target`, which indicates it should stay in `Target`, and that `Target` should be responsible for rooting the device.

marcbonnici:
> AFAICS that attribute would now "lie" since that is now up to the connection to decide what to do. 

That is a very good point. To prevent this and avoid need needing to implement the logic behind tracking the connected state across connections, it seems to make sense to just remove the caching behind the `connected_as_root` property and let the root status be queried as needed assuming this is property that would not be accessed too frequently to cause a performance impact.  

> That means that once moved to the connection object, _connected_as_root would need to be kept in sync by Target, which indicates it should stay in Target, and that Target should be responsible for rooting the device.

If we go with the approach stated above for determining `connected_as_root` as a target property then this would prevent the requirement to keep the status synchronised like you said, however as for the responsibility of rooting the device, a `Target` is independent of it's connection type and the method of rooting the device for that particular connection (e.g. `adb` vs `ssh`) would vary, so I think that the responsibility would make more sense to be part of the connection class itself.

douglas-raillard-arm:
> it seems to make sense to just remove the caching behind the `connected_as_root` property

`is_rooted` is now used in a few places in LISA to change the behavior of client code when some optional-but-good-to-have operations are not possible. 

Here is a suggestion: If all connection types support a kind of "rooting" concept, then all connections could gain an `is_rooted` property that is not cached (i.e. it will always poll the device, so it cannot go out of sync wrt other active connections). The `Target` class could just read that property after creating the connection, cache it and re-expose it as `Target.is_rooted`. `AndroidTarget.adb_root()` would update that cached attribute.

> as for the responsibility of rooting the device, a Target is independent of it's connection type and the method of rooting the device for that particular connection (e.g. adb vs ssh) would vary

Good point.

marcbonnici:
> The Target class could just read that property after creating the connection, cache it and re-expose it as Target.is_rooted. AndroidTarget.adb_root() would update that cached attribute.

I think there should be a distinction between `connected_as_root` and the `is_rooted` property. `is_rooted` is currently a property of a `Target` which indicates whether a target itself is able to provide root access in some way, (e.g. being connected directly as root or via `su` etc.) and that is independent of the connection and should not change over the lifetime of a `Target`, unlike `connected_as_root` so I think these two properties should stay unrelated in that sense?

douglas-raillard-arm:
Makes sense, so let me rephrase my previous suggestion as:

* un-cache `Target.connected_as_root`, so it's always reflecting reality.
* cache `Target.is_rooted`, and invalidate the cache from `AndroidTarget.adb_root()`.

This means that the responsibility for connecting as root is entirely left to the connection, but the responsibility for exposing the `connected_as_root` status to the user is left to the target, using the current `execute("id")` method. Since `is_rooted` can actually change for `AndroidTarget` because of `adb_root()`, its caching is manually managed instead of using `memoized()`.

marcbonnici:
When attempting to implement this a few potential issues were uncovered so as a result this update has un-memoized `is_rooted` and implemented its own caching which also relys on the `connected_as_root` property which is now left to the connection to track and cache as appropriate per connection. 

The adb connection root status is cached manually and is updated / invalided either by calling  `adb_root`, which is now a method of the `AdbConnection` also used by `AndroidTarget` for compatibility, or from target methods that reset/reboot the device. 
This allows for the state of multiple connections to the same android target to share the rooted state while not affecting connection to other devices.

I think this should allow for everything to reflect an up to date state of the connection while avoiding unnecessary calls to the target for each call.

douglas-raillard-arm:
Looks like it would solve the issue, thanks for looking into that @setrofim :)

PR #414, opened 2019-07-30
target: android: Run adb root early

### LINKED ISSUE #406: AndroidTarget: adb_root() needed before module loading

douglas-raillard-arm:
Run adb root in AndroidTarget before calling super().connect() when username is
"root" since some devlib modules will require root permissions to initialize.

This introduces a "username" connection setting for Android that is mostly
ignored, but taken as a hint to run adb_root().

Fix #406

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1190,6 +1190,10 @@ class AndroidTarget(Target):
             # indefinitely when making calls to the device. To avoid this,
             # always disconnect first.
             adb_disconnect(device)
+
+        if self.connection_settings.get('username') == 'root':
+            self.adb_root(device=self.connection_settings.get('device', ''))
+
         super(AndroidTarget, self).connect(timeout=timeout, check_boot_completed=check_boot_completed)
 
     def kick_off(self, command, as_root=None):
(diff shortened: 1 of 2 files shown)


douglas-raillard-arm:
Looks like it is some kind of duplicate of #413

setrofim:
> Looks like it is some kind of duplicate of #413

Yes, #413 solves the same issue, in a similar way. 

>This introduces a "username" connection setting for Android that is mostly
ignored, but taken as a hint to run adb_root().

A more explicit parameter (e.g. `adb_as_root` as in #413 ) would be more appropriate than a `username` that is ignored except for one specific value.

douglas-raillard-arm:
Closing as #413 is now merged

PR #417, opened 2019-09-02
module/cpufreq: Make use_governor() affect only online CPUs

valschneider:
Turns out you can't change cpufreq attributes on an offlined
CPU (no big surprise!), so use_governor() will fail if a whole
frequency domain has been hotplugged out.

Change the default behaviour to only target online CPUs.


valschneider:
I just realized the commit header is a bit ambiguous - this change only applies to the default behaviour where no CPU is specified. Users are still free to specify hotplugged CPUs, and they'll get an exception for it.

PR #419, opened 2019-09-03
Set LC_ALL

douglas-raillard-arm:
Set LC_ALL=C by default before executing commands, so that commands have a predictable output.

Also some small fixes on PATH env var.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -396,7 +396,7 @@ class Target(object):
 
         # Ensure to use deployed command when availables
         if self.executables_directory:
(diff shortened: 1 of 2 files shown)


marcbonnici:
I remember we had some issues with encodings on the host side when when dealing with non ascii output, would this introduce additional encoding problems on the target?

setrofim:
IIRC, the issues we had were to do with host-side not being utf-8, and encountering chars it couldn't decode. I don't _think_ this would impact things, as we're now enforcing utf-8 host-side. `C` is the  the simplest locale, supporting single-byte chars, so host should be able to handle target output generated with with locale.

PR #420, opened 2019-09-03
trace: dmesg: Add KernelLogEntry.from_dmesg_output() classmethod

douglas-raillard-arm:
Allow building a list of KernelLogEntry from a full dmesg output, in addition to
building just one entry using KernelLogEntry.from_str() .

### DIFF
--- a/devlib/trace/dmesg.py
+++ b/devlib/trace/dmesg.py
@@ -98,6 +98,19 @@ class KernelLogEntry(object):
             msg=msg.strip(),
         )
 
+    @classmethod
+    def from_dmesg_output(cls, dmesg_out):
+        """
+        Return a generator of :class:`KernelLogEntry` for each line of the
+        output of dmesg command.
+
+        .. note:: The same restrictions on the dmesg output format as for
+            :meth:`from_str` apply.
+        """
+        for line in dmesg_out.splitlines():
+            if line.strip():
+                yield cls.from_str(line)
+
     def __str__(self):
         facility = self.facility + ': ' if self.facility else ''
         return '{facility}{level}: [{timestamp}] {msg}'.format(
@@ -156,18 +169,7 @@ class DmesgCollector(TraceCollector):
 
     @property
     def entries(self):
-        return self._parse_entries(self.dmesg_out)
-
-    @classmethod
-    def _parse_entries(cls, dmesg_out):
-        if not dmesg_out:
-            return []
-        else:
-            return [
-                KernelLogEntry.from_str(line)
-                for line in dmesg_out.splitlines()
-                if line.strip()
-            ]
+        return KernelLogEntry.from_dmesg_output(self.dmesg_out)
 
     def reset(self):
         self.dmesg_out = None


devlib/trace/dmesg.py
```diff
@@ -98,6 +98,23 @@ def parse_timestamp_msg(line):
...
+            :meth:`from_str` apply.
+        """
+        if not dmesg_out:
+            return []
+        else:
+            return [
+                cls.from_str(line)
+                for line in dmesg_out.splitlines()
+                if line.strip()
+            ]
```
setrofim: Do we want this to return a list or an iterator?

douglas-raillard-arm: I usually apply these criteria to decide for generator vs list:
* dmesg_out is a string, so it has to fit in memory. Parsing gigabytes of dmesg output would already not work, so we don't benefit from constant memory consumption brought by laziness.
* Iterating over the generator is slower than building the list and consuming it in my experience. * Generators are somehow annoying to use, since they can be iterated over only once. Half of the time, people will end up putting that in a list anyway to pass it around.

The only thing a generator would provide is allowing to process entries up to a line that fails to parse. In the current context, I don't see any use case that would need that property.

setrofim: > so we don't benefit from constant memory consumption brought by laziness.

In some cases, we still do. In the case of the `entries` property at least, the string is going to remain in memory regardless, as it is an attribute of the `DmesgCollector` so in creating the list, you're effectively doubling memory usage.

> Iterating over the generator is slower than building the list and consuming it in my experience.

I haven't  observed this, but am willing to believe that this is the case. However, in cases such as this, most of the time will sent actually parsing the log entries (especially since regex's are involved), so this is negligible. If anything, using a generator may result in a significant speed up, if you routinely break out early after finding what you need (see last point).

>  Generators are somehow annoying to use, since they can be iterated over only once. Half of the time, people will end up putting that in a list anyway to pass it around.

Loops are expensive. If you have a large number elements (such as log entries), you're better of looping over them once to extract everything you need in one place, and then pass the result around, rather than the source. I would argue that what you're describing is an antipattern, and the fact that generators make it more annoying is a feature, not a bug. :)

In other cases, where you _do_ actually want a list, converting an iterator into a list is trivial (converting a list into an iterator is also trivial, but by that point most of the benefit of the latter is lost). Thus making this a generator will offer the client code the most flexibility.

> The only thing a generator would provide is allowing to process entries up to a line that fails to parse.

Or only up to the line the client code is interested in, if it's looking for a specific entry in `dmesg` output and does not need to parse further.

douglas-raillard-arm: > If anything, using a generator may result in a significant speed up, if you routinely break out early after finding what you need (see last point).

Good point.

> looping over them once to extract everything you need in one place

It really depends on how much different things need to be done on that list. Fusing multiple functions just because they take the same input might and ending up with a Frankenstein loop doing a bunch of totally unrelated things to it might not lead to very maintainable code.

> I would argue that what you're describing is an antipattern, and the fact that generators make it more annoying is a feature, not a bug. 

In lazy languages, a simple list both avoids doing the computation ahead of time (just like generators in Python) but also provide a straightforward interface, while still being able to "restart from the beginning", which gives the best of both worlds.

Since we can benefit from lazy parsing as you shown, the balance shifts toward the generator, but in the absence of such benefit, generators usually just mean the client code will need to be sparkled with `list()` for no real reason.

douglas-raillard-arm:
Updated with a generator function instead of returning a list

PR #430, opened 2019-09-20
Target: Allow `TargetTransientError`s when checking for root

marcbonnici:
On some unrooted devices the checking of root status can cause a
`TargetTransientError` therefore update `except` statement to accommodate
these.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -46,7 +46,8 @@ from devlib.module import get_module
 from devlib.platform import Platform
 from devlib.exception import (DevlibTransientError, TargetStableError,
                               TargetNotRespondingError, TimeoutError,
-                              TargetTransientError, KernelConfigKeyError) # pylint: disable=redefined-builtin
+                              TargetTransientError, KernelConfigKeyError,
+                              TargetError) # pylint: disable=redefined-builtin
 from devlib.utils.ssh import SshConnection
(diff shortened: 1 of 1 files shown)


devlib/target.py
```diff
@@ -116,7 +116,7 @@ def is_rooted(self):
             try:
                 self.execute('ls /', timeout=5, as_root=True)
                 self._is_rooted = True
-            except (TargetStableError, TimeoutError):
+            except (TargetStableError, TargetTransientError, TimeoutError):
```
setrofim: Just replace `TargetStableError` with `TargetError` instead?

marcbonnici: Updated.

devlib/target.py
```diff
@@ -46,7 +46,8 @@
 from devlib.platform import Platform
 from devlib.exception import (DevlibTransientError, TargetStableError,
                               TargetNotRespondingError, TimeoutError,
-                              TargetTransientError, KernelConfigKeyError) # pylint: disable=redefined-builtin
+                              TargetTransientError, KernelConfigKeyError,
```
setrofim: `TargetTransientError` no longer necessary here as i s covered b `TargetError`

marcbonnici: The `TargetTransientError` is still raised explicitly in other parts of the code so the import is still required.

PR #431, opened 2019-10-02
instruments: Add teardown method to clean up tempfiles

marcbonnici:
Implement the `teardown` method in instruments that utilise tempfiles
which were previously left behind.

### DIFF
--- a/devlib/instrument/acmecape.py
+++ b/devlib/instrument/acmecape.py
@@ -58,12 +58,14 @@ class AcmeCapeInstrument(Instrument):
                  iio_capture=which('iio-capture'),
                  host='baylibre-acme.local',
                  iio_device='iio:device0',
-                 buffer_size=256):
+                 buffer_size=256,
+                 keep_raw=False):
         super(AcmeCapeInstrument, self).__init__(target)
         self.iio_capture = iio_capture
         self.host = host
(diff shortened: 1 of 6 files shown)


setrofim:
This behaviour should be documented (in particular, `get_raw()` documentation should mention that it will no work after `teardown()` is called, and `teardown()` documentation currently appears to be missing.

It should be noted that "raw" files and not necessarily temporary -- the user may wish to hold on to them in case they want to extract further data from them at a later date. However, cleaning them up as the "common case" seems reasonable; but it should be made clear to the user that if they want to keep them, the will have to created copies themselves.

You should also check (if you haven't already) that any of the files that are now being deleted are not being added as "raw" Artefacts by WA.

marcbonnici:
The documentation has been updated to reflect the changes. 

In WA we currently only use the `get_raw` method in the fps instrument however the files do not get recorded as atifacts. Although this instrument does have a `keep_raw` parameter which conditionally deletes the raw files, so was wondering whether it would be worth adding a similar parameter to the devlib instruments if this was still a relatively common use case?

setrofim:
Yes, adding `keep_raw` parameter makes sense.

marcbonnici:
Updated.

PR #432, opened 2019-10-16
module/thermal: List directories with as_root=True

douglas-raillard-arm:
Listing thermal zones directories in sysfs fails when running as non-root.


douglas-raillard-arm:
@marcbonnici Updated with `target.is_rooted`

Issue #412, opened 2019-07-30
Enable post initalisation of modules onto a `Target` 


marcbonnici:
Allow the installation of modules to be performed after the initial connection of the `Target` has been created.

douglas-raillard-arm:
@marcbonnici On-demand loading would be very useful in LISA, as we currently load all modules to workaround the limitation. This would reduce the time it takes to connect to a target.

(note: the "on-demand" part can be handled in LISA side if that's something considered too dynamic for something like devlib)

PR #433, opened 2019-10-21
[RFC] Module installation improvements

marcbonnici:
- Implements post initialised module installation. https://github.com/ARM-software/devlib/issues/412 
- Displays explicit message warning user upon module installation failure.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -796,6 +796,18 @@ class Target(object):
                                                 strip_null_chars)
         return _build_path_tree(value_map, path, self.path.sep, dictcls)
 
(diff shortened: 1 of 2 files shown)


devlib/target.py
```diff
@@ -42,7 +42,7 @@
 from enum import Enum
 
 from devlib.host import LocalConnection, PACKAGE_BIN_DIRECTORY
-from devlib.module import get_module
+from devlib.module import get_module, Module
```
setrofim: Why is `Module` imported?

marcbonnici: Good point, removed.

PR #435, opened 2019-11-01
ftrace: Enable alternative tracers

douglas-raillard-arm:
"function_graph" tracer allows getting funcgraph_entry/funcgraph_exit events for
listed functions. This allows getting precise information on when a given
function was called, and how long its execution took (to build a time-based
heatmap for example).

This can be enabled using:
     FtraceCollector(target, functions=['foo', 'bar'], tracer='function_graph')

If needed, children functions can also be traced with
trace_children_functions=True .

### DIFF
--- a/devlib/trace/ftrace.py
+++ b/devlib/trace/ftrace.py
@@ -125,9 +125,6 @@ class FtraceCollector(TraceCollector):
             self.target_binary = 'trace-cmd'
 
         # Validate required events to be traced
-        available_events = self.target.execute(
-                'cat {}'.format(self.available_events_file),
-                as_root=True).splitlines()
         selected_events = []
         for event in self.events:
             # Convert globs supported by FTrace into valid regexp globs
@@ -136,7 +133,7 @@ class FtraceCollector(TraceCollector):
                 _event = '*' + event
             event_re = re.compile(_event.replace('*', '.*'))
             # Select events matching the required ones
-            if not list(filter(event_re.match, available_events)):
+            if not list(filter(event_re.match, self.available_events)):
                 message = 'Event [{}] not available for tracing'.format(event)
                 if strict:
                     raise TargetStableError(message)
@@ -188,6 +185,14 @@ class FtraceCollector(TraceCollector):
         List of ftrace tracers supported by the target's kernel.
         """
         return self.target.read_value(self.available_tracers_file).split(' ')
+
+    @property
+    @memoized
+    def available_events(self):
+        """
+        List of ftrace events supported by the target's kernel.
+        """
+        return self.target.read_value(self.available_events_file).splitlines()
 
     def reset(self):
         if self.buffer_size:
--- a/devlib/trace/ftrace.py
+++ b/devlib/trace/ftrace.py
@@ -125,21 +125,33 @@ class FtraceCollector(TraceCollector):
             self.target_binary = 'trace-cmd'
 
         # Validate required events to be traced
-        selected_events = []
-        for event in self.events:
-            # Convert globs supported by FTrace into valid regexp globs
-            _event = event
-            if event[0] != '*':
-                _event = '*' + event
-            event_re = re.compile(_event.replace('*', '.*'))
-            # Select events matching the required ones
-            if not list(filter(event_re.match, self.available_events)):
-                message = 'Event [{}] not available for tracing'.format(event)
-                if strict:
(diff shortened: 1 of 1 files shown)


devlib/trace/ftrace.py
```diff
@@ -177,9 +187,16 @@ def start(self):
...
+            function_string = self.function_string
+        else:
+            function_string = ''
+
+        tracer_string = '-p {}'.format(self.tracer) if self.tracer else ''
+
         self.target.write_value(self.trace_clock_file, self.trace_clock, verify=False)
         self.target.write_value(self.save_cmdlines_size_file, self.saved_cmdlines_nr)
-        self.target.execute('{} start {}'.format(self.target_binary, self.event_string),
+        self.target.execute('{} start {} {} {}'.format(self.target_binary, self.event_string, tracer_string, function_string),
```
marcbonnici: Can this be line wrapped a bit nicer?

devlib/trace/ftrace.py
```diff
@@ -177,9 +187,16 @@ def start(self):
         if self._reset_needed:
             self.reset()
 
+        if self.tracer is not None and 'function' in self.tracer:
+            function_string = self.function_string
+        else:
+            function_string = ''
+
```
marcbonnici: If we're changing to use a local version of `function_string` here would it make sense for line 215 to also be updated for consistency?

douglas-raillard-arm: Using `function_string` to write to sysfs would lead to write an empty string when `self.tracer is None`, which is true when executing line 215. That local `function_string` is only meant to be given to trace-cmd. Maybe I should rename it to `tracecmd_functions` instead.

devlib/trace/ftrace.py
```diff
@@ -163,7 +161,19 @@ def __init__(self, target,
...
+            if self.tracer is None:
+                self.function_string = _build_trace_functions(selected_functions)
+                # If function profiling is enabled we always need at least one event.
+                # Thus, if not other events have been specified, try to add at least
+                # a tracepoint which is always available and possibly triggered few
+                # times.
+                if not selected_events:
+                    selected_events = ['sched_wakeup_new']
+            elif self.tracer == 'function_graph':
+                self.function_string = _build_graph_functions(selected_functions, trace_children_functions)
```
setrofim:
```
else:   # unknown  tracer
````

File: devlib/trace/ftrace.py [hunk 4]
```diff
@@ -190,7 +214,7 @@ def start(self):
             self.logger.debug('Trace CPUIdle states')
             self.target.cpuidle.perturb_cpus()
         # Enable kernel function profiling
-        if self.functions:
+        if self.functions and self.tracer is None:
```
setrofim: `and self.tracer != 'function_graph'`?

devlib/trace/ftrace.py
```diff
@@ -177,10 +187,24 @@ def start(self):
         if self._reset_needed:
             self.reset()
 
+        if self.tracer is not None and 'function' in self.tracer:
```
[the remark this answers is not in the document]
douglas-raillard-arm: The `-l` and `-g` options apply to both `function` and `function_graph` tracers.

devlib/trace/ftrace.py
douglas-raillard-arm: We don't want to execute this block when `self.trace == "function"` as trace-cmd will handle it by itself.

devlib/trace/ftrace.py
douglas-raillard-arm: There are quite a few other tracers, and they seem to work (haven't tried them all but a couple of them): `man trace-cmd-record` gives this list: function, function_graph, preemptirqsoff,  irqsoff, preemptoff and wakeup
but in practice the list of supported tracers depends on the kernel.

setrofim: Can we query the available tracers from debugfs and then match against that?

douglas-raillard-arm: Yes, do you want me to add some validation on the tracer ?

setrofim: That would be great.

douglas-raillard-arm:
Updated with:
* validation of the tracer
* new memoized properties: `available_tracers` and `available_events`
* report unavailable events all at once, rather than one by one.

PR #436, opened 2019-11-01
instrument/trace-cmd: Do not set default for number of cmdlines

marcbonnici:
Not all devices have the `saved_cmdlines_size` node exposed and therefore
attempting to set this will fail. Only attempted to set this if
this value is explicitly passed in.

### DIFF
--- a/devlib/trace/ftrace.py
+++ b/devlib/trace/ftrace.py
@@ -80,6 +80,7 @@ class FtraceCollector(TraceCollector):
         self.automark = automark
         self.autoreport = autoreport
         self.autoview = autoview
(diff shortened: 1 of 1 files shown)


marcbonnici:
@douglas-raillard-arm Does this approach work for you?

douglas-raillard-arm:
Another approach would be to raise when strict=True and just warn when strict=False. LISA will always set it to 4096 anyway, so it might be better to just use a best effort approach in devlib when possible like for the events.

marcbonnici:
Updated to utilise the state of `strict`.

PR #438, opened 2019-11-04
ftrace: Ensure /proc/kallsyms contains symbol addresses

douglas-raillard-arm:
The content of /proc/kallsyms depends on the value of
/proc/sys/kernel/kptr_restrict:
* If 0, restriction is lifted and kallsyms contains real addresses
* If 1, kallsyms will contain null pointers

Since trace-cmd records the content of kallsyms into the trace.dat and uses that
to pretty-print function names (function tracer/grapher), ensure that its
content is available.

### DIFF
--- a/devlib/trace/ftrace.py
+++ b/devlib/trace/ftrace.py
@@ -20,6 +20,7 @@ import time
 import re
 import subprocess
 import sys
+import contextlib
 from pipes import quote
 
 from devlib.trace import TraceCollector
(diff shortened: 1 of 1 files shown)


devlib/trace/ftrace.py
```diff
@@ -21,6 +21,7 @@
 import subprocess
 import sys
 from pipes import quote
+import contextlib
```
setrofim: Import should go above `from`

PR #443, opened 2019-11-26
daq: move to daqpower 2.0

JaviMerino:
daqpower 2.0 has a new interface and it is more stable.

### DIFF
--- a/devlib/instrument/daq.py
+++ b/devlib/instrument/daq.py
@@ -52,6 +52,7 @@ class DaqInstrument(Instrument):
         self.keep_raw = keep_raw
         self._need_reset = True
         self._raw_files = []
+        self.tempdir = None
         if DaqClient is None:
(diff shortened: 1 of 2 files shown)


devlib/instrument/daq.py
```diff
@@ -156,10 +153,7 @@ def get_raw(self):
         return self._raw_files
 
     def teardown(self):
-        self.execute('close')
+        self.daq_client.close()
         if not self.keep_raw:
             if os.path.isdir(tempdir):
```
marcbonnici: It looks like this might have been an existing issue however `tempdir` is not defined here.

JaviMerino: Yes, this has been an issue since it was introduced in bb1552151a32e6d21f4c97763d73a071557661b5. I'll add a commit to fix it.

JaviMerino: Fixed in 01c5376

PR #446, opened 2019-11-26
module/sched: Fix get_capacities() on !SCHED_DEBUG kernels

valschneider:
While reading the DT-provided capacity values (exposed in sysfs) is
sufficient, get_capacities() also unconditionally fetches data from the
sched_domain procfs, which is only populated on kernels compiled with
CONFIG_SCHED_DEBUG.

Tweak the logic to only call get_sd_info() if it is both possible and
required.

### DIFF
--- a/devlib/module/sched.py
+++ b/devlib/module/sched.py
@@ -469,12 +469,26 @@ class SchedModule(Module):
         cpus = self.target.list_online_cpus()
 
         capacities = {}
-        sd_info = self.get_sd_info()
 
         for cpu in cpus:
             if self.has_dmips_capacity(cpu):
                 capacities[cpu] = self.get_dmips_capacity(cpu)
-            elif self.has_em(cpu, sd_info.cpus[cpu]):
+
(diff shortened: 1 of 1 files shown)


devlib/module/sched.py
```diff
@@ -469,12 +469,25 @@ def get_capacities(self, default=None):
...
+        missing_cpus = set(cpus).difference(capacities.keys())
+        if not missing_cpus:
+            return capacities
+
+        if not SchedProcFSData.available(self.target):
+            if default != None:
+                capacities.update({cpu : default for cpu in missing_cpus})
+                return capacities
+            else:
+                raise RuntimeError('No capacity data for cpus {}'.format(missing_cpus))
```
douglas-raillard-arm: It's a good idea to turn sets into sorted lists before leaking them outside, since the iteration order is not defined. This allows useful diff from outputs of one run to another.

valschneider: Right, let me fix and resend.

Issue #410, opened 2019-07-30
Remove adb specific methods from `AndroidTarget`


marcbonnici:
Currently `AndroidTarget` contains adb specific methods such as `adb_kill_server`, `adb_root` etc. These do not belong here as adb is not a part of a `Target` as it is entirely possible to have an `SSHConnection` to communicate with an `AndroidTarget` and therefore these methods would not make sense.

marcbonnici:
Implemented with https://github.com/ARM-software/devlib/pull/453

PR #453, opened 2019-12-05
[Breaking Change] ADB Refactor

marcbonnici:
As per https://github.com/ARM-software/devlib/issues/410, `AndroidTarget` should be independent to the underlying connection, therefore move ADB specific methods into the `AdbConnection`.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1225,13 +1225,6 @@ class AndroidTarget(Target):
 
     def connect(self, timeout=30, check_boot_completed=True):  # pylint: disable=arguments-differ
         device = self.connection_settings.get('device')
-        if device and ':' in device:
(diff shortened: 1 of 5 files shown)


devlib/target.py
```diff
@@ -1530,28 +1530,6 @@ def clear_logcat(self):
     def get_logcat_monitor(self, regexps=None):
         return LogcatMonitor(self, regexps)
 
-    def adb_kill_server(self, timeout=30):
```
setrofim: Some of these should still be exposed via `AndroidTarget` (by calling through to the connection method)?

marcbonnici: Updated. `wait_for_device` and `reboot_bootloader` are now exposed in AndroidTarget
as generic methods.

PR #456, opened 2019-12-17
devlib/trace/ftrace.py: Fix reset and stop states

ambroise-arm:
Issue: when using an FtraceCollector object with a nop tracer followed by a function_graph tracer, the second one will not work.

This patch makes sure the state is set on reset and left in a proper state on stop.

### DIFF
--- a/devlib/collector/ftrace.py
+++ b/devlib/collector/ftrace.py
@@ -227,6 +227,8 @@ class FtraceCollector(CollectorBase):
             self._set_buffer_size()
         self.target.execute('{} reset'.format(self.target_binary),
                             as_root=True, timeout=TIMEOUT)
+        if self.functions:
+            self.target.write_value(self.function_profile_file, 0, verify=False)
         self._reset_needed = False
 
     def start(self):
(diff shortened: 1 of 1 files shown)


devlib/trace/ftrace.py
```diff
@@ -225,6 +225,8 @@ def reset(self):
             self._set_buffer_size()
         self.target.execute('{} reset'.format(self.target_binary),
                             as_root=True, timeout=TIMEOUT)
+        self.target.execute('echo 0 > {}'.format(self.function_profile_file),
```
marcbonnici: This currently will be executed regardless of whether the function profiler was to be used / supported on the device which can cause errors if unavailable. 
I would suggest adding a guard here as in other parts of the code as to whether this should be reset.

devlib/trace/ftrace.py
```diff
@@ -225,6 +225,9 @@ def reset(self):
             self._set_buffer_size()
         self.target.execute('{} reset'.format(self.target_binary),
                             as_root=True, timeout=TIMEOUT)
+        if self.functions:
```
marcbonnici: Might also check that `self.tracer` is `None`.

ambroise-arm: I don't know about `None`, but it should at least be checked against `function_graph` since that is the goal of the patch.

marcbonnici: I'm just thinking if there would be cases that do not have this exposed and still want to use `function_graph`. 
In which case maybe change this echo to `self.target.write_value(self.function_profile_file, 0, verify=False)` so that the value is written if it can be and if not, any errors can be safely ignored as they should be picked up elsewhere in the code that was actively using it?

PR #465, opened 2020-01-30
Use target connections as contexts.

setrofim:
Make it easier to use multiple connections to the target from the same thread.


setrofim:
Cc @douglas-raillard-arm: this should be largely orthogonal to #450, I believe, but FYI, just in case.

doc/target.rst
```diff
@@ -184,6 +184,13 @@ Target
              blocking operation (e.g. using ``background()``) while at the same
              time doing something else in the same host-side thread.
 
+.. method:: Target.set_connection(conn)
+
+   Set the Target's connection for the current thread to the one specified
+   (typically, one that has previously been returned by the call to
+   ``get_connection``). Returns the old connection to the current thread -- it
+   is up to the coller to keep track of it and restore it if they wish.
```
marcbonnici: /s/coller/caller.

doc/connection.rst
```diff
@@ -249,3 +249,32 @@ The only methods discussed below are those that will be overwritten by the
...
+thread. You can then use these connections as context managers to execute
+target commands using them rather than the default conection for the thread:
+
+```
+conn = target.get_connection()
+with conn:
+    target.execute('ls')  # uses conn rather than the default connection.
+```
+
+If the connection object is being uses withn another function, you do not need
```
marcbonnici: uses->used

marcbonnici: /s/withn/within

doc/connection.rst
```diff
@@ -249,3 +249,32 @@ The only methods discussed below are those that will be overwritten by the
 .. method:: _wait_for_boot(self)
 
     Wait for the gem5 simulated system to have booted and finished the booting animation.
+
+
+.. _multiple-connections:
+
+Multipe Connections With One Target
```
marcbonnici: Is it worth adding a note somewhere that Gem5Connections and TelnetConnections do not support this?

doc/connection.rst
```diff
@@ -249,3 +249,32 @@ The only methods discussed below are those that will be overwritten by the
 .. method:: _wait_for_boot(self)
 
     Wait for the gem5 simulated system to have booted and finished the booting animation.
+
```
marcbonnici: Update the info on line 19 about a common base?

doc/connection.rst
setrofim: Good point. I'll add a note for `Gem5Connection`. `TelnetConnection` actually derives from `SshConnection` so it would be supported.

douglas-raillard-arm:
You might want to take the patches that fixes thread local variables to store the connection first, they somehow conflicts, especially on the `set_connection()` side that is already handled by `tls_property` out of the box:
https://github.com/ARM-software/devlib/pull/450/commits/d8f9fac796b4293be785e89c1ff5016083d606da
https://github.com/ARM-software/devlib/pull/450/commits/aa3b0e357cf27927685db9c70266ee936559dc66

setrofim:
> There is also a ConnectionBase class introduced:
> d7908d1
> But it should be straightforward to merge the features of both as they don't seem to overlap.

Ah cool, I'll hold off on this then, until that is merged, and will then add the changes to that. One nit -- could you rename the file to `connection.py` (singular) to be consistent with the other files.

doc/connection.rst
marcbonnici: In which case for `TelnetConnection`, a call to the `BaseConnection` `initi` method will also need to be added.

setrofim: Ah, yes. Good catch. Fixed.

PR #467, opened 2020-02-18
utils/android: Add support for using aapt2

marcbonnici:
aapt is now depreciated in favour of aapt2 therefore prefer using the
newer binary if it is found on the system, otherwise fallback to the
old implementation.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -93,6 +93,7 @@ android_home = None
 platform_tools = None
 adb = None
 aapt = None
+aapt_version = None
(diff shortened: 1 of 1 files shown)


setrofim:
Is the `aapt2` CLI the same as `aapt`'s?

marcbonnici:
On second thoughts no they're not.

marcbonnici:
The CLI interface remains compatible for the invocations within devlib however the `aapt_version` attribute is now exposed to indicate which version has been selected to allow for any differences.

PR #480, opened 2020-04-16
Logcat collector fix

JaviMerino:
The documentation for the collectors uses LogcatCollector() as an example, but I don't think it has ever worked with python3. This PR fixes the collector so that it works as the documentation expects it to.


JaviMerino:
While I was reading this code, I was pondering if the correct fix for the LogcatMonitor is to not use pexpect and instead use `subprocess.check_output()` or `Popen`. Let me know if you prefer this.

PR #489, opened 2020-06-03
[RFC] Update android tools discovery

marcbonnici:
Allows falling back to discovering of the currently required android tools from PATH instead of an Android SDK installation. 

Enable the discovery of `aapt2` and default to this if a suitable binary is found:
- This is required due to the fact that `appt` is now depreciated and cannot handle some apks which have valid characters in their description.
- Unfortunately `aapt2` versioning does not seem very mature yet and does not have a nice way of determining whether a particular version supports the subcommands that devlib requires. For example the `version` output from a known working version is `Android Asset Packaging Tool (aapt) 2.19-SOONG BUILD NUMBER PLACEHOLDER`. 
- Therefore to detect if a suitable version is present, try to execute a sub command and look for an expected error message as opposed to a unknown command error. 

Supersedes: https://github.com/ARM-software/devlib/pull/467

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -637,6 +637,7 @@ class _AndroidEnvironment(object):
     def __init__(self):
         self.android_home = None
         self.platform_tools = None
+        self.build_tools = None
         self.adb = None
         self.aapt = None
         self.fastboot = None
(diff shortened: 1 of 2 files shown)


devlib/utils/android.py
```diff
@@ -672,31 +677,66 @@ def _init_common(env):
...
-    env.build_tools = build_tools_directory
+    if os.path.isdir(build_tools_directory):
+        env.build_tools = build_tools_directory
+
+def _check_supported_aapt2(binary):
+    # At time of writing the version argument of aapt2 is not helpful as
+    # the output is only a placeholder that does not distinguish between versions
+    # with and without support for badging. Unfortunately aapt has been
+    # deprecated and fails to parse some valid apks so we will try to favour
+    # appt2 if possible else will fall back to aapt.
```
setrofim, line 688: typo: appt2 --> aapt2

PR #493, opened 2020-06-16
Unify globbing

douglas-raillard-arm:


devlib/connection.py
```diff
@@ -72,6 +72,38 @@ def __del__(self):
...
+        # characters such as wildcard which would defeat the entire purpose of
+        # that function.
+        for c in [' ', "'", '"']:
+            pattern = pattern.replace(c, '\\' + c)
+
+        cmd = "printf '%s\n' {}".format(pattern)
+        result = self.execute(cmd, strip_colors=False)
+        paths = result.splitlines()
+        if not paths:
+            raise TargetStableError('No file matching: {}'.format(pattern))
```
marcbonnici, line 97: Even when there are no matches will `printf` not echo the pattern itself back?

douglas-raillard-arm: If there is no match, printf won't be called. What happens is shell dependent: on most shells, $? will be set to non-zero value and the shell itself will print an error message saying there was no match. On busybox ash, the command won't be called and $?=0. On ZSH with the right option enabled, it can also decide to just ignore the non-matching pattern and just skip that parameter entirely.

AFAIK, no shell will ever call the command with the pattern itself though if there is no match.

douglas-raillard-arm:
>Would there be any benefit to mirroring this functionality for push as well?

We could implement it directly in Target.push(), as we can easily glob on the host side.

setrofim:
If pull() is being globbed at connection level, so should the push().

douglas-raillard-arm:
>If pull() is being globbed at connection level, so should the push().

Yes I realized that and factored out the connection globbing in Target.pull(). Connection.pull() now takes a list of sources instead of just one since most connections can transfer multiple files in one "connection" (as in SFTP connection/adb command/...)

douglas-raillard-arm:
@marcbonnici @setrofim I've updated the PR. It has not been tested a lot so I'll do that from now, but I think it's worth doing the review now given that there are quite a few paths to check.

Also:
* localhost and paramiko connection push/pull were not behaving exactly like the other ones when it came to folder transfer, so I aligned it.
* Added Target.makedirs() since it's a pretty common operation (at least in LISA), and it turned out to be useful once again in that PR so here it is.
* Optimized push/pulling on adb and scp connections so that only one command is used for multiple files. As a side effect, ADB escaping should now be more robust than before.
* Optimized-away globbing when there is not special char in the pattern, to save an execute()

Hopefully the resulting code should be easier to deal with:
* target does globbing in a unified way regardless of the connection (using busybox sh), and it's output can be easily inspected
* connections are only expected to take a list of sources and send them to destination, as-is without applying any sort of transformation (so if there is a quoting issue, it should be obvious)

Remaining question:
Should globbing be on or off by default ? Turning it off leads to more robust user code by default, since it does not require as much external input validation.

douglas-raillard-arm:
>I still need to investigate where this is going wrong but when running with this PR in a WA run using an ADB connection (SSH seemed to work) I get the initial part of the push command being duplicated with the incorrect destination .E.g.:

I unfortunately don't have access to any android target currently, but I cannot see where this could come from by looking at the code. No value is duplicated anywhere, and especially since it seems to be happening pretty late (when the whole adb command is created, which is in `adb_command()`)

> however if it was already on by default turning it off by default may cause code breakages..
Potentially, although the current paramiko SSH connection have been there for a while now and nobody complained so far. An alternative could be to to use a tristate value for globbing:
* True=enabled
* False=disabled
* None=enabled (default) but deprecation warning asking to set globbing=True if the filename is actually a pattern (containing * ? [ ]).

Since we wrap the devlib target into a custom object in LISA, I can set a different default since most of our uses are burried in library code that should behave well in presence of generated names.

devlib/utils/android.py
```diff
@@ -287,28 +288,24 @@ def __init__(self, device=None, timeout=None, platform=None, adb_server=None,
...
-                command = "pull {} {}".format(quote(line.strip()), quote(dest))
-                adb_command(self.device, command, timeout=timeout, adb_server=self.adb_server)
-            return
-        command = "pull {} {}".format(quote(source), quote(dest))
-        return adb_command(self.device, command, timeout=timeout, adb_server=self.adb_server)
+        paths = sources + [dest]
+
+        # Quote twice to avoid expansion by host shell, then ADB globbing
+        do_quote = lambda x: quote(glob.escape(x))
+        paths = ' '.join(map(do_quote, sources))
```
marcbonnici, line 299: Found the mistake causing the destination to be dropped. `sources` -> `paths`.

douglas-raillard-arm: Updated the PR with the fix, thanks

setrofim:
Regarding globbing being being turned off by default -- I think that is the right move. I'm not a fan of the tristate solution as that will generate warnings for every push/pull.  Needing globbing is the exceptional condition, so requiring an explicit `globbing=False` every time you _don't_ need it does not seem reasonable to me.

_sigh_ I hate breaking changes, but given the choice between that and making the API permanently worse, I think we're better off eating the break now.

This change should definitely be thoroughly documented in the release notes.

PR #501, opened 2020-07-10
target: support threads in ps

stephenkyle-ARM:
By passing 'threads=True' to target.ps(), a list of PsThreadEntrys will be
returned for all threads in the target, not just processes. These entries
include an extra value 'tid': the thread ID of this thread. The 'pid' value
represents the pid of the process the thread belongs to.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1163,16 +1163,20 @@ class LinuxTarget(Target):
         else:
             return []
 
-    def ps(self, **kwargs):
-        command = 'ps -eo user,pid,ppid,vsize,rss,wchan,pcpu,state,fname'
+    def ps(self, threads=False, **kwargs):
+        ps_flags = '-eo'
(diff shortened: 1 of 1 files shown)


stephenkyle-ARM:
I've tested this on both Linux (sshing into a workstation) and Android device. This provides some functionality for a future PR: the perf instrument attaching to threads of processes based on search terms.

setrofim:
It's not a good idea to have the result change based on the argument. Why not just modify the existing `PsEntry` to have the `tid`?

stephenkyle-ARM:
Fair, but what should the value of `tid` be in that case? The same as `pid`? `None`?

setrofim:
It should be the `tid`  for the main thread of the process; i.e. the same as  `pid`.

stephenkyle-ARM:
Updated, PTAL.

PR #503, opened 2020-07-21
ADB file transfer fixes

marcbonnici:


devlib/target.py
```diff
@@ -494,7 +494,7 @@ def do_pull(sources, dest):
             for source in sources:
                 with self._xfer_cache_path(source) as device_tempfile:
                     self.execute("cp -r -- {} {}".format(quote(source), quote(device_tempfile)), as_root=True)
-                    self.execute("chmod 0644 -- {}".format(quote(device_tempfile)), as_root=True)
+                    self.execute("chmod 0644 {}".format(quote(device_tempfile)), as_root=True)
```
setrofim, line 497: Does the `busybox` version?

marcbonnici: Yes it does. Updated to use that instead.

PR #505, opened 2020-08-11
devlib/target: Enable screen stay-on mode:

jrpdotjpg:
Adds the ability to set the android global setting
``stay_on_while_plugged_in``.

This setting has 4 main modes:
- 0: never stay on
- 1: stay on when plugged in to AC charger
- 2: stay on when plugged in to USB charger
- 4: stay on when wirelessly charged

These values can be OR-ed together to produce combinations.


setrofim:
This needs to be documented.

setrofim (approved):

marcbonnici (approved):

PR #509, opened 2020-11-04
utils/ssh: Remove sudo prompt from output

valschneider:
On a target where sudo is required, target.file_exists() erroneously
returns True despite the execute() output being:

  '[sudo] password for valsch01: 0\n'

The sudo prompt is being written to stderr (as per sudo -S), but this is
still merged into the final execute() output. Get rid of the prompt to
prevent it from interfering with any command output processor.

### DIFF
--- a/devlib/host.py
+++ b/devlib/host.py
@@ -101,7 +101,7 @@ class LocalConnection(ConnectionBase):
             if self.unrooted:
                 raise TargetStableError('unrooted')
             password = self._get_password()
-            command = 'echo {} | sudo -S -- sh -c '.format(quote(password)) + quote(command)
+            command = 'echo {} | sudo -p ' ' -S -- sh -c '.format(quote(password)) + quote(command)
         ignore = None if check_exit_code else 'all'
         try:
             return check_output(command, shell=True, timeout=timeout, ignore=ignore)[0]
@@ -118,7 +118,7 @@ class LocalConnection(ConnectionBase):
             if self.unrooted:
                 raise TargetStableError('unrooted')
             password = self._get_password()
-            command = 'echo {} | sudo -S '.format(quote(password)) + command
+            command = 'echo {} | sudo -p ' ' -S '.format(quote(password)) + command
 
         # Make sure to get a new PGID so PopenBackgroundCommand() can kill
         # all sub processes that could be started without troubles.
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -284,7 +284,7 @@ class SshConnectionBase(ConnectionBase):
                  keyfile=None,
                  port=None,
                  platform=None,
-                 sudo_cmd="sudo -S -- sh -c {}",
+                 sudo_cmd="sudo -p ' ' -S -- sh -c {}",
                  strict_host_check=True,
                  ):
         super().__init__()
@@ -364,7 +364,7 @@ class SshConnection(SshConnectionBase):
                  port=22,
                  timeout=None,
                  platform=None,
-                 sudo_cmd="sudo -S -- sh -c {}",
+                 sudo_cmd="sudo -p ' ' -S -- sh -c {}",
                  strict_host_check=True,
                  use_scp=False
                  ):


setrofim (approved):

marcbonnici (commented):
Just tried this on 2 of my machines, one of them worked as expected however the other (WSL) still printed the prompt. I wonder if setting the prompt to a single space would be more reliable as this should usually be stripped from the output?
```
marc@WSL~$ sudo -p '' -S -- sh -c ls
[sudo] password for marc:
marc@WSL ~ $ sudo -p ' ' -S -- sh -c ls

```

valschneider:
> Just tried this on 2 of my machines, one of them worked as expected however the other (WSL) still printed the prompt. I wonder if setting the prompt to a single space would be more reliable as this should usually be stripped from the output?
> 
> ```
> marc@WSL~$ sudo -p '' -S -- sh -c ls
> [sudo] password for marc:
> marc@WSL ~ $ sudo -p ' ' -S -- sh -c ls
> ```

Can confirm I have the same behaviour on my home WSL with:
```
$ sudo --version
Sudo version 1.8.21p2
Sudoers policy plugin version 1.8.21p2
Sudoers file grammar version 46
Sudoers I/O plugin version 1.8.21p2
```

This seems to work, but I'll have to check whether this won't cause further breakage:
```
# https://serverfault.com/a/967883
sudo -S 2>/dev/null -- sh -c ls
```

marcbonnici:
> This seems to work, but I'll have to check whether this won't cause further breakage:
> https://serverfault.com/a/967883
> sudo -S 2>/dev/null -- sh -c ls

This will also hide any output to stderr from the command itself so unfortunately does not seem like it would be a compatible solution.

valschneider:
@douglas-raillard-arm noted that the host target suffered from the same issue, so I fixed it up as well. There still is the `TelnetConnection` class that may need fixing, but the current actually waits for the prompt, so it didn't seem like a great idea to remove it.

I also hit the `-p ''` snag on an x86 machine running a truly ancient distro, and `-p ' '` worked fine, so I went with that.

douglas-raillard-arm:
> @douglas-raillard-arm noted that the host target suffered from the same issue

I noted that there *might* be an issue as well :)
Turns out we don't, but that's not necessarily a good thing: local connection's execute() method simply discards stderr:
line 109 in host.py, `[0]` is stdout, `[1]` is stderr:
```
            return check_output(command, shell=True, timeout=timeout, ignore=ignore)[0]
```

I suppose we should either align this connection to the ssh (and adb ones ?) where stderr is merged with stdout, in which case we need Valentin's change. The only problem being that "merging" is typically not what the user would expect (there is no synchronisation between stdout and stderr), but the issue is already present for ssh where you receive chunks of output, not individual lines, and that's not really solvable anyway.

marcbonnici:
> I also hit the -p '' snag on an x86 machine running a truly ancient distro, and -p ' ' worked fine, so I went with that.

@valschneider Right that seems like a reasonable approach if that solves the problem. Out of curiosity do you have a reproducer for the original issue? I haven't manged to hit this this so far.

> There still is the TelnetConnection class that may need fixing, but the current actually waits for the prompt, so it didn't seem like a great idea to remove it.

Yea I don't think we could apply the same idea here, although I'm not sure we would actually hit the issue with the way it is implemented, would need to check that out. 

> Note that adb connection also just concatenates both output, so we might just do that for localhost too

@douglas-raillard-arm I agree this seems like the best approach and is more consistent with the other connection types. @setrofim Do you have any thoughts on why we had originally discarded the stderr output for local targetst? 

> local connection has a typo that prevents stderr from being returned anyway when stdout is empty

Good catch thanks.

valschneider:
> > I also hit the -p '' snag on an x86 machine running a truly ancient distro, and -p ' ' worked fine, so I went with that.
> 
> @valschneider Right that seems like a reasonable approach if that solves the problem. Out of curiosity do you have a reproducer for the original issue? I haven't manged to hit this this so far.

Initially I got it triggering on an Ampere eMAG where creating the Target failed on `devlib.platform Platform._set_model_from_target()` - that's an arm64 target but without devicetree (it uses ACPI instead). So
`target.file_exists("/proc/device-tree/model")` would return True, but `target.execute("cat /proc/device-tree/model")` would raise an Exception.

TBH I was surprised when I got this triggering on that x86 machine, it sounds like something that should be noticed fairly quickly. There might be something effed up about my setup, but I can't see what straight away.

marcbonnici (approved):

setrofim (approved):

PR #511, opened 2020-11-09
instrument/daq: Use clock boottime for the time column of the energy measurements

JaviMerino:
92e16ee87374 ("instrument/daq: Add an explicit time column to the DAQ
measurements") added a time column to the DAQ measurements in order to
help correlate them with those of other collectors like
FtraceCollector.  Sadly, FTrace uses CLOCK_BOOTTIME instead of
CLOCK_MONOTONIC by default.  CLOCK_MONOTONIC is like CLOCK_BOOTTIME,
except that it stops when the device suspends, which is why I hadn't
spot the issue until now.

Switch to CLOCK_BOOTTIME to get the intended behaviour of the original
commit.


marcbonnici (commented):
If we can configure the clock used by Ftrace would there be value for having support for both clocks here and changing the default to match the FTrace default?

JaviMerino:
It is possible to change the default clock for ftrace, true. I don't know of anything that uses clock monotonic by default, so I'm unsure of the value of supporting both clocks (or all the clocks in linux) in this devlib instrument. I introduced this in 92e16ee because I thought it was the correct one. I now know I made a mistake, that's why I'm suggesting the switch. We can support all the clocks for the instrument, but I'd rather do it only if there is a use for it.

setrofim:
I think just switching which clock is used is the right call. There is no point in adding the complexity of an additional configuration point if it's not needed.

setrofim (approved):

marcbonnici:
Ok sure, if there are no other use cases at the moment that seems reasonable.

marcbonnici (approved):

PR #515, opened 2020-11-18
Fix quote

douglas-raillard-arm:


marcbonnici (approved):

devlib/host.py
```diff
@@ -112,14 +113,21 @@ def execute(self, command, timeout=None, check_exit_code=True,
...
 
     def background(self, command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, as_root=False):
         if as_root and not self.connected_as_root:
             if self.unrooted:
                 raise TargetStableError('unrooted')
             password = self._get_password()
-            command = 'echo {} | sudo -p ' ' -S '.format(quote(password)) + command
+            # The sudo prompt will add a space on stderr, but we cannot filter
+            # it out here
+            command = "echo {} | sudo -p ' ' -S ".format(quote(password)) + command
```
marcbonnici, line 130: Update this concatenation as well?

douglas-raillard-arm:
@marcbonnici Updated with the other concatenation removed.

I also added a commit to use "sh -c" when using sudo with background(), so we can use arbitrary shell statements (like in execute())

marcbonnici (approved):

PR #516, opened 2020-11-30
utils/apk_info: Handle apks that do not contain classes.dex

marcbonnici:
Some apks do not contain the file that we use to determine app methods
so return an empty list in this case.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -215,10 +215,14 @@ class ApkInfo(object):
     @property
     def methods(self):
         if self._methods is None:
+            # Only try to extract once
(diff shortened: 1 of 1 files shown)


setrofim:
What is the reason for the file being absent? Is the alternative executable format? and if so, can it be parsed to extract the same information?

marcbonnici:
Good question. This issue arose when WA pulled a stock version of the gmail apk for analysis, checking inside the apk manually there does not appear to be any alternatives present, the files are just missing. 
I think that this is an outlier for a badly implemented apk file rather than something we should expect to hit frequently.

PR #523, opened 2021-02-01
collector/perf: Run perf as root if the target is rooted

JaviMerino:
If you want to collect events by event id (eg. in simpleperf, "rNNN" with NNN a number), you must run it as root.  Otherwise, simpleperf fails with a SIGABRT:

    simpleperf F environment.cpp:856] Check failed: tmp_file->fd != -1 (tmp_file->fd=-1, -1=-1)
    Aborted

Run simpleperf as root by default if the target is rooted to avoid this.

### DIFF
--- a/devlib/collector/perf.py
+++ b/devlib/collector/perf.py
@@ -141,7 +141,7 @@ class PerfCollector(CollectorBase):
 
     def start(self):
         for command in self.commands:
-            self.target.kick_off(command)
+            self.target.kick_off(command, as_root=self.target.is_rooted)
 
     def stop(self):
         self.target.killall(self.perf_type, signal='SIGINT',


setrofim (approved):

JaviMerino:
Alternatively, I could change this to only run it as root if using simpleperf and the event matches "r[0-9]+". I think this is better, but feel free to suggest otherwise.

marcbonnici (commented):
This solution letting it use root for all events sounds good to me.

As a formatting nit, while very useful in the PR description, would you be able to remove the complete error message from the commit message itself to help maintain the max line length?

JaviMerino:
> As a formatting nit, while very useful in the PR description, would you be able to remove the complete error message from the commit message itself to help maintain the max line length?

Done!

marcbonnici (approved):

PR #526, opened 2021-02-11
modules/sched: Fix sched domain flags parsing

douglas-raillard-arm:
Recent kernels can have a space-separated list of textual flags rather
than a bitfield packed in an int.

### DIFF
--- a/devlib/module/sched.py
+++ b/devlib/module/sched.py
@@ -15,7 +15,6 @@
 
 import logging
 import re
-from enum import Enum
 
 from past.builtins import basestring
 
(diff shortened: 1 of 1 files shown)


marcbonnici (approved):

devlib/module/sched.py
```diff
@@ -209,12 +209,22 @@ class SchedDomain(SchedProcFSNode):
...
-        for flag in list(SchedDomainFlag):
-            if self.flags & flag.value == flag.value:
-                obj_flags.add(flag)
+        # Recent kernels now have a space-separated list of flags instead of a
+        # packed bitfield
+        flags = self.flags
+        if isinstance(flags, str):
+            flags = flags.split()
+            def has_flag(flags, flag):
+                return flag in flags
```
valschneider, line 218: Unfortunately this change (kernel side I mean) also removed a flag which caused a shift in value for all the others. The point was that moving forward only the flag *name* would have value to external users, as the value cannot be used anymore (there was a writable interface which has now been removed).

I'm regretting the Enum here as I don't see a nice way to dynamically discover available flags while keeping it backwards-compatible. We *could* "fake" an Enum with `.name` being the name read above and `.value` either returning a bogus value + issuing a warning or raising an exception altogether.

Also, worth noting I'm the one who made those kernel changes, so send me your alcohol bill...

douglas-raillard-arm: Is there really an issue in the end ? If the value has no use at all on new kernels, it does not really matter that much that it is bogus. If we are scared someone will use it to debug in kernel issues or that it comes back via e.g. the need to decode an ftrace event field, we can either put a "bomb" value that will raise when used or maybe simply compute the right value if the kernel outputs the names in bitfield order with a known starting point

valschneider: Ack for the value not being an issue for newer kernels

As for runtime discovery of flag names, I don't have any better idea if we're to keep the Enum.

douglas-raillard-arm:
What shall we do then ?

We could replace the Enum in the user-facing API by a class that wraps the flag string so that:
    1. the string is accessible in `.name` for backward compat
    2. `.value` can be a property that either:
        * raises an exception on recent kernels
        * returns the actual value on old kernels, using the current hardcoded table
    3. for any given string, it would always return the same instance so that comparison with `is` keeps working (like for Enum members).
 
This way current code would still mostly work (unless it relies on some more detailed API of Enum members but it's unlikely), and we are not lying about the value (we raise exceptions when we know we don't know).

douglas-raillard-arm:
@marcbonnici @setrofim @valschneider PR updated with an implementation of the above.

Old client code ran against old kernel should work just as well (the Enum is emulated to a good level), and new code should work fine as long as it does not attempt to do things like `x is SchedDomainFlag.SD_LOAD_BALANCE`, because the class attribute is the "old style flag" that gets a value, and therefore cannot be equal to the new-style flag that has a name but no value.

One could argue that we should make these equal, but that would be a mistake: `__eq__` is supposed to return True when the instances are completely interchangeable without any visible difference, not just a random equivalence definition that only works in one context. If this rule is violated, one can end up with unexpected behavior when using it as a key in a dict or in a set, or anywhere else that relies on that. Making it non-hashable solves half of the issue, because there could be places where the value is stored in a list.

devlib/module/sched.py
```diff
@@ -197,24 +197,108 @@ def check_version(cls, target, logger):
...
+
+    def __getitem__(self, i):
+        return self._flags[i]
+
+    def __len__(self, i):
+        return len(self._flags)
+
+    # These would be provided by collections.abc.Sequence, but using it on a
+    # metaclass seems to have issues around __init_subclass__
+    def __iter__(self, i):
```
valschneider, line 227: Testing this on a v4.14 target yields:
```
sd_info = target.sched.get_sd_info()
  File "/data/work/lisa/external/devlib/devlib/module/sched.py", line 484, in get_sd_info
    return SchedProcFSData(self.target)
  File "/data/work/lisa/external/devlib/devlib/module/sched.py", line 332, in __init__
    super(SchedProcFSData, self).__init__(procfs)
  File "/data/work/lisa/external/devlib/devlib/module/sched.py", line 139, in __init__
    self._dyn_attrs[dest][i] = self._build_node(key, nodes[key])
  File "/data/work/lisa/external/devlib/devlib/module/sched.py", line 115, in _build_node
    return SchedProcFSNode._build_directory(node_name, node_data)
  File "/data/work/lisa/external/devlib/devlib/module/sched.py", line 98, in _build_directory
    return SchedProcFSNode(node_data)
  File "/data/work/lisa/external/devlib/devlib/module/sched.py", line 139, in __init__
    self._dyn_attrs[dest][i] = self._build_node(key, nodes[key])
  File "/data/work/lisa/external/devlib/devlib/module/sched.py", line 115, in _build_node
    return SchedProcFSNode._build_directory(node_name, node_data)
  File "/data/work/lisa/external/devlib/devlib/module/sched.py", line 96, in _build_directory
    return SchedDomain(node_data)
  File "/data/work/lisa/external/devlib/devlib/module/sched.py", line 297, in __init__
    for flag in SchedDomainFlag
TypeError: __iter__() missing 1 required positional argument: 'i'
```

AFAICT none of the methods here except `__getitem__()` should have this `i` parameter.

Other than this, I believe this is the right approach, and it works fine on a bleeding edge kernel.

douglas-raillard-arm: ah right, copy paste issue, strange that I haven't hit it in my tests, thanks

douglas-raillard-arm:
Updated with the fix for the extra parameter and with a better `__repr__` implementation

marcbonnici (approved):

PR #531, opened 2021-03-24
target: Add Target.background(timeout=...) parameter

douglas-raillard-arm:
Use a Timer daemonic thread to cancel the command after the given
timeout.


marcbonnici (approved):
Would you be able to add this parameter to the documentation as well?

douglas-raillard-arm:
@marcbonnici Updated the PR with extended doc for both force_locale and timeout

PR #533, opened 2021-03-30
Perf with report samples

JaviMerino:


setrofim (approved):

devlib/collector/perf.py
```diff
@@ -217,7 +234,7 @@ def _pull_target_file_to_host(self, label, extension, output_path):
         target_file = self._get_target_file(label, extension)
         host_relpath = os.path.basename(target_file)
         host_file = _f(os.path.join(output_path, host_relpath))
-        self.target.pull(target_file, host_file)
+        self.target.pull(target_file, host_file, timeout=600)
```
marcbonnici, line 237: Does this require a timeout here? 
If possible it might be better to exclude this to be able to use the transfer polling mechanism as needed?

marcbonnici (approved):

PR #537, opened 2021-04-06
connection: PopenTransferManager: Reset last_sample when we start a n…

JaviMerino:
…ew transfer

last_sample is initialized in the PopenTransferManager constructor.
However, there is only one PopenTransferManager instance, which is
initialized during the construction of AndroidTarget.  Afterwards, the
transfer manager is reused for every file transfer, without going
through __init__().  Therefore, after pushing/pulling a big file, the
next file transfer has compares the current size to the last sample of
the previous file transfer.  This makes it believe that the transfer
is inactive.

Reinitialize last_sample every time we start a new transfer to avoid
this.

### DIFF
--- a/devlib/connection.py
+++ b/devlib/connection.py
@@ -466,6 +466,7 @@ class PopenTransferManager(TransferManagerBase):
         if self.transfer:
             self.transfer.cancel()
             self.transfer = None
+            self.last_sample = None
 
     def isactive(self):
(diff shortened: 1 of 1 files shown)


JaviMerino:
Alternatively, we could just create a TransferManager in AdbConnection:_push_pull(), ie, only when we need the context manager. However, doing the same in SSHConnection seems non-trivial, so I have opted for the less intrusive solution.

marcbonnici (approved):

PR #542, opened 2021-04-16
Busybox Binary Updates

marcbonnici:
With relation to https://github.com/ARM-software/devlib/pull/540

Update the arm32 and add i368 v1.32.1 busybox binaries statically linked
with uclibc rather than the previous version with glibc which
caused segfaults on some devices.

Update ppc64le to v1.32.1 however linked with musl as uclib does not appear 
to be a supported build configuration.


benjaminmordaunt:
As uClibc-ng doesn't support 64-bit PowerPC architecture, should we put a note somewhere that users should always expect hostid = 0 on these platforms?

PR #545, opened 2021-04-26
collector/perf: warn if report_options or report_sample_options are specified when not using perf/simpleperf record

JaviMerino:


devlib/collector/perf.py
```diff
@@ -126,6 +126,10 @@ def __init__(self,
             self.command = command
         else:
             raise ValueError('Unsupported perf command, must be stat or record')
+        if report_options and (command != 'record'):
+            self.logger.warning('report_options specified, but command is not record.  report_options will be ignored')
```
setrofim, line 130: Should probably raise an error here, as this is something that the can correct.

JaviMerino: I have changed it to raise a `ValueError`

setrofim (approved):

marcbonnici (approved):

PR #547, opened 2021-05-07
modules/sched: Add awareness of new debug directory root

valschneider:
Scheduler debug information is being unified under /sys/kernel/debug/sched
for Linux v5.13. Plug in awareness for the new path while still trying the
old one(s) for backwards compatibility.

Tested against v5.12 and upcoming v5.13-rc1 on a Juno r0.

### DIFF
--- a/devlib/module/sched.py
+++ b/devlib/module/sched.py
@@ -21,7 +21,7 @@ from past.builtins import basestring
 from devlib.module import Module
 from devlib.utils.misc import memoized
 from devlib.utils.types import boolean
-
+from devlib.exception import TargetStableError
 
 class SchedProcFSNode(object):
     """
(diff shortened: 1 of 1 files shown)


marcbonnici (approved):

devlib/module/sched.py
```diff
@@ -309,13 +309,23 @@ class SchedProcFSData(SchedProcFSNode):
...
-    sched_domain_root = '/proc/sys/kernel/sched_domain'
+
+    @classmethod
+    def get_data_root(cls, target):
+        # Location differs depending on kernel version
+        for root in ['/sys/kernel/debug/sched/domains/', '/proc/sys/kernel/sched_domain']:
+            if target.file_exists(root):
+                return root
+
+        return None
```
douglas-raillard-arm, line 320: Looks like this should raise an exception. Returning None from a function that usually returns a valid value is one of the greatest footgun, which is already being triggered in that very PR:
```
        if path is None:
            path = SchedProcFSData.get_data_root(target)

        procfs = target.read_tree_values(path, depth=self._read_depth)
```

There is not guarantee that `path is not None`, which will result in a shell call:
```
read_tree_values "''" 1
```
which will probably fail with a non obvious error.

In the absence of good static typing along with something like Rust's Option<T>, exceptions are much more desirable. They have basically the same typical behavior (i.e. bypassing the rest of the code and "returning" early), and they can be caught in the odd case where you want to turn it into a boolean like in `def available`.

devlib/module/sched.py
```diff
@@ -362,6 +375,18 @@ def probe(target):
 
         return schedproc or debug or dmips
 
+    def __init__(self, target):
+        super().__init__(target)
+
+        self.sched_features_path = self.get_sched_features_path(self.target)
+
+    @classmethod
+    def get_sched_features_path(cls, target):
```
douglas-raillard-arm, line 384: Looks like there is the same issue as in the former version of `get_data_root`:
```
self.sched_features_path = self.get_sched_features_path(self.target)     
... 
feats = self.target.read_value(self.sched_features_path)
```

valschneider:
I had forgotten about this until it came back to bite me; @douglas-raillard-arm all good for you now?

Issue #528, opened 2021-03-22
Target.execute() not reentrant


douglas-raillard-arm:
Target.execute() is not reentrant when using (at least) the LocalConnection connection type.

This manifests as a deadlock in `devlib.misc.get_subprocess()`, when it tries to acquire the `check_output_lock`. This can happen when an object's `__del__` uses `target.execute()` and happens to be garbage collected at the same time as `Target.execute()` is being called. The `__del__` method is called at a random point where the lock is already taken, and the gc deadlocks on the lock. Since the gc is called from the main thread, the whole interpreter hangs.

In general, it's not a huge problem because `__del__` should never be relied upon. However:
1. It's sometimes useful to do some last-resort cleanup in `__del__`, which this problem makes it impossible (failing to cleanup is ok, deadlocking not so much)
2. There is a useful idiom used to "package" setup/teardown code:
```
@contextlib.contextmanager
def setup_teardown():
     setup()
     try:
         yield
     finally:
          teardown()


with setup_teardown():
    ....

# But if we want to keep the CM as an implementation detail for backward compat and allow a manual setup()/teardown() API:

cm = setup_teardown()
setup = lambda: cm.__enter__()
teardown = lambda: cm.__exit__(None, None, None)

setup()
# May never happen at all if the user never calls it
teardown()
```

Now we have a problem: the user might never calls teardown(). Since the context manager is based on a generator, it will eventually be closed (manually or via `__del__`), leading to a GeneratorExit being raised at the "yield" point. When this happens, it will basically be as-if teardown() was called, except it can happen asynchronously at any point. Ideally, we would have a generator type that does not call `close()` from `__del__`, but there is nothing we can do for builtin types. The only "fix" is to keep a reference alive forever (and leak memory) so that they are not gc'ed before `__exit__` is called.

That leaves a few choices:
* Make devlib's Target.execute() reentrant
* Make devlib's Target.execute() sort-of-reentrant (i.e. raise a RuntimeError. This will randomly prevent `__del__` from doing useful things and litter the stderr with the backtraces of swallowed exceptions but at least won't deadlock).
* Avoid calling target.execute() from `__del__`, which prevents useful cases.

Maybe connections are actually already reentrant and all we need is to turn the lock into a reentrant lock.

Otherwise, we can probably create a new connection object for nested execute() and background() calls, so connection objects don't have to worry about that issue. It should only happen in exceptional circumstances so I don't think performance is a problem. We already do it for multithreading so I expect it to work.

[...]

PR #548, opened 2021-05-12
target: Create new connection for reentrant calls

douglas-raillard-arm:
When Target.conn property is required while the current connection is
already in use, provide a fresh connection to avoid deadlocks. This is
enabled by the @call_conn decorator that is used on all Target methods
that use self.conn directly.

Fixes https://github.com/ARM-software/devlib/issues/528

note2: if we consider that:
1. we are not interested in passing Target methods as callbacks to connection methods
2. the impact of call_conn is negligible

then we could just apply it to all instance methods of Target automatically in a metaclass (or using `__init_subclass__` on Python >= 3.6) 

note 3: a new connection will be created for every decorated target method call that is done while a call to the connection is active. This this adds overhead, but since it only happens in exceptional situations (from `__del__`) it should be acceptable. Regular use of the target should not be impacted by that.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -76,6 +76,48 @@ GOOGLE_DNS_SERVER_ADDRESS = '8.8.8.8'
 
 installed_package_info = namedtuple('installed_package_info', 'apk_path package')
 
+
+def call_conn(f):
+    """
+    Decorator to be used on all :class:`devlib.target.Target` methods that
+    directly use a method of ``self.conn``.
+
+    This ensures that if a call to any of the decorated method occurs while
+    executing, a new connection will be created in order to avoid possible
+    deadlocks. This can happen if e.g. a target's method is called from
+    ``__del__``, which could be executed by the garbage collector, interrupting
+    another call to a method of the connection instance.
+
+    .. note:: This decorator could be applied directly to all methods with a
+        metaclass or ``__init_subclass__`` but it could create issues when
+        passing target methods as callbacks to connections' methods.
+    """
+
+    @functools.wraps(f)
+    def wrapper(self, *args, **kwargs):
+        reentered = self.conn.is_in_use
+        disconnect = False
+        try:
+            # If the connection was already in use we need to use a different
+            # instance to avoid reentrancy deadlocks. This can happen even in
+            # single threaded code via __del__ implementations that can be
+            # called at any point.
+            if reentered:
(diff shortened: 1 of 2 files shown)


douglas-raillard-arm:
Now tested on linux targets and lightly tested on local target. No testing on android target has been done (we don't have any android device in our CI). It seems to work well, with no unnecessary connection created.

marcbonnici:
Hi @douglas-raillard-arm Thanks for putting this together! 
Do you happen to have a minimal example handy to demonstrate this issue that I could use to test things out on my side and I shall also kick off some tests on my android devices.

douglas-raillard-arm:
Hi @marcbonnici , my main test for this series was to apply a patch like that:
```
diff --git a/devlib/host.py b/devlib/host.py
index 8bd0edf9..2e9b356a 100644
--- a/devlib/host.py
+++ b/devlib/host.py
@@ -96,6 +96,13 @@ class LocalConnection(ConnectionBase):
     # pylint: disable=unused-argument
     def execute(self, command, timeout=None, check_exit_code=True,
                 as_root=False, strip_colors=True, will_succeed=False):
+
+        try:
+            if strip_colors:
+                from __main__ import breakthings
+                breakthings()
+        except ImportError:
+            pass
         self.logger.debug(command)
         use_sudo = as_root and not self.connected_as_root
         if use_sudo:
```

And then call that from a script defining a `breakthings()` function which would call `target.execute()` again to reenter `LocalConnection.execute()`. The actual issue that made me open https://github.com/ARM-software/devlib/issues/528 was based on `__del__` being invoked by the garbage collector at the wrong moment, which depends on various timings on a Juno R0 with a specific workload so it's not easy to reproduce.

If you just want to check that there is no reentrant call on your setup (let's say a regular invocation of WA), you can just apply a patch like that:
```
diff --git a/devlib/target.py b/devlib/target.py
index 3b26a95f..571d213b 100644
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -96,6 +96,7 @@ def call_conn(f):
     @functools.wraps(f)
     def wrapper(self, *args, **kwargs):
         reentered = self.conn.is_in_use
+        assert not reentered
         disconnect = False
         try:
             # If the connection was already in use we need to use a different
```

marcbonnici (approved):

setrofim (approved):

PR #560, opened 2021-08-10
ssh: Align _pull_folder to cp behavior

douglas-raillard-arm:
target.pull('/foo/bar', '.') is expected to create a ./bar folder on the
host, rather than pulling the content of /foo/bar/* directly in "."

This also fixes the issue of trying to delete the "." destination, which
can have very bad consequences.

### DIFF
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -558,6 +558,12 @@ class SshConnection(SshConnectionBase):
         sftp.get(src, dst, callback=self._get_progress_cb())
 
     def _pull_folder(self, sftp, src, dst):
+        if os.path.isdir(dst):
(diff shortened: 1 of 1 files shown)


marcbonnici (commented):
Good catch, however currently this change still doesn't allow doing something like:
``` target.pull('/path/to/file/', '.')``` as shutil fails with `[Errno 22] Invalid argument: './'`. 
As we're aligning with `cp` I wonder if it makes sense to just strip the trailing `/` from the source before doing your change here?

Also just noticed that if the destination is a file, this means it will still be deleted to allow the folder to be pulled which is undesirable. Maybe add an `else` to your `is_dir` check and bail with the error of trying to overwrite a file with a directory if the file exists?

PR #562, opened 2021-08-11
ssh: Fix SFTP pull speed issue

douglas-raillard-arm:
Speed up SFTP pulls by 20x by working around this bug:
https://github.com/paramiko/paramiko/issues/1080

Tested with:
* file sizes of 3.5 and 100MB
* packet size of 1024 and 4096 (max actually supported by the setup I use)

It appears that there is no difference between 1024 and 4096, even for large files (100MB), so I went for 1024 to avoid future issues. The speed I observed with that patch are more in line with `scp` command.

### DIFF
--- a/devlib/utils/ssh.py
+++ b/devlib/utils/ssh.py
@@ -36,6 +36,7 @@ from future.utils import raise_from
 
 from paramiko.client import SSHClient, AutoAddPolicy, RejectPolicy
 import paramiko.ssh_exception
+from paramiko.sftp_file import SFTPFile
 from scp import SCPClient
 # By default paramiko is very verbose, including at the INFO level
 logging.getLogger("paramiko").setLevel(logging.WARNING)
@@ -58,6 +59,15 @@ from devlib.utils.types import boolean
 from devlib.connection import (ConnectionBase, ParamikoBackgroundCommand, PopenBackgroundCommand,
                                SSHTransferManager)
 
+
+# Workaround a speed issue in paramiko:
+# https://github.com/paramiko/paramiko/issues/1080
+# If left to its default value, paramiko SFTPClient.get() will prefetch
+# assuming MAX_REQUEST_SIZE blocks. The server will then send a smaller block,
+# rendering the prefetching useless, and paramiko falls back on sequential
+# read, which destroys performance (more than 20x slow down in a case with a
+# 3.5MB file)
+SFTPFile.MAX_REQUEST_SIZE = 4096
 
 DEFAULT_SSH_SUDO_COMMAND = "sudo -k -p ' ' -S -- sh -c {}"
 


douglas-raillard-arm:
Ok so plot twist:
* Setting the max packet size to 1024 does affect transfer speed of larger files
* The speed issue only appears when pulling from sysfs. If the exact same file is pulled from /root, there is no problem

So that probably is a bug on OpenSSH server side, probably because sysfs files report a size of 0 (or another similar weirdness).

We may want to address that by making a copy to a temp location when pulling from sysfs, but this PR is probably not what we want.
Otherwise, we can set it to 4096, which anecdotally happens to improve speed for larger files (25s vs 36 for 100MB on my setup, not sure how well it would generalize)

setrofim:
> We may want to address that by making a copy to a temp location when pulling from sysfs, but this PR is probably not what we want.

Yeah.  We already do that why transferring as root (e.g.  https://github.com/ARM-software/devlib/blob/master/devlib/target.py#L556), to get around permission/ownership issues. We could update to use this for non-root transfers as well. However, given that this would significantly increase the number of host<->target calls if used always, we probably want to make make it contingent on a parameter; or alternatively, identify sysfs/debugfs mount points during target initialisation and check whether the on-target path is under one of those....

douglas-raillard-arm:
On of the problem with sysfs is that since the default reported file size is 0, there is no way for devlib to know in advance whether a particular file is going to exhibit the issue (small files under 4k will not suffer from it). I'll see if that can be solved on the kernel side first, maybe sysfs does allow setting a file size and hopefully OpenSSH should be able to handle the file correctly after that.

marcbonnici:
Just to add some more data points, these are some results transferring 2 larger non sysfs files, to/ from one machine to another running a docker container over a gigabit network link with varying `MAX_REQUEST_SIZE` values. 
```
96MB Apk File
Size (Default): Push time: 2.955543041229248
Size (Default): Pull time: 1.1721482276916504
Size (4096): Push time: 8.055017709732056
Size (4096): Pull time: 6.701773643493652
Size (2048): Push time: 16.456724643707275
Size (2048): Pull time: 7.375733375549316
Size (1024): Push time: 32.48330855369568
Size (1024): Pull time: 120.822101354599
```
```
636MB Linux ISO
Size (Default): Push time: 19.224534511566162
Size (Default): Pull time: 7.31620454788208
Size (4096): Push time: 55.20979642868042
Size (4096): Pull time: 31.22826313972473
Size (2048): Push time: 109.95740556716919
Size (2048): Pull time: 49.61593508720398
Size (1024): Push time: 219.5595223903656
Size (1024): Pull time: 561.242030620575
```
So unfortunately it does seem rather environment dependent to be able to pick a "optimum" value.

douglas-raillard-arm:
I guess that settles the matter then. We can expose a parameter on Target.push/pull to use a target tempfile, the same way we do for sudo so people can opt in if they know they have the issue, but it's probably better to not change anything else.
Big sysfs files are exceptional, so I don't think it's a good idea to systematically use a temp file for them.

douglas-raillard-arm:
Note: This PR adds a `via_temp` parameter to Target.pull(), to work around the issue:
https://github.com/ARM-software/devlib/pull/563

PR #570, opened 2021-08-20
target: Speedup Target.write_value()

douglas-raillard-arm:
Avoid an execute() by doing the check in the same command. This also
allows to return early if the write is fast, and to extend for longer if
the write is slow. The speed at which you can observe a write in sysfs
depends on the backing kernel handlers, so there is a wide variety of
situations.

This provides a 2x speed up when verify=True, as well as it makes it more robust against sluggish content update


devlib/target.py
```diff
@@ -736,12 +736,43 @@ def batch_revertable_write_value(self, kwargs_list):
...
+fi
+'''
+        else:
+            cmd = 'printf "%s" {value} > {path}'
+        cmd = cmd.format(path=quote(path), value=quote(value))
+
+        try:
+            self.execute(cmd, check_exit_code=True, as_root=True)
+        except TargetStableError as e:
+            if e.returncode == 10:
```
douglas-raillard-arm, line 768: This is actually dodgy, there are many sources for this exception and only one will have this attribute set. We sh ok old introduce a subclass that has these attributes, so we can filter it correctly in the except statement

douglas-raillard-arm:
This PR is not actually ready yet, see
https://github.com/ARM-software/devlib/pull/570/files#r693256957

douglas-raillard-arm:
PR updated with new exception types but has not been tested yet

douglas-raillard-arm:
Now tested on localhost and SSH target

devlib/utils/android.py
```diff
@@ -335,6 +335,14 @@ def execute(self, command, timeout=None, check_exit_code=False,
...
             return adb_shell(self.device, command, timeout, check_exit_code,
                              as_root, adb_server=self.adb_server, su_cmd=self.su_cmd)
+        except subprocess.CalledProcessError as e:
+            cls = TargetTransientCalledProcessError if will_succeed else TargetStableCalledProcessError
+            raise cls(
+                e.returncode,
+                command,
+                e.output,
+                e.stderr,
+            )
```
marcbonnici, line 345: I haven't spotted what the problem is yet, but testing with an ADB connection I get the following error:
```
2021-09-28 10:17:40,224 ERROR        signal:             self.target.grant_package_permission(self.apk_info.package, permission)
2021-09-28 10:17:40,224 ERROR        signal:           File "/mnt/1TBData/marc/Work/devlib/devlib/target.py", line 1717, in grant_package_permission
2021-09-28 10:17:40,224 ERROR        signal:             if 'is not a changeable permission type' in e.message:
2021-09-28 10:17:40,224 ERROR        signal:           File "/mnt/1TBData/marc/Work/devlib/devlib/exception.py", line 27, in message
2021-09-28 10:17:40,224 ERROR        signal:             if self._message is not None:
2021-09-28 10:17:40,224 ERROR        signal:
2021-09-28 10:17:40,226 DEBUG        signal:         Sending error-logged from <ErrorSignalHandler (DEBUG)>
2021-09-28 10:17:40,226 ERROR        signal:         AttributeError('TargetStableCalledProcessError' object has no attribute '_message')
```

devlib/target.py
```diff
@@ -736,12 +736,43 @@ def batch_revertable_write_value(self, kwargs_list):
...
+           cat {path}
+           exit 11
+       fi
+       sleep 0.01
+       trials=$((trials + 1))
+   done
+fi
+'''
+        else:
+            cmd = 'printf "%s" {value} > {path}'
```
marcbonnici, line 762: Not all targets may have `printf`, use the busybox implementation for consistency?

devlib/utils/android.py
douglas-raillard-arm: TargetCalledProcessError inherits its `__init__` from subprocess.CalledProcessError, which means `DevlibError.__init__` is not called. I fixed the issue by dealing with missing `_message` to keep it as close as possible to CalledProcessError.

devlib/target.py
douglas-raillard-arm: I always forget that we don't put busybox symlinks in PATH, will do

douglas-raillard-arm:
@marcbonnici PR updated following your comments

PR #571, opened 2021-09-01
Simplify ftrace

douglas-raillard-arm:
Note: this will probably require updating the trace-cmd binary of devlib, which is quite old now and lacks some options.

trace-cmd 2.9.4 (current released version) should work correctly, as LISA ships a binary compiled from a few commits after 2.9.1 (there were some necessary fixes). That binary in LISA has been used on android as well.


douglas-raillard-arm:
Do you have an working recipe for these binaries ? LISA uses a tool it ships (`lisa-build-asset`) in order to build all the binaries we have for all architectures supported in LISA to put an end to the hours lost fiddling with dependencies and trying to setup the correct build system environment. The binaries it produces are statically linked against the musl libc, in an Alpine container with QEMU userspace emulation to avoid the need for cross toolchain (since half of projects have broken cross compilation).

If that works for you, I can build the binaries using that. However, we only support:
* arm
* arm64
* x86_64

I suppose x86 and ppc64le could be added since Alpine supports them, but I don't know what `armeabi` is compared to plain `arm`.

setrofim:
> Do you have an working recipe for these binaries?

I don't believe we do, but @marcbonnici  can confirm.  Our basic "recipe" for building binaries is to enable all features, except for those that necessitate dynamic linking, and then build, linking statically. The reasoning being that, as long as you link statically, the exact configuration of the build environment should be irrelevant functionally speaking.

> If that works for you, I can build the binaries using that.

Provided LISA configuration works for WA as well, i.e. has all features enabled (as described above), that sounds good. (Note: it's been an absolute age since I've build `trace-cmd` so, atm, I don't recall whether it even has any feature configuration at build time).

> However, we only support:

Ideally, we'd want parity with what is currently included. So if it's possible to add `ppc64le` without too much hassle, please do so. However, the two absolute must-haves are `x86_64` and `arm64` -- as those cover that vast majority of existing usages; however, if there is currently a binary that you're not rebuilding, please delete them part of the same commit, as they would be effectively broken.


 @marcbonnici :  might be worth having a look at `lisa-build-asset` to see if we can adopt something similar for WA/devlib.

douglas-raillard-arm:
Sounds good, I've been able to build all targets except armeabi, since I'm not sure what that is. In the current tree, `arm` only gets a `simpleperf` tool and armeabi gets `trace-cmd` and others, so I guess what is branded as `arm` in lisa maps to `armeabi` in devlib, and `arm` in devlib is historical.

setrofim:
Oh, good point. Yeah `armeabi` should be the same as `arm` (i.e armv7le) ; not sure why there are both now. Incidentally, the `arm` configuration in devlib is currently broken, as it doesn't contain a `busybox` which is essential. We should align those (maybe symlink `armeabi` to `arm`). Either way, that's orthogonal to this pull.

douglas-raillard-arm:
Updated the PR with the binaries, but realized they are not stripped, which should be done for both LISA and devlib. I'll fix that and update again before this gets merged.

douglas-raillard-arm:
PR updated with stripped binaries for all archs that arlready had a trace-cmd binary in devlib

setrofim (approved):

marcbonnici:
> might be worth having a look at lisa-build-asset to see if we can adopt something similar for WA/devlib.

Agreed, this does looks like it would be very useful and a better way of doing things, it's on my todo list to investigate further.

douglas-raillard-arm:
Updated PR to remove f-string

marcbonnici (approved):

PR #572, opened 2021-09-27
target: Fix fallback path for kernel config loading

douglas-raillard-arm:
The main path is reading /proc/config.gz. If it does not exists, the
following paths are tested:
'/boot/config', '/boot/config-$(uname -r)'

Since the 2nd path contains a command to be executed, remove quoting of
the path when using "cat".

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -249,9 +249,9 @@ class Target(object):
         try:
             return KernelConfig(self.execute('zcat /proc/config.gz'))
         except TargetStableError:
-            for path in ['/boot/config', '/boot/config-$(uname -r)']:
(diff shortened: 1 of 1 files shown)


setrofim:
The second path should probably be `'/boot/config-$({} uname -r)'.format(self.busybox)`.

douglas-raillard-arm:
@setrofim Updated the PR. I also switched the order of /boot/config vs /boot/config-$(uname) so that the version-specific one is tested first. This should increase the chances of getting the right config rather than some unspecific one from another installed kernel.

setrofim (approved):

marcbonnici (approved):

PR #581, opened 2022-02-22
android: Fix adb background command PID detection

marcbonnici:
Sometimes the reported PID for a background command would be reported
incorrectly. This was due to an assumption that the lowest PID
was always the command we were looking for, however this is not always
true and PID's could be recycled resulting in the PID of the search
command being reported instead.

To avoid this, filter the reported processes on the the grep command
itself to eliminate matching processes in the system and improve
detection of the PID of the desired background process.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -605,27 +605,14 @@ def adb_background_shell(conn, command,
     if as_root:
         command = 'echo {} | su'.format(quote(command))
 
-    # Attach a unique UUID to the command line so it can be looked for without
-    # any ambiguity with ps
-    uuid_ = uuid.uuid4().hex
-    uuid_var = 'BACKGROUND_COMMAND_UUID={}'.format(uuid_)
-    command = "{} sh -c {}".format(uuid_var, quote(command))
+    command = 'echo $$; sh -c {}'.format(quote(command))
 
     adb_cmd = get_adb_command(device, 'shell', adb_server)
     full_command = '{} {}'.format(adb_cmd, quote(command))
     logger.debug(full_command)
     p = subprocess.Popen(full_command, stdout=stdout, stderr=stderr, stdin=subprocess.PIPE, shell=True)
 
-    # Out of band PID lookup, to avoid conflicting needs with stdout redirection
-    find_pid = '{} ps -A -o pid,args | grep {}'.format(conn.busybox, quote(uuid_var))
-    ps_out = conn.execute(find_pid)
-    pids = [
-        int(line.strip().split(' ', 1)[0])
-        for line in ps_out.splitlines()
-    ]
-    # The line we are looking for is the first one, since it was started before
-    # any look up command
-    pid = sorted(pids)[0]
+    pid = int(p.stdout.readline().strip())
     return (p, pid)
 
 def adb_kill_server(timeout=30, adb_server=None):
--- a/devlib/utils/misc.py
+++ b/devlib/utils/misc.py
@@ -257,6 +257,7 @@ def redirect_streams(stdout, stderr, command):
     def redirect(stream, redirection):
         if stream == subprocess.DEVNULL:
             suffix = '{}/dev/null'.format(redirection)
+            stream = subprocess.PIPE
         elif stream == subprocess.STDOUT:
             suffix = '{}&1'.format(redirection)
             # Indicate that there is nothing to monitor for stderr anymore


devlib/utils/android.py
```diff
@@ -617,15 +617,26 @@ def adb_background_shell(conn, command,
...
+        for line in ps_lines
     ]
-    # The line we are looking for is the first one, since it was started before
-    # any look up command
-    pid = sorted(pids)[0]
+
+    # Handle the case where the command has already exited otherwise
+    # we should only be left with our desired command
+    if not pids:
+        pid = None
```
marcbonnici, line 636: Should we return `None` or something else here to indicate an invalid PID? Perhaps 0 to be type consistent?

douglas-raillard-arm: In my experience, returning None or a made up value rather than raising does a lot more harm than good, since anything that actually needs the PID will not work with the fake value.

I guess we can end up in this situation if the process does not use stdin/stout/stderr and just exits immediately. On the ssh paramiko connection, PID detection is done by adding a print $$ before exec-ing the command, which is maybe the way to go.

douglas-raillard-arm:
Was the problem really because of PID reuse or just that if the command terminates before the lookup of its PID, the current code would just get the PID of the grep command ?

In any case, that's a good catch, thanks for debugging that.

marcbonnici:
> Was the problem really because of PID reuse or just that if the command terminates before the lookup of its PID, the current code would just get the PID of the grep command?

Perhaps reuse might not apply in this case and there was some delay introduce in spawning the initial command instead, however the result was the reported PID for the background command was returning the PID of the command used to find the PID rather than the background progress itself e.g.

```
In [13]: bg = ta.background("sleep 10")
    ...: print('PID: {}.format(bg.pid))
    ...: print('PID Found ', str(bg.pid) in ta.execute("ps"))

23534 /system/bin/sh -c (/data/local/tmp/bin/busybox ps -A -o pid,args | grep BACKGROUND_COMMAND_UUID=704c9a73dabc46e2af5da97d0abe6b61); echo "\n$?"
23536 /system/bin/sh -c BACKGROUND_COMMAND_UUID=704c9a73dabc46e2af5da97d0abe6b61 sh -c 'sh -c '"'"'export PATH=/data/local/tmp/bin:$PATH && export LC_ALL=C && sleep 1'"'"'  '
23539 grep BACKGROUND_COMMAND_UUID=704c9a73dabc46e2af5da97d0abe6b61

PID: 23534  # Should be: 23536 
PID Found: False  

```

This explains the "Process Not found" errors when trying to kill the processes as the grep had already completed, along with the leftover sleep processes preventing progress.

devlib/utils/android.py
marcbonnici: > In my experience, returning None or a made up value rather than raising does a lot more harm than good, since anything that actually needs the PID will not work with the fake value.

That's a fair point. My concern over explicitly raising an error is that a usecase I can see for backgrounding a command is that you just want it to execute and don't care if its a short or long running task. In the case that it is short running and we start raising errors this might break existing flows that don't care about the PID. 

> On the ssh paramiko connection, PID detection is done by adding a print $$ before exec-ing the command, which is maybe the way to go.

That seems like a good idea, I'll try switching to that mechanism so that we can try and avoid this issue.

douglas-raillard-arm: > In the case that it is short running and we start raising errors this might break existing flows that don't care about the PID.

I guess exceptions would have to be handled inside devlib mostly, e.g. when getting the PID to kill the command (although I think we would not do anything if the command is already finished). In any case, having a reliable PID would be a better if possible at all. It might complicate things in case of stream redirections though.

marcbonnici: Agreed I think getting the reliable PID detection is the best way forward here. I've updated the PR with an attempt to align with the mechanism the SSH connection uses. From what I can tell the stream redirection was actually already taking place in this connection as well however the `redirect_streams` function was missing the modification to the returned stream. Please let me know if these changes look reasonable to you.

devlib/utils/android.py
```diff
@@ -605,27 +605,14 @@ def adb_background_shell(conn, command,
     if as_root:
         command = 'echo {} | su'.format(quote(command))
 
-    # Attach a unique UUID to the command line so it can be looked for without
-    # any ambiguity with ps
-    uuid_ = uuid.uuid4().hex
-    uuid_var = 'BACKGROUND_COMMAND_UUID={}'.format(uuid_)
-    command = "{} sh -c {}".format(uuid_var, quote(command))
+    command = 'echo $$; sh -c {}'.format(quote(command))
```
douglas-raillard-arm, line 608: That's minor but we could use `echo $$; exec sh -c` to avoid a layer

devlib/utils/misc.py
```diff
@@ -257,6 +257,7 @@ def redirect_streams(stdout, stderr, command):
     def redirect(stream, redirection):
         if stream == subprocess.DEVNULL:
             suffix = '{}/dev/null'.format(redirection)
+            stream = subprocess.PIPE
         elif stream == subprocess.STDOUT:
```
douglas-raillard-arm, line 261: We might need to double check the use of redirect_stream in ssh.py as well. Once the pipe is setup, it has to be consumed and closed properly. For Popen uses, it should be ok as Popen's context manager closes the streams on `__exit__()`.

douglas-raillard-arm: Also we might want to do that only for stdout, since stderr should still be DEVNULL as we don't print anything to it.

devlib/utils/android.py
douglas-raillard-arm, line 608: `echo` is unreliable on some shells, should be `printf "%s\n" $$`

douglas-raillard-arm:
@marcbonnici I've updated my async PR with a patch that fixes that without needing to deal with complex redirections: https://github.com/ARM-software/devlib/pull/569/commits/b9671948d4a48912db009bdbf5af76e541092d75

PR #585, opened 2022-04-01
Install busybox

douglas-raillard-arm:
Use busybox symlinks to simplify uses of busybox tools in devlib.

Note that it is currently rebased on https://github.com/ARM-software/devlib/pull/569 since it touches some of the places updated by that other PR. It is however relatively independent.

TODO:
* fix the parsing of `ps` output, that used to by the system's `ps` output and is now busybox ones.

### DIFF
--- a/devlib/collector/ftrace.py
+++ b/devlib/collector/ftrace.py
@@ -28,6 +28,7 @@ from devlib.collector import (CollectorBase, CollectorOutput,
 from devlib.host import PACKAGE_BIN_DIRECTORY
 from devlib.exception import TargetStableError, HostError
 from devlib.utils.misc import check_output, which, memoized
+from devlib.utils.asyn import asyncf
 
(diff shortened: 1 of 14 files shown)


douglas-raillard-arm:
Actually I'm now wondering whether this is a good idea. It simplifies some code but also prevents the user from choosing between the buxybox and whatever is provided by the system. We surely don't want to rely on system binaries in devlib code itself (modulo a couple of exceptions), but user's execute() might want that.

douglas-raillard-arm:
>In which case perhaps we don't progress with this option for now unless we find a strong benefit to do so in the future?

yes let's close this PR for now. Maybe in the future we could have something like `execute_busybox()` that adds busybox symlinks to PATH or something like that if we really want it, or maybe as an extra option flag to `execute()/background()`.

PR #587, opened 2022-04-08
Fix android bg

douglas-raillard-arm:
Splitting this change from the async PR as the async PR no longer relies on background():
https://github.com/ARM-software/devlib/pull/569


douglas-raillard-arm:
@marcbonnici it looks like the branch was merged in `master` but this PR is still open, have you done it manually and should we just close the PR ?

marcbonnici:
Ah it looks like these changes was accidentally merged as part of #639 so it looks like we can close this for now.

PR #603, opened 2022-08-12
Fix adb root

douglas-raillard-arm:
This PR solves the freeze observed on some Pixel 6 root images when connecting to the target. The freeze appears to be related to some race condition in adb server when rooting/unrooting it.

Note that this PR needs a final round of validation on the actual device before being merged.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -30,6 +30,7 @@ import tempfile
 import time
 import uuid
 import zipfile
+import threading
 
 from collections import defaultdict
 from io import StringIO
@@ -258,7 +259,7 @@ class AdbConnection(ConnectionBase):
 
     # maintains the count of parallel active connections to a device, so that
     # adb disconnect is not invoked untill all connections are closed
-    active_connections = defaultdict(int)
+    active_connections = (threading.Lock(), defaultdict(int))
     # Track connected as root status per device
     _connected_as_root = defaultdict(lambda: None)
     default_timeout = 10
@@ -301,7 +302,10 @@ class AdbConnection(ConnectionBase):
                             'poll_period': transfer_poll_period,
                             }
         self.transfer_mgr = PopenTransferManager(self, **transfer_opts) if poll_transfers else None
-        AdbConnection.active_connections[self.device] += 1
+        lock, nr_active = AdbConnection.active_connections
+        with lock:
+            nr_active[self.device] += 1
+
         if self.adb_as_root:
             try:
                 self.adb_root(enable=True)
@@ -377,12 +381,17 @@ class AdbConnection(ConnectionBase):
         return bg_cmd
 
     def _close(self):
-        AdbConnection.active_connections[self.device] -= 1
-        if AdbConnection.active_connections[self.device] <= 0:
+        lock, nr_active = AdbConnection.active_connections
+        with lock:
+            nr_active[self.device] -= 1
+            disconnect = nr_active[self.device] <= 0
+            if disconnect:
+                del nr_active[self.device]
+
+        if disconnect:
             if self.adb_as_root:
                 self.adb_root(enable=False)
             adb_disconnect(self.device, self.adb_server)
-            del AdbConnection.active_connections[self.device]
 
     def cancel_running_command(self):
         # adbd multiplexes commands so that they don't interfer with each
@@ -391,7 +400,11 @@ class AdbConnection(ConnectionBase):
         pass
 
     def adb_root(self, enable=True):
-        if AdbConnection.active_connections[self.device] > 1:
+        lock, nr_active = AdbConnection.active_connections
+        with lock:
+            can_root = nr_active[self.device] <= 1
+
+        if not can_root:
             raise AdbRootError('Can only restart adb server if no other connection is active')
 
         cmd = 'root' if enable else 'unroot'
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -40,7 +40,7 @@ try:
 except ImportError:
     from pipes import quote
 
-from devlib.exception import TargetTransientError, TargetStableError, HostError, TargetTransientCalledProcessError, TargetStableCalledProcessError
+from devlib.exception import TargetTransientError, TargetStableError, HostError, TargetTransientCalledProcessError, TargetStableCalledProcessError, AdbRootError
 from devlib.utils.misc import check_output, which, ABI_MAP, redirect_streams, get_subprocess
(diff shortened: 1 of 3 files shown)


devlib/utils/android.py
```diff
@@ -377,12 +381,16 @@ def _background(self, command, stdout, stderr, as_root):
         return bg_cmd
 
     def _close(self):
-        AdbConnection.active_connections[self.device] -= 1
-        if AdbConnection.active_connections[self.device] <= 0:
+        lock, nr_active = AdbConnection.active_connections
+        with lock:
+            nr_active[self.device] -= 1
+            disconnect = nr_active[self.device] <= 0
+            del nr_active[self.device]
```
marcbonnici, line 388: Why is this moved here? Will this not delete the entry even if there are still connections left?

devlib/utils/android.py
```diff
@@ -301,10 +301,17 @@ def __init__(self, device=None, timeout=None, platform=None, adb_server=None,
...
         if self.adb_as_root:
-            self.adb_root(enable=True)
+            try:
+                self.adb_root(enable=True)
+            # Exception will be raised if we are not the only connection
+            # active. adb_root() requires restarting the server, which is not
+            # acceptable if other connections are active and can apparently
+            # lead to commands hanging forever in some situations.
+            except RuntimeError:
+                pass
```
marcbonnici, line 317: Would it make the intention clearer here to only try and enable adb_root if the connection is not greater than 1? This would prevent hiding any other runtime errors that may occur?

devlib/utils/android.py
douglas-raillard-arm: Should be moved a line lower to be inside the `if` indeed

devlib/utils/android.py
douglas-raillard-arm: I did not want to duplicate the checking logic as adb_root() is also exposed to the user, in order to use this path on a typical run (and therefore avoid bitrot). Maybe I should make a more specific exception type though.

douglas-raillard-arm:
@marcbonnici Updated with:
* AdbRootError exception to avoid a  generic RuntimeError
* fixed the active devices count.

marcbonnici:
I think this is unrelated to these changes but something that showed up during testing. If I create an AndroidTarget and then try to enable adb_root it will fail due to other connections being active (`max_async + 1`).

```
t = AndroidTarget()
t.conn.adb_root(True)
> AdbRootError: Can only restart adb server if no other connection is active
t.conn.active_connections
> (<unlocked _thread.lock object at 0x7f3bbde1e1e0>,
> defaultdict(int, {'emulator-5554': 51}))
```

It looks like once the connections are created during the benchmark stage they are not closed correctly. Do you see the same behavior on your setup?

douglas-raillard-arm:
> It looks like once the connections are created during the benchmark stage they are not closed correctly

It's on purpose: https://github.com/ARM-software/devlib/blob/492284f46dcb3b9a4633f6c284c76d36a20e8ba2/devlib/target.py#L446
Establishing a connection can be somewhat costly for e.g. ssh so that avoids re-doing that extra work, which would be needed anyway. 

Wrt to adb_root(), I think it's an historical accident to have exposed that. It's really a property of the target and not any individual connection as it affects the whole server, as the "active connections" class attribute dict hack shows. On top of that, it's something that can only be safely done when the Target is created as it will interrupt brutally any sort of background command or other thread using the Target. Ideally, we should probably:
* deprecate it and force people to state what they want in Target ctor
* Only allow adb_root(enable=True/False) when it's a no-op wrt to what was setup when the Target was created, raise otherwise.
* Alternatively, turn adb_root() into a simple flag internally and use `su` or `sudo` to emulate it on a per-command basis (possibly a static build ?)

marcbonnici:
> Establishing a connection can be somewhat costly for e.g. ssh so that avoids re-doing that extra work, which would be needed anyway.

Right thanks for the reminder, that makes sense. I guess for ADB connections this is less important but best to keep the consistency if we can. 

> Wrt to adb_root(), I think it's an historical accident to have exposed that. It's really a property of the target and not any individual connection as it affects the whole server, as the "active connections" class attribute dict hack shows.

I think there's an argument for both sides here, as this is an ADB connection specific functionality rather than a target itself. The discrepancy is as you say this affects all connections rather than individual ones. 
Having said that I don't generally expect direct modifications to the connection itself, tbh I'd almost treat this as an internal method, as this bypassed the abstraction layer we're trying to present for runtime and therefore relies on users knowing what they're doing so I don't think this should be a problem (just confirming internally). I just wanted to ensure that this behaviour was expected.

>  Ideally, we should probably:
    deprecate it and force people to state what they want in Target ctor
    Only allow adb_root(enable=True/False) when it's a no-op wrt to what was setup when the Target was created, raise otherwise.
    Alternatively, turn adb_root() into a simple flag internally and use su or sudo to emulate it on a per-command basis (possibly a static build ?)

I'm not sure this would benefit us much in terms of usability, I think it might be better / provide a more consistent experience to just raise an error message, as it currently does. 
If we allow it as a no-op this could lead to inconsistent behaviour depending on the original state of the device and also we may not always have su etc. installed on the device so I think it would be best to inform the user of the issue explicitly and let them decide on the best way to solve the problem, i.e. ensure that this is set when the connection is first established.

douglas-raillard-arm:
So I guess the rule is more or less that if someone instantiate a connection instance by hand, they are free to use adb_root() and it is somewhat a public API, but if the connection is managed of a Target, it becomes private. Ideally we would need a 3rd object on top of Target + Connection that would be managed by the Connection but attached to the Target. But since it's the only case where that is needed, the mutable class attr hack is probably good enough for now.

Using `su` to emulate adb root would probably open a whole can of worms anyway (is it even possible to install a binary without rooting the device in the first place ?).

marcbonnici:
> So I guess the rule is more or less that if someone instantiate a connection instance by hand, they are free to use adb_root() and it is somewhat a public API, but if the connection is managed of a Target, it becomes private. Ideally we would need a 3rd object on top of Target + Connection that would be managed by the Connection but attached to the Target. But since it's the only case where that is needed, the mutable class attr hack is probably good enough for now.

Agreed, that seems like a reasonable summary to me.

> Using su to emulate adb root would probably open a whole can of worms anyway (is it even possible to install a binary without rooting the device in the first place ?).

Most definitely agree, for the majority of cases I think we'd be out of luck, however in any case I think that would be something that would need to be solved on a per device basis rather than something we could abstract reliably.

marcbonnici (approved):

PR #604, opened 2022-08-16
devlib.target: Fix AndroidTarget unpickle

douglas-raillard-arm:
Fix __setstate__ to call super().__setstate__ in order to handle the
generic part of the deserialization.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1798,7 +1798,7 @@ class AndroidTarget(Target):
         }
 
     def __setstate__(self, dct):
-        self.__dict__.update(dct)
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
Without that, unpickling an AndroidTarget gives back a broken object missing the `_unused_conns` attribute.

That pickling/unpickling is used in LISA to transfer a target to a separate process, which is unfortunately necessary in order to run code in a separate user and mount namespace used to mount overlayfs as a regular user.

marcbonnici (approved):

PR #605, opened 2022-08-23
target: Add Target.revertable_write_value(may_not_exist=False)

douglas-raillard-arm:
Add a may_not_exist parameter to Target.revertable_write_value() that
will turn the write into a no-op if the file does not exist or cannot be
read. If may_not_exist is a callable, it will be called with the
parameters of the function. This can be used to e.g. log the issue at a
higher logging level.

### DIFF
--- a/doc/target.rst
+++ b/doc/target.rst
@@ -398,10 +398,16 @@ Target
        some sysfs entries silently failing to set the written value without
        returning an error code.
 
-.. method:: Target.revertable_write_value(path, value [, verify])
+.. method:: Target.revertable_write_value(path, value [, verify [, may_not_exist]])
 
    Same as :meth:`Target.write_value`, but as a context manager that will write
    back the previous value on exit.
+
+   :param may_not_exist: If the file does not exist, the context manager will
+      silently do nothing, or call ``may_not_exist`` if it is a callable. It
+      will be passed the exception, the target, the ``path``, the ``value`` and
+      ``verify``.
+   :type may_not_exist: bool or collections.abc.Callable
 
 .. method:: Target.batch_revertable_write_value(kwargs_list)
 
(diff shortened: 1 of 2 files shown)


doc/target.rst
```diff
@@ -398,11 +398,17 @@ Target
...
        returning an error code.
 
-.. method:: Target.revertable_write_value(path, value [, verify])
+.. method:: Target.revertable_write_value(path, value [, verify [, may_not_exist]])
 
    Same as :meth:`Target.write_value`, but as a context manager that will write
    back the previous value on exit.
 
+   :param may_not_exist: If the file does not exist, the context manager will
+      silently do nothing, or call ``may_not_exist`` if it is a callable. It
```
marcbonnici, line 407: Would you be able to elaborate a little on this use case for this? 
It seems a little strange that `may_not_exist` could be a boolean or a callable. I was wondering if it would make more sense to leave it to the caller to handle in case of an error, or whether these could be 2 separate parameters? One to allow the operation to not error, and another as an error handler?

douglas-raillard-arm: The use case is to write a debugfs file triggering a behavior in the kernel (changing niceness of workqueue kworkers) for the duration of a test. The issue is that the file may not exist as it's brought by a non-mainline patch. If it does not exist it's not the end of the world but we might want to do explicit logging to warn the user.

> leave it to the caller to handle in case of an error

That is unfortunately not possible unless you call manually `__enter__` and `__exit__`.

> 2 separate parameters

We could but this opens the door to combinations that do not make sense (a callback and may_not_exist=False). Combining both ensures consistency.

That said, I just realized that this could be done as a context-manager combinator rather than handled for each specific context manager so I'll probably try that approach first, e.g.:
```
with suppress_excep(TargetStableError, target.revertable_write_value(...), possible_callback):
   ...
```

Would you be interested in a PR for that in devlib or should I just add it to LISA's misc utils module ? I don't mind either way.

douglas-raillard-arm: EDIT: Here is the LISA PR to introduce such a combinator: 
https://github.com/ARM-software/lisa/pull/1883

PR #618, opened 2023-02-06
devlib.collector: Add PerfettoCollector

mrkajetanp:
Add a Collector for accessing Google's Perfetto tracing infrastructure. The Collector takes a path to an on-device config file, starts tracing in the background using the perfetto binary and then stops by killing the tracing process.

Signed-off-by: Kajetan Puchalski <kajetan.puchalski@arm.com>

### DIFF
--- a/devlib/collector/perfetto.py
+++ b/devlib/collector/perfetto.py
@@ -1,0 +1,120 @@
+#    Copyright 2023 ARM Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import os
+import subprocess
+from shlex import quote
+
+from devlib.host import PACKAGE_BIN_DIRECTORY
+from devlib.collector import (CollectorBase, CollectorOutput,
+                              CollectorOutputEntry)
+from devlib.exception import TargetStableError, HostError
+
+OUTPUT_PERFETTO_TRACE = 'devlib-trace.perfetto-trace'
+
+
+class PerfettoCollector(CollectorBase):
+    """
+    Perfetto is a production-grade open-source stack for performance instrumentation
+    and trace analysis developed by Google. It offers services and libraries for
+    recording system-level and app-level traces, native + java heap profiling,
+    a library for analyzing traces using SQL and a web-based UI to visualize and
+    explore multi-GB traces.
+
+    This collector takes a path to a perfetto config file saved on disk and passes
+    it directly to the tool.
+
+    On Android platfroms Perfetto is included in the framework starting with Android 9.
+    On Android 8 and below, follow the Linux instructions below to build and include
+    the standalone tracebox binary.
+
+    On Linux platforms, either traced (Perfetto tracing daemon) needs to be running
+    in the background or the tracebox binary needs to be built from source and placed
+    in the Package Bin directory. The build instructions can be found here:
+
+    It is also possible to force using the prebuilt tracebox binary on platforms which
+    already have traced running using the force_tracebox collector parameter.
+
+    https://perfetto.dev/docs/contributing/build-instructions
+
+    After building the 'tracebox' binary should be copied to devlib/bin/<arch>/.
+
+    For more information consult the official documentation:
+    https://perfetto.dev/docs/
+    """
+
+    def __init__(self, target, config=None, force_tracebox=False):
+        super().__init__(target)
+        self.bg_cmd = None
+        self.config = config
+        self.target_binary = 'perfetto'
+        target_output_path = self.target.working_directory
+
+        install_tracebox = force_tracebox or (target.os in ['linux', 'android'] and not target.is_running('traced'))
+
+        # Install Perfetto through tracebox
+        if install_tracebox:
+            self.target_binary = 'tracebox'
+            if not self.target.get_installed(self.target_binary):
+                host_executable = os.path.join(PACKAGE_BIN_DIRECTORY,
+                                               self.target.abi, self.target_binary)
+                if not os.path.exists(host_executable):
+                    raise HostError("{} not found on the host".format(self.target_binary))
+                self.target.install(host_executable)
+        # Use Android's built-in Perfetto
+        elif target.os == 'android':
+            os_version = target.os_version['release']
+            if int(os_version) >= 9:
+                # Android requires built-in Perfetto to write to this directory
+                target_output_path = '/data/misc/perfetto-traces'
+                # Android 9 and 10 require traced to be enabled manually
+                if int(os_version) <= 10:
+                    target.execute('setprop persist.traced.enable 1')
+
+        self.target_output_file = target.path.join(target_output_path, OUTPUT_PERFETTO_TRACE)
+
+    def start(self):
+        cmd = "cat {} | {} --txt -c - -o {}".format(
+            quote(self.config), quote(self.target_binary), quote(self.target_output_file)
+        )
+        # start tracing
+        if self.bg_cmd is None:
+            self.bg_cmd = self.target.background(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        else:
+            raise TargetStableError('Perfetto collector is not re-entrant')
+
+    def stop(self):
+        # stop tracing
+        self.bg_cmd.cancel()
+        self.bg_cmd = None
+
+    def set_output(self, output_path):
+        if os.path.isdir(output_path):
+            output_path = os.path.join(output_path, os.path.basename(self.target_output_file))
+        self.output_path = output_path
+
+    def get_data(self):
+        if self.output_path is None:
+            raise RuntimeError("Output path was not set.")
+        if not self.target.file_exists(self.target_output_file):
+            raise RuntimeError("Output file not found on the device")
+        self.target.pull(self.target_output_file, self.output_path)
+        output = CollectorOutput()
+        if not os.path.isfile(self.output_path):
+            self.logger.warning('Perfetto trace not pulled from device.')
+        else:
+            output.append(CollectorOutputEntry(self.output_path, 'file'))
+        return output
+
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -54,7 +54,8 @@ from devlib.platform import Platform
 from devlib.exception import (DevlibTransientError, TargetStableError,
                               TargetNotRespondingError, TimeoutError,
                               TargetTransientError, KernelConfigKeyError,
-                              TargetError, HostError, TargetCalledProcessError) # pylint: disable=redefined-builtin
+                              TargetError, HostError, TargetCalledProcessError,
+                              TargetStableCalledProcessError)  # pylint: disable=redefined-builtin
 from devlib.utils.ssh import SshConnection
 from devlib.utils.android import AdbConnection, AndroidProperties, LogcatMonitor, adb_command, adb_disconnect, INTENT_FLAGS
 from devlib.utils.misc import memoized, isiterable, convert_new_lines, groupby_value
@@ -278,6 +279,12 @@ class Target(object):
         if self._shutils is None:
             self._setup_shutils()
         return self._shutils
+
+    def is_running(self, comm):
+        cmd_ps = f'''{self.busybox} ps -A -T -o stat,comm'''
+        cmd_awk = f'''{self.busybox} awk 'BEGIN{{found=0}} {{state=$1; $1=""; if ($state != "Z" && $0 == " {comm}") {{found=1}}}} END {{print found}}' '''
+        result = self.execute(f"{cmd_ps} | {cmd_awk}")
+        return bool(int(result))
 
     @tls_property
     def _conn(self):
--- a/devlib/__init__.py
+++ b/devlib/__init__.py
@@ -46,6 +46,7 @@ from devlib.derived.energy import DerivedEnergyMeasurements
 from devlib.derived.fps import DerivedGfxInfoStats, DerivedSurfaceFlingerStats
 
 from devlib.collector.ftrace import FtraceCollector
+from devlib.collector.perfetto import PerfettoCollector
 from devlib.collector.perf import PerfCollector
 from devlib.collector.serial_trace import SerialTraceCollector
 from devlib.collector.dmesg import DmesgCollector


devlib/collector/perfetto.py
```diff
@@ -0,0 +1,72 @@
...
+            output_path = os.path.join(output_path,
+                                       os.path.basename(self.target_output_file))
+        self.output_path = output_path
+
+    def get_data(self):
+        if self.output_path is None:
+            raise RuntimeError("Output path was not set.")
+        # adjust the pull timeout based on the size of the file
+        pull_timeout = 10 * (self.stop_time - self.start_time)
+        self.target.pull(self.target_output_file, self.output_path, timeout=pull_timeout)
```
marcbonnici, line 65: Do we need to provide a timeout here? If not specified devlib should fall back to using the polling mechanism which *should* be more reliable. 
Have you seen behavior contrary to this?

mrkajetanp: That's what the FtraceCollector does in get_data() so I assumed there was a reason there and did the same for consistency

marcbonnici: Ah, I guess that means the FtraceCollector has not been updated since this functionality has been added. Thanks for pointing that out, we should probably get that updated as well.

marcbonnici: Yep, I think that would be good thanks.

devlib/collector/perfetto.py
```diff
@@ -0,0 +1,68 @@
...
+        self.output_path = None
+        self.config = config
+
+        if not self.target.is_installed('perfetto'):
+            raise TargetStableError('No perfetto found on device.')
+        self.target_binary = 'perfetto'
+
+    def start(self):
+        # start tracing and save the PID to stop it later
+        self.pid = self.target.execute(
```
douglas-raillard-arm, line 41: It may be a good idea to use the target.background() and the BackgroundCommand API to handle that instead of re-growing an ad-hoc version of it:

```
def start(self):
    ...
    if self.bg_cmd is None:
        self.bg_cmd = self.target.background(cmd)
    else:
        raise TargetStableError('Perfetto collector is not re-entrant'')
    ...

def stop(self):
    ...
    self.bg_cmd.kill()
    self.bg_cmd = None
    ...
```

devlib/collector/perfetto.py
```diff
@@ -0,0 +1,68 @@
...
+            "cat {} | {} --background --txt -c - -o {}".format(
```
douglas-raillard-arm, line 42: paths need to be escaped using shlex.quote() to avoid typical shell interpolation issues

devlib/collector/perfetto.py
```diff
@@ -0,0 +1,68 @@
...
+from devlib.utils.misc import check_output, which, memoized
+
+OUTPUT_PERFETTO_TRACE = 'trace.perfetto-trace'
+
+class PerfettoCollector(CollectorBase):
+
+    def __init__(self, target, config=None):
+        super(PerfettoCollector, self).__init__(target)
+        self.target_output_file = target.path.join(self.target.working_directory, OUTPUT_PERFETTO_TRACE)
+        self.output_path = None
```
douglas-raillard-arm, line 32: already done in `CollectorBase.__init__`

devlib/collector/perfetto.py
```diff
@@ -0,0 +1,68 @@
...
+                              CollectorOutputEntry)
+from devlib.exception import TargetStableError, HostError
+from devlib.utils.misc import check_output, which, memoized
+
+OUTPUT_PERFETTO_TRACE = 'trace.perfetto-trace'
+
+class PerfettoCollector(CollectorBase):
+
+    def __init__(self, target, config=None):
+        super(PerfettoCollector, self).__init__(target)
```
douglas-raillard-arm, line 30: `super(PerfettoCollector, self)` can be shortened into `super()` in Python 3

douglas-raillard-arm (commented):
Some drive-by comments

mrkajetanp:
I updated the PR with Douglas' suggestions.
I also had to bring back the manual pull timeout bit. I noticed that without it I'd end up getting a pull timeout exception when trying to pull the trace from the device. No idea why but having the manual timeout fixes the problem.

douglas-raillard-arm:
> I noticed that without it I'd end up getting a pull timeout exception when trying to pull the trace from the device.

That's strange, from a cursory look I could not find any non-None default timeout on the pull() path, except in the transfer manager but it's set to 3600

douglas-raillard-arm:
I would definitely use background command. The previous version of the code was straightfoward and yet already had issues:
1. if the process dies, we would still attempt to kill an old PID even though it might point at an entirely different task. Android can exhaust PIDs pretty quickly.
2. it allowed starting perfetto twice, which would result in failures down the line
3. less likely, but if perfetto hanged at SIGINT/SIGTERM, devlib would freeze. With BackgroundCommand.cancel(), SIGTERM is sent and then SIGKILL after a timeout.
4. if devlib disconnects, perfetto will be left running. Background commands are terminated properly when the target is destroyed or disconnected.

The change to the background command API fixed 1. and made 2. obvious. On top of that, we can now capture stdout/stderr if required for debug, instead of having to go hunt for some log file somewhere. Process control is a tricky business and it's definitely the sort of stuff we don't want to duplicate. Ctrl-C in a terminal is simply SIGINT, which will likely be handled like SIGTERM so I expect perfetto to work fine with that.

devlib/collector/perfetto.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if not self.target.is_installed('perfetto'):
+            raise TargetStableError('No perfetto found on device.')
+        self.target_binary = 'perfetto'
+
+    def start(self):
+        cmd = "cat {} | {} --txt -c - -o {}".format(
+            quote(self.config), quote(self.target_binary), quote(self.target_output_file)
+        )
+        self.start_time = time.time()
```
douglas-raillard-arm, line 44: We need to figure out why the pull times out as it's not really expected to happen instead of this hack. Also time.time() is not monotonic so `end - start < 0` is possible

devlib/collector/perfetto.py
marcbonnici: @mrkajetanp Do you have any logs of the runs where timeouts were occuring to take a look at?

devlib/collector/perfetto.py
mrkajetanp: It's not even that it times out, more that there's an issue if we don't specify an explicit timeout?
```
> /home/kajpuc01/tools/lisa/external/devlib/devlib/collector/perfetto.py(69)get_data()
-> self.target.pull(self.target_output_file, self.output_path)
(Pdb) s
--Call--
> /home/kajpuc01/tools/lisa/external/devlib/devlib/utils/asyn.py(202)__get__()
-> def __get__(self, *args, **kwargs):
(Pdb) n
> /home/kajpuc01/tools/lisa/external/devlib/devlib/utils/asyn.py(203)__get__()
-> return self.__class__(
(Pdb)
> /home/kajpuc01/tools/lisa/external/devlib/devlib/utils/asyn.py(204)__get__()
-> asyn=self.asyn.__get__(*args, **kwargs),
(Pdb)
> /home/kajpuc01/tools/lisa/external/devlib/devlib/utils/asyn.py(205)__get__()
-> blocking=self.blocking.__get__(*args, **kwargs),
(Pdb)
> /home/kajpuc01/tools/lisa/external/devlib/devlib/utils/asyn.py(203)__get__()
-> return self.__class__(
(Pdb)
--Return--
> /home/kajpuc01/tools/lisa/external/devlib/devlib/utils/asyn.py(203)__get__()-><devlib.utils...x7f326301bc10>
-> return self.__class__(
(Pdb)
--Call--
> /home/kajpuc01/tools/lisa/external/devlib/devlib/utils/asyn.py(208)__call__()
-> def __call__(self, *args, **kwargs):
(Pdb)
> /home/kajpuc01/tools/lisa/external/devlib/devlib/utils/asyn.py(209)__call__()
-> return self.blocking(*args, **kwargs)
(Pdb)
WARNING          Cancelling file transfer ['/sdcard/devlib-target/trace.perfetto-trace'] -> perfetto_test/wk1-speedometer-1/trace.perfetto-trace due to 'transfer inactive'
WARNING          Cancelling file transfer ['/sdcard/devlib-target/trace.perfetto-trace'] -> perfetto_test/wk1-speedometer-1/trace.perfetto-trace due to 'exception during transfer'
TimeoutError: adb -s 18251FDF60083A pull /sdcard/devlib-target/trace.perfetto-trace perfetto_test/wk1-speedometer-1/trace.perfetto-trace
```

I think the issue might be here:
```
    def _push_pull(self, action, sources, dest, timeout):
        # [...]
        if timeout or not self.poll_transfers:
            adb_command(self.device, command, timeout=timeout, adb_server=self.adb_server)
        else:
            with self.transfer_mgr.manage(sources, dest, action):
                bg_cmd = adb_command_background(self.device, command, adb_server=self.adb_server)
                self.transfer_mgr.set_transfer_and_wait(bg_cmd)
```

Depending on whether a timeout was specified we either use `adb_command` or `adb_command_background`. The former seems to work fine, the latter not so much in this case.

mrkajetanp:
I dropped the explicit timeout argument but that requires #619 so it should be either merged after the other one or at the same time

marcbonnici (approved):

mrkajetanp:
Since the transfer manager PR has been merged we should now be able to go ahead and merge this as well?

marcbonnici (commented):
Apologies for the delay yes we can proceed with this PR. I just wanted to spend some time understand this a bit and get it running on my end as I had found some issues during testing but it turns out they were unrelated to this PR. (Opened a PR to address these https://github.com/ARM-software/devlib/pull/628)  

However I have a couple of questions for you.

IIUC running the perfetto binary requires a running service on the device to connect to, and if this is not running then currently we fail silently. On a linux target (my test device) this is provided by `traced` binary or both can packaged as part of the `tracebox` binary.

I'm wondering is you know an easy way to detect whether this service is running, or even adding some config to enable running the service as a background command as part of the collector (I'm currently testing with manually running with `./tracebox traced` before starting a workload.

Secondly do you know if we would be able to include a compiled perfetto binary (or tracebox) similar to what we do for trace-cmd etc to make it easier for users to get started?

devlib/collector/perfetto.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def set_output(self, output_path):
+        if os.path.isdir(output_path):
+            output_path = os.path.join(output_path, os.path.basename(self.target_output_file))
+        self.output_path = output_path
+
+    def get_data(self):
+        if self.output_path is None:
+            raise RuntimeError("Output path was not set.")
+        self.target.pull(self.target_output_file, self.output_path)
```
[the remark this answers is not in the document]
mrkajetanp: sure, done

mrkajetanp:
First of all, these are not issues for Android - on Android Perfetto always comes included and is always running so it'll just be fine. For Linux targets there's a bit more steps, true.

> I'm wondering is you know an easy way to detect whether this service is running
I'm not sure if there's any official or proper way to check, we could always grep the `ps` output or something like that if we really wanted to?

> or even adding some config to enable running the service as a background command
We could just make the tool use `tracebox` if it can't find the running service on Linux?

> Secondly do you know if we would be able to include a compiled perfetto binary (or tracebox) similar to what we do for trace-cmd etc to make it easier for users to get started?

As mentioned above, it comes bundled with Android so on Android there's no point. On Linux if the tracebox binary already works for you then I see no issues with bundling it.

https://perfetto.dev/docs/quickstart/linux-tracing

devlib/collector/perfetto.py
```diff
@@ -0,0 +1,72 @@
...
+from shlex import quote
+
+from devlib.collector import (CollectorBase, CollectorOutput,
+                              CollectorOutputEntry)
+from devlib.exception import TargetStableError, HostError
+from devlib.utils.misc import check_output, which, memoized
+
+OUTPUT_PERFETTO_TRACE = 'trace.perfetto-trace'
+
+class PerfettoCollector(CollectorBase):
```
marcbonnici, line 28: Can we add a doc string to give a quick overview of what this collector adds and perhaps links to the getting started guidance?
Also do you know what version of Android added support for this as that would also be useful to know?

marcbonnici (commented):
I see, thanks for the information. I understand that the requirements for Android and Linux are different, one option would be to check for the OS of the target and either auto deploy the binary and start it as a background task for Linux or assume it is present for Android?

At the minimum I think we should outline what needs to be done in the docsctring for the collector. Additionally it might be worth trying to ensure that if something fails, that we can provide some error message to indicate this (my initial testing was failing silently with an empty output file so it wasn't immediately obvious what steps needed to be performed).

devlib/collector/perfetto.py
mrkajetanp: Sure, added. The Android framework support was added in Android 9 but it's disabled in 9 and 10 by default on non-Pixel phones. I added a check in the init to manually enable it on those two versions. I think it's pretty unlikely anyone would be using this with Android 8 and below at this point?

mrkajetanp:
Sorry for the delay, here are the updates. I added a check to switch Perfetto on for Android 9 & 10, as well as a check to install tracebox on Linux targets if the service is missing. The tracebox binary was bundled along with the copied Perfetto license.

I also added a "is_running" function in Target, as far as I can tell there was no already existing equivalent and it seems like a useful thing to have so I made it a helper inside Target instead of implementing it just inside the collector.

devlib/target.py
```diff
@@ -279,6 +280,14 @@ def shutils(self):
             self._setup_shutils()
         return self._shutils
 
+    def is_running(self, process):
+        cmd = "pgrep {}"
+        try:
+            self.execute(cmd.format(process))
```
douglas-raillard-arm, line 286: Can use f-string here

devlib/target.py
```diff
@@ -279,6 +280,14 @@ def shutils(self):
             self._setup_shutils()
         return self._shutils
 
+    def is_running(self, process):
+        cmd = "pgrep {}"
```
douglas-raillard-arm, line 284: This needs to use `self.busybox`. Also just because the PID exists does not mean the task is running, it could be stopped (we may still consider that "running") or zombie (that is definitely not "running" by any metric)

devlib/target.py
mrkajetanp: Marc previously asked me to use .format in WA so I'm sticking with the convention here, I think it was for backwards compatibility.

devlib/target.py
mrkajetanp: Any ideas for how to do this differently then? The intention is just to have a way to check if a background deamon is present in the system. From the kernel side the distinction between running, sleeping etc matters but this is a userspace tool so we're probably just interested in knowing whether the process exists or not.

devlib/target.py
douglas-raillard-arm: For backward compat it's not needed anymore since [July 2022](https://github.com/ARM-software/devlib/pull/569), we use [    python_requires='>= 3.7',](https://github.com/ARM-software/devlib/blob/be988bb42b766574bb9bd9191fa3532a65091716/setup.py#L85) for devlib and f-string are from 3.6

In WA the situation is different, as it must work with the last release of devlib in PyPI which is older than that, but inside devlib we can use anything that 3.7 offers.

devlib/target.py
douglas-raillard-arm: `busybox ps -A -o stat,pid` + `busybox awk` . You will also need to have an `as_root=False` parameter that you forward to the `target.execute()`. It's unfortunate but nothing is ever simple with POSIX process model, it's full of traps that need to be handled carefully if you don't want the API to join the cohort of half-broken stuff like `target.kick_off()`.

Otherwise you can us `target.read_file(f'/proc/{pid}/state')` and parse the output:
```
>>> cat /proc/1/status | grep State 
State:  S (sleeping)
```
if the file does not exist, then you can return false, otherwise you can return true as long as it's not a zombie

mrkajetanp:
Updated with using the fstring and a more proper implementation of is_running that takes a regexp so that we can match the name more accurately and avoid a false positive. An added upside is that awk will just make execute return an empty string instead of an exception in the absent case so we can avoid the exception handler.

devlib/target.py
```diff
@@ -279,6 +280,11 @@ def shutils(self):
             self._setup_shutils()
         return self._shutils
 
+    def is_running(self, process_regex):
+        cmd = f"{self.busybox} ps -A -o comm,stat | {self.busybox} awk '$1 ~ /{process_regex}/ && $2 !~ /Z/'b"
```
douglas-raillard-arm, line 284: I'd be _very_ wary of introducing a regex here. We had done that in the freezer cgroup with a grep regex function and that is a great recipe for maintenance nightmare: the second you add that, you tied the implementation of that function to awk regex, forever.

If we really need that feature, I would heavily recommend getting the whole list of PID, and applying the regex on the Python side, using a Python regex, which will not bake any implementation detail in the API. On top of that, that would mean the devlib API would now get 3 flavors of regex (Python, awk, grep), which clearly is not great.

mrkajetanp:
I updated with just directly matching the comm to not overcomplicate things. Busybox ps solves the potential spaces problem already as far as I can tell.
`busybox ps` with no arguments:

```
PID   USER     TIME  COMMAND
    1 0         1h22 /system/bin/init second_stage
    2 0         0:03 [kthreadd]
    3 0         0:00 [rcu_gp]
```
`busybox ps -o comm,stat`
```
COMMAND          STAT
init             S
kthreadd         SW
rcu_gp           IW<
```
None of the tasks that have spaces in the default output have them in the `-o` variant, it abbreviates the task names to a fixed width column and gets rid of whitespace.

douglas-raillard-arm:
>None of the tasks that have spaces in the default output have them in the -o variant, it abbreviates the task names to a fixed width column and gets rid of whitespace.

That's good to know, have you tried the `args` instead of `comm` ? That would contain the full command line though, which is a bit different.

mrkajetanp:
> have you tried the args instead of comm

comm just gives us exactly what we want, args appears slightly broken in the busybox implementation. Namely `args,stat` only shows the args part while `stat,args` shows both as expected. Still not really relevant here, the use case for this function is to check whether a background deamon is present or not and just matching `comm` does that more than adequately imo.

devlib/collector/perfetto.py
mrkajetanp: no problem, sure

douglas-raillard-arm:
> comm just gives us exactly what we want

This is not what we want:

> None of the tasks that have spaces in the default output have them in the -o variant, it abbreviates the task names to a fixed width column and gets rid of whitespace.

Fortunately I found that it's actually not true, the busybox binary shipped in devlib does report comm with spaces properly, e.g.
```
/devlib/devlib/bin/x86_64/busybox ps -A -o stat,comm | grep foo
S    foo bar.sh
```
for a script called "foo bar.sh"

I think the confusion comes from the full command line (`args` in ps parlance) and the task name (`comm`). `comm` is not a simplification of `args`, it's 2 separate strings:
* comm is the same as in the kernel and is set using `prctl(PR_SET_NAME, ...)` syscall (`task_rename` ftrace event)
* cmdline (`args` for busybox ps) is taken from the auxiliary vector of the process, i.e. the concatenated content of `argv` parameter of `main()` in C.

devlib/target.py
```diff
@@ -279,6 +280,11 @@ def shutils(self):
             self._setup_shutils()
         return self._shutils
 
+    def is_running(self, process):
```
douglas-raillard-arm, line 283: nit: s/process/comm/. "process" could mean either PID or comm, but here it's specifically comm that we expect, we are not flexible on the input.

devlib/target.py
```diff
@@ -279,6 +280,11 @@ def shutils(self):
...
+        cmd = f'''{self.busybox} ps -A -o comm,stat | {self.busybox} awk '$1 == "{process}" && $2 != "Z"' '''
```
douglas-raillard-arm, line 284: This breaks for a task named with a space, such as:
```
$ ../devlib/devlib/bin/x86_64/busybox ps -A -o comm,stat | grep foo 
foo bar.sh       S  
```
Because `$2` will be set to "bar.sh".
Instead, you need to use `stat,comm` so that $1 is always the state (which AFAIK can never contain a space), and then match on the rest by emptying $1 (note that `$0` has a leading space in the value, probably a leftover separator from $1/$2, so we need to add it to `{process}` before matching):
```
        cmd = f'''{self.busybox} ps -A -o stat,comm | {self.busybox} awk 'alive=($1 != "Z"); $1=""; $alive && $0 == " {process}"'''
```

douglas-raillard-arm (changes requested):

devlib/target.py
mrkajetanp: Ah right, bit of an academic distinction as there's exactly 0 tasks either on my Android phone or laptop that would have a space in the comm name so this is pretty much exclusive to the "I'll make a script to create this edge case" scenario but sure still worth fixing.
The command
```
/data/local/tmp/bin/busybox ps -A -o stat,comm | /data/local/tmp/bin/busybox awk 'alive=($1 != "Z"); $1=""; $alive && $0 == " traced"'
```
however just returns all the tasks in the system, are you sure there's not an issue here somewhere?

douglas-raillard-arm: > Ah right, bit of an academic distinction

It's unfortunately not. I never hit the issue, _except_ on Android devices: https://bugzilla.kernel.org/show_bug.cgi?id=210193
The problematic comm was `shell srvc 1574`

A quick study on some WA example traces from 2021 I have on my laptop show:
```
AsyncTask #1
AsyncTask #3
DataSender Main
DataSender Send
GLThread 107
GLThread 113
Jit thread pool
magnifier pixel
netlink socket
Profile Saver
Runtime worker 
shell svc 6791
shell svc 6800
Shutdown thread
```

A trace from april 2023 from a pixel 6 shows:
```
AsyncTask #1
shell svc 14342
shell svc 14346
shell svc 14350
shell svc 14354
shell svc 14381
shell svc 14385
```

>  so this is pretty much exclusive to the "I'll make a script to create this edge case" scenario but sure still worth fixing.

After a few years you will realize that a good deal of the alleged "edge cases" are just a convenient statement to sweep issues under the rug. 1/2 of my contributions to devlib have been to fix real-world problems coming out of this sort of shortcuts. 1/4 of my contributions have been about fixing real-world problems coming out of similar code, but wrote as an oversight (including by myself) rather than a conscious decision to ignore the problem.

When you pile up all the features devlib has and how much of that is used in a real-world use case like WA, you quickly realize that if you don't design things to be robust on purpose, you will have problems literally all the time. That used to be the situation with LISA, and to some extent devlib, and a lot of work has been done to align everything and ensure robustness, so these days I place the bar on robustness of the code pretty high for new stuff, as we can't really afford coming back to everything a month later when we realize it's broken.


> however just returns all the tasks in the system, are you sure there's not an issue here somewhere?

Yes, I had a typo that made it not match anything. I can't come up with something that actually work, so I'd suggest just getting the output and processing in Python, as it will be trivial. Something along the line:
```
process = process.strip()
def parse(row):
    state, comm = row.split(' ', 1)
    return state.strip(), comm.strip()
[
    comm.
    for state, comm in map(parse, out.splitlines())
    if comm == process and state != "Z" 
]
```

douglas-raillard-arm: Ah actually here we go, this seems to work:
```
busybox ps -A -o stat,comm | busybox awk 'BEGIN{found=0} {state=$1; $1=""; if ($state != "Z" && $0 == " zsh") {found=1}} END {print found}'
```
f-string can still contain literal `{` and `}` if you use `{{`  `}}`

mrkajetanp: I'm very happy for us to get this corner case fixed of course and you're right on the merit but the things you're listing are output from `trace-cmd`, not output from `busybox ps`. Things like `AsyncTask #1`  that my GB6 traces are full of are threads of a process called `labs.geekbench6` and the only thing that shows up in the `ps` output is the latter.

The help option for `busybox ps` gives this:
```
oriole:/data/local/tmp/bin # ./busybox ps --help
BusyBox v1.32.1 (2021-04-15 16:06:54 BST) multi-call binary.

Usage: ps [-o COL1,COL2=HEADER] [-T]

Show list of processes

	-o COL1,COL2=HEADER	Select columns for display
	-T			Show threads

```
Since we're not passing the -T option, these threads will not show up in the output. Even when passing the -T option while running an iteration of GB6 I still could not find even one field that'd have a space in the `comm` but that's a different story. 

Regardless, the approach you suggested works so I updated the PR.

douglas-raillard-arm: > Since we're not passing the -T option

Maybe we should ? I did not really think about threads vs process as the distinction is a bit of a POSIX compat technicality in Linux (not completely but mostly). I'd rather ere on the side of listing more than less, otherwise a user might want to match on a thread name, and not realize the test will always return false.

mrkajetanp: Sure we might as well, now that we have the spaces in the names handled as well it can't hurt. Should be good to go now.

douglas-raillard-arm: LGTM

devlib/collector/perfetto.py
```diff
@@ -0,0 +1,106 @@
...
+        if target.os == 'android' and os_version > 8:
+            # Android 9 and 10 require traced to be enabled manually
+            if os_version == '9' or os_version == '10':
+                target.execute('setprop persist.traced.enable 1')
+        elif target.os in ['linux', 'android'] and not target.is_running('traced'):
+            self.target_binary = 'tracebox'
+            if not self.target.get_installed(self.target_binary):
+                host_executable = os.path.join(PACKAGE_BIN_DIRECTORY,
+                                               self.target.abi, self.target_binary)
+                self.target.install(host_executable)
```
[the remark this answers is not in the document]
mrkajetanp: If we're installing the binary this means we don't have to :)
Essentially there's two options for using this system:
* `traced` is running -> use `perfetto` to make use of it
* `traced` is not running -> install `tracebox` and use it as a self-contained tool

I just tested it on my laptop, `tracebox` will just launch its own traced and then kill it once it exits.
```
< cat test/configs/scheduling.cfg | out/linux/tracebox --txt -c - -o trace_file.perfetto-trace
[762.274]          service.cc:240 Started traced, listening on @traced-p-355264 @traced-c-355264
[762.324]           probes.cc:104 Starting traced_probes service
[762.324]  probes_producer.cc:341 Connected to the service
[762.327]   perf_producer.cc:1162 Connected to the service
[762.328]    perfetto_cmd.cc:1024 Connected to the Perfetto traced service, TTL: 10s
[762.328] ing_service_impl.cc:988 Configured tracing session 1, #sources:2, duration:10000 ms, #buffers:1, total buffer size:20480 KB, total sessions:1, uid:51991 session name: ""
```

devlib/collector/perfetto.py
```diff
@@ -0,0 +1,106 @@
...
+        self.bg_cmd = None
+        self.config = config
+
+        os_version = target.os_version['release']
+        self.target_binary = 'perfetto'
+        if target.os == 'android' and os_version > 8:
+            # Android 9 and 10 require traced to be enabled manually
+            if os_version == '9' or os_version == '10':
+                target.execute('setprop persist.traced.enable 1')
+        elif target.os in ['linux', 'android'] and not target.is_running('traced'):
```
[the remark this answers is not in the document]
mrkajetanp: Hmm that's intentional though. We only check for `traced` so that we can use `tracebox` if it's missing. If `traced` is present this means that we don't need to use `tracebox` and should call `perfetto` instead.
Calling `tracebox` with `traced` running in the background would be an odd halfway setup I think. From their docs:

> Due to Perfetto's [service-based architecture](https://perfetto.dev/docs/concepts/service-model), in order to capture a trace, the traced (session daemon) and traced_probes (probes and ftrace-interop daemon) need to be running. As per Perfetto v16, the tracebox binary bundles together all the binaries you need in a single executable (a bit like toybox or busybox).

mrkajetanp:
I updated the PR so that for Android targets it'll write to Android's default
perfetto trace directory instead of the devlib one since Perfetto seems to be
having permission issues on non-rooted phones. This makes the collector work
without root access on Android phones, tested on Android 13 + Pixel 6. I also
renamed the output file to 'devlib-trace.perfetto-trace' just to avoid
potential clashes with perfetto traces collected in other ways.

devlib/target.py
```diff
@@ -280,10 +280,11 @@ def shutils(self):
...
         return self._shutils
 
-    def is_running(self, process):
-        cmd = f'''{self.busybox} ps -A -o comm,stat | {self.busybox} awk '$1 == "{process}" && $2 != "Z"' '''
-        result = self.execute(cmd, as_root=False)
-        return bool(result)
+    def is_running(self, comm):
+        cmd_ps = f'''{self.busybox} ps -A -T -o stat,comm'''
+        cmd_awk = f'''{self.busybox} awk 'BEGIN{{found=0}} {{state=$1; $1=""; if ($state != "Z" && $0 == " {comm}") {{found=1}}}} END {{print found}}' '''
+        result = self.execute(f"{cmd_ps} | {cmd_awk}", as_root=False)
```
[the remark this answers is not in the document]
mrkajetanp: Hmm I don't think so, must be an oversight from me - I'll get rid of it

marcbonnici (approved):

PR #622, opened 2023-03-30
target: Add persistent_data_path to AndroidTarget

mrkajetanp:
Some apps, such as Unity-based applications, can store some files in their 'persistent data path' which is usually under '/storage/emulated/0/Android/data'. Add a handle to easily access the path within AndroidTarget in the same way as package_data_directory currently is accessible.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1750,6 +1750,11 @@ class AndroidTarget(Target):
 
     @property
     @memoized
+    def external_storage_app_dir(self):
+        return self.path.join(self.external_storage, 'Android', 'data')
+
+    @property
+    @memoized
     def screen_resolution(self):
         output = self.execute('dumpsys window displays')
         match = ANDROID_SCREEN_RESOLUTION_REGEX.search(output)


devlib/target.py
```diff
@@ -1785,6 +1786,7 @@ def __init__(self,
                                             is_container=is_container,
                                             max_async=max_async)
         self.package_data_directory = package_data_directory
+        self.persistent_data_path = persistent_data_path
```
marcbonnici, line 1789: I assume this should be the same for a ChomeOS target? In which case should we include a similar parameter for that target as well?

mrkajetanp:
> All I can find when looking for "persistent data path" is Unity, are we sure we want to add some app-specific concepts in AndroidTarget ?

That's the only place I found the name itself used but the path is definitely used by other apps as well here and there. It's analogous to the `/data/data` one but does not require root access. I'm not sure Android has any 'official' name for it which is why "persistent data path" is what I went with but if we have any better names then that works just as well I think.

devlib/target.py
mrkajetanp: Hm would it? I don't really know anything about ChromeOS from that side..

douglas-raillard-arm:
That makes sense, does Android allow querying for that path in any way or do apps have to know it in advance ?

mrkajetanp:
Ah apparently it can, Android's Context has a method for getting the application's data directory:

https://developer.android.com/reference/android/content/Context#getExternalFilesDir(java.lang.String)

We could always rename this from persistent_data_path to external_files_path or dir?

marcbonnici:
Looking into this more it looks like the `persistent_data_path` or `external_files_path` term actually refers a specific applications data path. From [1].

`Returns absolute paths to application-specific directories on all shared/external storage devices where the application can place persistent files it owns.`

So we would be providing the root of this directory and users would need to manually join the package name to this path.

IIUC looking at how this path is generated [2] it uses the external directory function (which we expose with `external_storage`) along with the constant sub-path `Android/data`. Therefore I think we can avoid adding a new parameter for this and instead add a convenience property with the equivalent of `target.path.join(target.external_storage', 'Android', 'data')`.

As for the name perhaps to keep inline with the android terminology we could expose something like `external_storage_app_dir`? (Open to better naming suggestions).

What do you think?  

[1] https://developer.android.com/reference/android/content/Context#getExternalFilesDir(java.lang.String) 
[2] https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/os/Environment.java#225

devlib/target.py
marcbonnici: For completeness the way we handle ChromeOS targets (which support Android apps) is essentially creating both a Linux and Android Target together so any parameters (or properties) we add should also be usable by the Android portion of the ChromeOs target.

mrkajetanp:
> instead add a convenience property

Ah yeah that makes a lot of sense, `/sdcard` is a symlink to `/storage/emulated/0/Android/data` so that works. This also allows us to do it exclusively here instead of need a separate WA PR so that's a win.

> What do you think?

The name seems fine to me. I've updated this PR accordingly and will close the WA one.

devlib/target.py
```diff
@@ -1785,6 +1785,7 @@ def __init__(self,
                                             is_container=is_container,
                                             max_async=max_async)
         self.package_data_directory = package_data_directory
+        self.external_storage_app_dir = self.path.join(self.external_storage, 'Android', 'data')
```
marcbonnici, line 1788: I would suggest using the property decorator for this similar to the existing `external_storage` implementation to defer the evaluation.

mrkajetanp:
Updated, sorry for the delay..

marcbonnici (approved):

PR #624, opened 2023-04-04
Fix kick off

douglas-raillard-arm:


marcbonnici (approved):

mrkajetanp:
I can confirm that with the updated PR the instrument finally works as expected so simply getting rid of `kick_off` this way is probably a good idea.

marcbonnici (approved):
Just for some context, the original intention behind the `kick_off` method was to support platforms where having multiple connections was not possible (e.g. gem5) and therefore needed to be different to the regular background command. 

With all the new changes that have bee introduced this is not a configuration that is expected, however in theory it would be possible for a connection implementation to override the kick_off mechanism to try and accommodate such requirements.

PR #626, opened 2023-04-06
connection: Terminate background commands on close()

douglas-raillard-arm:
Ensure all background commands are terminated before we close the connection.

This PR adds tracking of all outstanding BackgroundCommand inside `ConnectionBase._current_bg_cmds`. This attribute is changed from WeakSet() to set(), and the BackgroundCommand themselves deregister themselves when they terminate. This allows properly canceling _all_ outstanding_ bg command when the connection is closed.

BackgroundCommand.__init__ polls current bg cmds of its connection to reap any terminated commands, so that APIs that leak objects like the change introduced by this PR:
https://github.com/ARM-software/devlib/pull/624
or any other non-careful user does not result in endless leak of associated resources.

### DIFF
--- a/devlib/connection.py
+++ b/devlib/connection.py
@@ -66,6 +66,7 @@ class ConnectionBase(InitCheckpoint):
         self._closed = False
         self._close_lock = threading.Lock()
         self.busybox = None
+        self.logger = logging.getLogger('Connection')
 
     def cancel_running_command(self):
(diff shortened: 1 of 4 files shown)


douglas-raillard-arm:
Updated PR with fix for `RuntimeError(Set changed size during iteration)`

marcbonnici (approved):

PR #627, opened 2023-04-14
ftrace: Set top-level buffer size

douglas-raillard-arm:
trace-cmd start -B devlib -b 42 will set the buffer size for the "devlib" ftrace instance but will not set the buffer size of the top-level buffer.

Unfortunately, some events still end up in the top-level buffer regardless of any configuration such as the "print" event. This can lead to lost events because the buffer size was too small.

Avoid that by using the buffer size for both top-level and devlib's instance.


devlib/collector/ftrace.py
```diff
@@ -240,6 +240,17 @@ def available_functions(self):
     def reset(self):
         self.target.execute('{} reset -B devlib'.format(self.target_binary),
                             as_root=True, timeout=TIMEOUT)
+
+        # trace-cmd start will not set the top-level buffer size if passed -B
+        # parameter, but unfortunately some events still end up there (e.g.
+        # print event). So we still need to set that size, otherwise the buffer
+        # might be too small and some event lost.
+        if self.buffer_size is not None:
+            self.target.target.write_value(
```
mrkajetanp, line 249: should be `self.target.write_value`

marcbonnici (approved):

PR #628, opened 2023-05-16
Fix bg cmds

marcbonnici:
Fix a number of issues discovered during testing of https://github.com/ARM-software/devlib/pull/618/.

- Update kill command syntax to match the currently support busybox implementation.
- Ensure the correct signal is to send to a command.
- Prevent commands from being killed multiple times.

### DIFF
--- a/devlib/connection.py
+++ b/devlib/connection.py
@@ -35,7 +35,7 @@ _KILL_TIMEOUT = 3
 
 
 def _kill_pgid_cmd(pgid, sig, busybox):
-    return '{} kill -{} -- -{}'.format(busybox, sig.value, pgid)
+    return '{} kill -{} -{}'.format(busybox, sig.value, pgid)
 
 def _popen_communicate(bg, popen, input, timeout):
     try:
--- a/devlib/connection.py
+++ b/devlib/connection.py
@@ -170,7 +170,7 @@ class BackgroundCommand(ABC):
         :type signal: signal.Signals
         """
         try:
-            self._send_signal(signal.SIGKILL)
+            self._send_signal(sig)
         finally:
             # Deregister if the command has finished
             self.poll()
--- a/devlib/connection.py
+++ b/devlib/connection.py
@@ -170,7 +170,7 @@ class BackgroundCommand(ABC):
         :type signal: signal.Signals
         """
         try:
-            self._send_signal(sig)
+            return self._send_signal(sig)
         finally:
             # Deregister if the command has finished
             self.poll()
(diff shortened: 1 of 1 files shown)


devlib/connection.py
```diff
@@ -485,11 +485,12 @@ def _cancel(self, kill_timeout):
         self.send_signal(signal.SIGTERM)
         # Check if the command terminated quickly
         time.sleep(10e-3)
-        # Otherwise wait for the full timeout and kill it
-        if self.poll() is None:
+        # Otherwise wait for the full timeout and if still alive kill it
+        if self.poll() is None and self.redirect_thread.is_alive():
```
marcbonnici, line 489: @douglas-raillard-arm I'm not sure if this is the best way to detect whether the the command is still alive (or if it even makes sense as it is). 
Do you have any thoughts on a better way to handle this?

douglas-raillard-arm: self.poll() really should be enough. If the command is alive, it's supposed to return None. When it's dead, it returns the exit code. If that's not the case it's poll() that needs fixing, not its call sites.

The API is modeled after Popen.poll: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.poll

marcbonnici: Ok cool thanks, during my testing `poll()` was always returning `None` regardless of the command status so will take a closer look at what is going on there.

douglas-raillard-arm: Which connection class where you using for your tests ?

marcbonnici: This was using an SSH connection.

douglas-raillard-arm: Ok I'll also try to have a look tomorrow then

marcbonnici: Thank you for the offer, however on further investigation please ignore me. 
After resetting my environment and board I can no long reproduce the original issue so I think I must have had something in a strange state and things are now working as expected. 
I am however tempted to leave the additional `poll` in place if that still makes sense.

douglas-raillard-arm: The issue is real, there was a missing `return` in `wait()` following https://github.com/ARM-software/devlib/pull/626 . I've been doing mostly Rust recently ...

Here is the fix, feel free to just integrate it in your PR if you want to avoid merge conflicts: https://github.com/ARM-software/devlib/pull/629

>  I am however tempted to leave the additional poll in place if that still makes sense.

The exact same condition is tested in `ParamikoBackgroundCommand._poll` so it won't add anything:
```
    def _poll(self):
        if self.redirect_thread.is_alive():
            return None
        elif self.chan.exit_status_ready():
            return self.wait()
        else:
            return None
```

marcbonnici: > The issue is real, there was a missing return

Oh great, thank you for finding that, I was a rather confused at what was going on..

> The exact same condition is tested in ParamikoBackgroundCommand._poll so it won't add anything

Good point thanks, I'll drop this change and pull your commitx in instead.

PR #629, opened 2023-05-17
connection: Make BackgroundCommand.wait() return non-None

douglas-raillard-arm:
Lack of return statement in wait() was making it return None instead of the exit code. Add appropriate return statement in wait() and other function to ensure return value is not lost.


marcbonnici:
Merged as part of https://github.com/ARM-software/devlib/pull/628

Issue #611, opened 2022-12-12
Remove (or shrink) the default set of modules loaded 


douglas-raillard-arm:
Cgroup module randomly fails to load on pixel 6 for unknown reasons. It seems to be related to how the USB is setup (via a hub or not) but we haven't been able to really understand the issue so far. Since the cgroups module is enabled by default, WA will crash because of unused feature, which is a shame.

LISA is not directly impacted because the approach to loading is different:
* we don't enable any module by default
* the modules are loaded on-demand
* there is a user-provided list of forbidden modules. Users typically don't care about needing module X or Y unless it makes their setup crash. The code already knows what module it needs, so all we need is the list of forbidden module.

Some modules have to be loaded in advance and by the user, but none of the ones in the default list seem to fall in that category:

```python
    default_modules = [
        'hotplug',
        'cpufreq',
        'cpuidle',
        'cgroups',
        'hwmon',
    ]
```

I haven't looked at the details but it seems that WA could also use a loading-on-demand system, in which case there would be no point in having that default module set anymore.

PR #634, opened 2023-06-27
Cleanup module load

douglas-raillard-arm:
Cleanup module code to improve robustness, modularity and enable lazy loading of modules.

The main features are:
* More checks (e.g. it is now impossible to add multiple modules of the same kind)
* Predictable Target.modules content: it used to be a mix of str and dict, now it's str only
* Modules don't have to be declared in `Target(modules=...)` anymore to be used. They can simply be accessed and that will load them lazily. Passing `Target(modules=[{mod: None}])` will prevent loading of the module in all circumstances, including lazily.

For now, modules passed to `Target(modules=...)` will be eagerly loaded, but that could be changed quite easily to enable full lazy loading if we wanted.

On top of that, the list of default modules is made empty. This fixes https://github.com/ARM-software/devlib/issues/611 as nothing will be loaded by default, but existing code will keep working thanks to lazy loading.

### DIFF
--- a/devlib/module/__init__.py
+++ b/devlib/module/__init__.py
@@ -15,16 +15,30 @@
 import logging
 from inspect import isclass
 
-from past.builtins import basestring
-
-from devlib.utils.misc import walk_modules
+from devlib.exception import TargetStableError
 from devlib.utils.types import identifier
+from devlib.utils.misc import walk_modules
+
+_module_registry = {}
+
+def register_module(mod):
+    if not issubclass(mod, Module):
+        raise ValueError('A module must subclass devlib.Module')
+
+    if mod.name is None:
+        raise ValueError('A module must define a name')
+
+    try:
+        existing = _module_registry[mod.name]
+    except KeyError:
+        pass
+    else:
+        if existing is not mod:
+            raise ValueError(f'Module "{mod.name}" already exists')
+    _module_registry[mod.name] = mod
 
 
-__module_cache = {}
-
-
-class Module(object):
+class Module:
 
     name = None
     kind = None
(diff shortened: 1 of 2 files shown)


douglas-raillard-arm:
Fixed unwanted error logging when using `Target.has()` to avoid littering WA output with:
```
ERROR    Module "hotplug" failed to install on target: Module "hotplug" is not supported by the target
```
Note that `Target.has()` still has to try to load the module, as only that can tell if the target supports the module or not. If the users want to speedup that process, `devlib` allows forcefully disabling a module no matter what, so I suppose WA could expose such deny list in its config.

marcbonnici (commented):
Please ignore.

marcbonnici (commented):
Apologies, please ignore the previous comment that was meant for a different PR.

douglas-raillard-arm:
PR updated to import `devlib.module` submodules in case a Module is not found to match the previous behavior. Otherwise, people would need explicit `import` statement to make the relevant classes available for auto-loading.

douglas-raillard-arm:
@marcbonnici Any update on this one ? AFAIR the last update to the PR fixed the outstanding issues.

douglas-raillard-arm:
(last push is just a rebase on current master, there was no conflict to solve)

marcbonnici (approved):
Apologies for missing the updates to this PR. 

Aside from the typo, this seems to works well to me.

douglas-raillard-arm:
@marcbonnici PR updated:
* eefine typo fixed
* rebased on master

marcbonnici (approved):

PR #638, opened 2023-08-02
utils.rendering: Fix activity matching

mrkajetanp:
Change the SurfaceFlingerFrameCollector to match activities by prefix instead of looking for an exact match. This will allow to account for activities with variable suffixes.
Raise an error if more than one activity matches the provided view.

Suggested-by: Andriani Mappoura <andriani.mappoura@arm.com>

### DIFF
--- a/devlib/utils/rendering.py
+++ b/devlib/utils/rendering.py
@@ -131,9 +131,18 @@ class SurfaceFlingerFrameCollector(FrameCollector):
         self.header = header or SurfaceFlingerFrame._fields
 
     def collect_frames(self, wfh):
-        for activity in self.list():
-            if activity == self.view:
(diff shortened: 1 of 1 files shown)


mrkajetanp:
Note: The current behaviour is to silently fail if the view doesn't match any of the activities found by SurfaceFlinger. I left that unchanged but I think it probably makes sense to add another check and throw a warning if the list is empty so that the user at least knows where the problem is coming from. I can update the PR if that's desirable to have.

marcbonnici (commented):
LGTM, and I think agree with your comment over raising an error if the view isn't found. Providing that feedback seems useful especially is we're already providing feedback on multiple matches.

mrkajetanp:
All right I updated the PR, should be ready now :)

marcbonnici (approved):

PR #641, opened 2023-08-07
bin: Update busybox

douglas-raillard-arm:
Update busybox to version 1.36.1 with defconfig and uniformly built on Alpine v3.18, statically linked to musl libc.


douglas-raillard-arm:
There seems to be an issue where `insmod` does not forward the return code from the module init function but returns `1` instead. I need to dig deeper to figure out if it's a busybox config issue or a change in behavior since the previous version.

douglas-raillard-arm:
Turns out to be due to CONFIG_MODPROBE_SMALL=y, which is enabled by default. I'll rebuild the binaries and update the PR

douglas-raillard-arm:
Updated with the fixed build config to get the "full" insmod command instead of the minimal one that breaks return code forwarding.

marcbonnici (approved):

PR #642, opened 2023-08-14
ftrace: Separate top_buffer_size from buffer_size

mrkajetanp:
Since we now set the top buffer size to be the same as the devlib buffer size, this effectively halves the maximum available buffer size that can be set while using devlib. Whatever size is passed as `buffer_size` will be allocated twice, even if the top buffer is hardly used at all.

This commit separates them into `buffer_size` and `top_buffer_size`. If the latter is not passed, the behaviour will not change compared to now.

Fixes: e0c53d09990b5501e493d048a5dce067d8990281

### DIFF
--- a/devlib/collector/ftrace.py
+++ b/devlib/collector/ftrace.py
@@ -60,6 +60,7 @@ class FtraceCollector(CollectorBase):
                  tracer=None,
                  trace_children_functions=False,
                  buffer_size=None,
+                 top_buffer_size=None,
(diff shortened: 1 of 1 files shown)


marcbonnici (approved):

douglas-raillard-arm:
LGTM, if we end up supporting reentrant collectors with alternative buffers then we will still have a single non-top-level buffer per collector, so the parameter will  not need to be generalized to a dict.

PR #643, opened 2023-08-23
ftrace: Support trace-cmd record mode

mrkajetanp:
The current implementation of the FtraceCollector only supports using ftrace in its start-stop mode which keeps the traces in memory and then writes to file at the end. Trace-cmd also supports a different mode - record which writes to file periodically meaning that it allows recording traces that would exceed the 'start' mode buffer size in exchange for added noise.

This commit adds support for selecting which mode trace-cmd should be run in. It adds a new collector parameter - 'trace_cmd_mode' which is used to distinguish between the two approaches.

While at it, it also cleans up the collector in some places, e.g. the teardown() method.

### DIFF
--- a/devlib/collector/ftrace.py
+++ b/devlib/collector/ftrace.py
@@ -74,6 +74,7 @@ class FtraceCollector(CollectorBase):
                  trace_clock='local',
                  saved_cmdlines_nr=4096,
                  trace_cmd_mode='start',
+                 record_interval=1000,
                  ):
         super(FtraceCollector, self).__init__(target)
         self.events = events if events is not None else DEFAULT_EVENTS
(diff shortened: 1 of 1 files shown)


mrkajetanp:
I pushed an update to make the entire collector just use a tempdir inside the working
directory from the start for both modes. If nothing else it'll make it more
self contained.

mrkajetanp:
Updated to wait until the background command actually exits after sending
SIGINT - otherwise there's a race condition with removing the temp directory in
the teardown() method.

douglas-raillard-arm:
Pending on: https://github.com/ARM-software/devlib/issues/645

douglas-raillard-arm:
Closing as it has a natural successor: https://github.com/ARM-software/devlib/pull/689

The parameter is reworded in a way that is not trace-cmd specific, so that the write-to-disk and write-to-memory modes can become a pattern in other collectors for which it might be relevant.

PR #650, opened 2023-09-21
ftrace: Do not read-verify buffer_size_kb value

cloehle:
The sysfs documentation mentions that the value written to buffer_size_kb ftrace field may be rounded up.
So skip the verify loop on this field.

The case we are worried about, a requested buffer
size that the target cannot fulfill is caught anyway, as the sysfs write returns with an error that is caught.


marcbonnici (approved):

cloehle:
Thanks!
FYI the change making that necessary was
commit 6d98a0f2ac3c ("tracing: Set actual size after ring buffer resize")

PR #651, opened 2023-10-03
devlib: Add ThermalZone type and policy support to thermal module

msrasmussen:
The thermal module currently only reads thermal zone ids and allow temperature reading. The mandatory thermal zone 'type' describes what the zone is and is therefore quite useful information. This commit also adds support for reading the current thermal zone policy and available policies.pp

### DIFF
--- a/devlib/module/thermal.py
+++ b/devlib/module/thermal.py
@@ -13,8 +13,11 @@
 # limitations under the License.
 
 import re
+import logging
+import devlib.utils.asyn as asyn
 
 from devlib.module import Module
+from devlib.exception import TargetStableCalledProcessError
 
 class TripPoint(object):
     def __init__(self, zone, _id):
@@ -27,19 +30,22 @@ class TripPoint(object):
     def target(self):
         return self.zone.target
 
-    def get_temperature(self):
+    @asyn.asyncf
+    async def get_temperature(self):
         """Returns the currently configured temperature of the trip point"""
         temp_file = self.target.path.join(self.zone.path, self.temp_node)
-        return self.target.read_int(temp_file)
+        return await self.target.read_int.asyn(temp_file)
 
-    def set_temperature(self, temperature):
+    @asyn.asyncf
+    async def set_temperature(self, temperature):
         temp_file = self.target.path.join(self.zone.path, self.temp_node)
-        self.target.write_value(temp_file, temperature)
+        await self.target.write_value.asyn(temp_file, temperature)
 
-    def get_type(self):
+    @asyn.asyncf
+    async def get_type(self):
         """Returns the type of trip point"""
         type_file = self.target.path.join(self.zone.path, self.type_node)
-        return self.target.read_value(type_file)
+        return await self.target.read_value.asyn(type_file)
 
 class ThermalZone(object):
     def __init__(self, target, root, _id):
@@ -47,6 +53,7 @@ class ThermalZone(object):
         self.name = 'thermal_zone' + _id
         self.path = target.path.join(root, self.name)
         self.trip_points = {}
+        self.type = self.target.read_value(self.target.path.join(self.path, 'type'))
 
         for entry in self.target.list_directory(self.path, as_root=target.is_rooted):
             re_match = re.match('^trip_point_([0-9]+)_temp', entry)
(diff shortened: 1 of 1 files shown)


msrasmussen:
@douglas-raillard-arm Could you take a look at this PR?

devlib/module/thermal.py
```diff
@@ -68,7 +70,25 @@ def set_enabled(self, enabled=True):
     def get_temperature(self):
         """Returns the temperature of the thermal zone"""
         temp_file = self.target.path.join(self.path, 'temp')
-        return self.target.read_int(temp_file)
+        try:
+            temp_value = self.target.read_int(temp_file)
+        except CalledProcessError:
+            temp_value = -1
```
douglas-raillard-arm, line 76: Unless strictly necessary we really don't want in-band signaling like that. If the file does not exist or is unreadable, that function should raise an exception. This way the caller will never be at risk of manipulating a meaningless value.

devlib/module/thermal.py
```diff
@@ -105,3 +125,10 @@ def disable_all_zones(self):
         """Disables all the thermal zones in the target"""
         for zone in self.zones.values():
             zone.set_enabled(False)
+
+    def get_all_temperatures(self):
+        """Returns dictionary with current reading of all thermal zones"""
+        tz_temps = {}
```
douglas-raillard-arm, line 131: can use dict comprehension for that

douglas-raillard-arm (changes requested):
Overall looks ok, but it would be good for new code to be async-ready. You can take example on the cpufreq module:
https://github.com/ARM-software/devlib/blob/master/devlib/module/cpufreq.py
and the docstring of `asyncf()`:
https://github.com/ARM-software/devlib/blob/9199d8884e7257155b57a8ab2baca21f78923916/devlib/utils/asyn.py#L295

You might also want to consider making a context manager to set a policy and then set back the previous value, in the style of cpufreq's `use_governor()`:
https://github.com/ARM-software/devlib/blob/9199d8884e7257155b57a8ab2baca21f78923916/devlib/module/cpufreq.py#L111

devlib/module/thermal.py
```diff
@@ -68,7 +70,25 @@ def set_enabled(self, enabled=True):
     def get_temperature(self):
         """Returns the temperature of the thermal zone"""
         temp_file = self.target.path.join(self.path, 'temp')
```
douglas-raillard-arm, line 72: `temp_file` is strongly associated with "temporary file" concept. Unless the rest of the module uses that convention, I'd suggest use something else like `temperature_file` to avoid any confusion

devlib/module/thermal.py
```diff
@@ -105,3 +125,10 @@ def disable_all_zones(self):
         """Disables all the thermal zones in the target"""
         for zone in self.zones.values():
             zone.set_enabled(False)
+
+    def get_all_temperatures(self):
```
douglas-raillard-arm, line 129: Given our offline discussion on error handling, you might want to add an `error='raise'` parameter, that the user could set to `raise` or `ignore`. `ignore` will just ignore the error and not add the key in the returned dict, `raise` would re-raise the exception.

douglas-raillard-arm: The `ignore` case should probably log at error level still if the reason of the problem is a driver issue (i.e. not part of normal operation for that file), so consumers have a chance of knowing there is a problem related to that if some other code misbehaves (e.g. code iterating on zones then looking up in the dict and failing with a `KeyError`)

devlib/module/thermal.py
```diff
@@ -93,15 +154,44 @@ def __init__(self, target):
...
+        Returns dictionary with current reading of all thermal zones.
+
+        :params error: Sensor read error handling (raise or ignore)
+        :type error: str
+
+        :returns: a dictionary in the form: {tz_type:temperature}
+        """
+
+        if not isinstance(error, str):
+            raise TypeError('error should a str')
```
douglas-raillard-arm, line 182: This can be removed and checked at the point of use

devlib/module/thermal.py
```diff
@@ -93,15 +154,44 @@ def __init__(self, target):
...
+
+        async def get_temperature_noexcep(item):
+            tzid, tz = item
+            try:
+                temperature = await tz.get_temperature.asyn()
+            except TargetStableCalledProcessError as e:
+                if error != 'ignore':
+                    raise e
+                self.logger.warning('Skipping thermal_zone_id={} thermal_zone_type={} error="{}"'.format(tzid, tz.type, e))
+                return None
```
douglas-raillard-arm, line 192: This can be an exhaustive check:
```
if error == 'ignore':
    self.logger(...)
elif error == 'raise':
    raise
else:
   raise ValueError(f'Unknown error parameter value: {error}')
```
And you can remove the earlier check on error being a string.

devlib/module/thermal.py
```diff
@@ -93,15 +154,44 @@ def __init__(self, target):
...
+            raise TypeError('error should a str')
+
+        async def get_temperature_noexcep(item):
+            tzid, tz = item
+            try:
+                temperature = await tz.get_temperature.asyn()
+            except TargetStableCalledProcessError as e:
+                if error != 'ignore':
+                    raise e
+                self.logger.warning('Skipping thermal_zone_id={} thermal_zone_type={} error="{}"'.format(tzid, tz.type, e))
```
douglas-raillard-arm, line 191: use f-strings instead of str.format()

douglas-raillard-arm (changes requested):

douglas-raillard-arm (approved):
LGTM

marcbonnici (approved):

PR #657, opened 2023-10-18
Fix config check

douglas-raillard-arm:


douglas-raillard-arm:
Tried it out and I'll need to rebuild busybox to include the `systcl` applet before this can be merged

douglas-raillard-arm:
Actually sysctl parameters are accessible in /proc/sys so I updated the PR accordingly, no need to recompile busybox. It's now tested.

marcbonnici (approved):

PR #658, opened 2023-11-13
UIBenchJankTests: modification to support Android 12/14 versions

eliftopuz:
dex file search for the package name is modified. In the apk, package name stated twice and the second one includes the tests information. Tested with other benchmarks (geekbench,pcmark,jankbench in Android 12) as well.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -231,12 +231,15 @@ class ApkInfo(object):
             parser = etree.XMLParser(encoding='utf-8', recover=True)
             xml_tree = etree.parse(StringIO(dump), parser)
 
-            package = next((i for i in xml_tree.iter('package')
-                           if i.attrib['name'] == self.package), None)
+            package = []
+            for i in xml_tree.iter('package'):
+                if i.attrib['name'] == self.package:
+                    package.append(i)
 
-            self._methods = [(meth.attrib['name'], klass.attrib['name'])
-                             for klass in package.iter('class')
-                             for meth in klass.iter('method')] if package else []
+            for elem in package:
+                self._methods.extend([(meth.attrib['name'], klass.attrib['name']) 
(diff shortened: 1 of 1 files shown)


devlib/utils/android.py
```diff
@@ -231,8 +231,10 @@ def methods(self):
             parser = etree.XMLParser(encoding='utf-8', recover=True)
             xml_tree = etree.parse(StringIO(dump), parser)
 
-            package = next((i for i in xml_tree.iter('package')
-                           if i.attrib['name'] == self.package), None)
+            package = None
```
douglas-raillard-arm, line 234: Isn't there a better criteria than just taking the first or last item ? It seems pretty arbitrary and not particularly stable, 6 months from now it might get reversed again or even become random.

It would be a better idea to just change the `if` clause's condition to select the item that actually contains the info we need

eliftopuz: You're right, I will make the changes.

devlib/utils/android.py
```diff
@@ -231,12 +231,17 @@ def methods(self):
...
+            for i in xml_tree.iter('package'):
+                if i.attrib['name'] == self.package:
+                    package = next((i for methods in i.iter('method')
+                                   if methods.attrib['name'].startswith('test')), None)
```
marcbonnici, line 238: What is the reason or only picking up methods that begin with test? This method should be generic enough to extract all available methods. Is it possible to filter where this information is consumed?

eliftopuz: Hi Marc,
I was just getting the package which includes methods starts with the 'test' with other methods inside that package. Because we need 'test' methods to run the workload, but as you said to keep it generic, I will collect all the methods. WA will get the methods starts with the 'test' anyway.

marcbonnici (approved):

PR #659, opened 2024-01-05
target: Fix indentation of a misaligned line and groundwork for unlocking device screen

metin-arm:
- [target: Fix indentation of a misaligned line](https://github.com/ARM-software/devlib/pull/659/commits/cb2945125a63084144525deb631f65a26f6037de)
- [target: Handle dozing case in checking Android screen state](https://github.com/ARM-software/devlib/pull/659/commits/50424e3f7567892ca2246ab89db030aa4e6c2182)
- [target: Add helper function to check Android screen's locking state](https://github.com/ARM-software/devlib/pull/659/commits/972c234804e3823dcf6ce87ce9bc673eceb67567)

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -68,7 +68,7 @@ import devlib.utils.asyn as asyn
 
 
 FSTAB_ENTRY_REGEX = re.compile(r'(\S+) on (.+) type (\S+) \((\S+)\)')
-ANDROID_SCREEN_STATE_REGEX = re.compile('(?:mPowerState|mScreenOn|mWakefulness|Display Power: state)=([0-9]+|true|false|ON|OFF|DOZE|Asleep|Awake)',
+ANDROID_SCREEN_STATE_REGEX = re.compile('(?:mPowerState|mScreenOn|mWakefulness|Display Power: state)=([0-9]+|true|false|ON|OFF|DOZE|Dozing|Asleep|Awake)',
                                         re.IGNORECASE)
 ANDROID_SCREEN_RESOLUTION_REGEX = re.compile(r'cur=(?P<width>\d+)x(?P<height>\d+)')
 ANDROID_SCREEN_ROTATION_REGEX = re.compile(r'orientation=(?P<rotation>[0-3])')
@@ -2220,6 +2220,8 @@ class AndroidTarget(Target):
         if match:
             if 'DOZE' in match.group(1).upper():
                 return True
+            if match.group(1) == 'Dozing':
+                return False
             if match.group(1) == 'Asleep':
                 return False
             if match.group(1) == 'Awake':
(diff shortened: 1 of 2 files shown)


devlib/target.py
```diff
@@ -2226,14 +2226,16 @@ async def is_screen_on(self):
                 return True
             return boolean(match.group(1))
         else:
-            raise TargetStableError('Could not establish screen state.')
+            self.logger.debug('Could not get screen state (probably off?).')
```
douglas-raillard-arm, line 2229: That's a bit dodgy. If we end up on that else branch, it does not mean the screen is off, it means we don't know, and the original code was reflecting that. I don't think it's a good idea to remove that and just blindly assume the screen is off. 

If client code does not care about distinction between screen off and screen unknown, we should make a new exception type (inheriting from TargetStableError) so we don't accidentally catch unrelated TargetStableError and then catch it in the code in question.

douglas-raillard-arm (changes requested):

devlib/target.py
metin-arm: Well, let's start with `ANDROID_SCREEN_STATE_REGEX` then. I'm unsure if `adb dumpsys power` is the correct way of probing screen state. Why not using `adb dumpsys window` if we are trying to determine whether the screen is on?
Maybe we should extend `is_screen_on()` to tell screen lock state (locked, unlocked), too. Probably it should be renamed (e.g., `get_screen_state()`) as well.

marcbonnici: I vaguely remember having issues with using `dumpsys window` to determine this state in the past and `dumpsys power` was the most reliable method at the time. It's possible that things have changed in later versions of Android so we could even have some hybrid to try and ensure we get a reliable reading rather than making any assumptions. 

Also I would suggest leaving `is_screen_on` as a standalone method as it is, but adding something else to determine the lock state sounds like it could be useful.

metin-arm: `is_screen_on()` raises `Could not establish screen state` exception if my Pixel 6 (Android-12) is in idle state (screen is completely off) instead of `Asleep`. This is why I removed that exception.
I could handle the exception and try to wake up device (via keyevent 26), but this does not sound right solution, too.

metin-arm: Added https://github.com/ARM-software/devlib/pull/659/commits/50424e3f7567892ca2246ab89db030aa4e6c2182 to handle `Could not establish screen state` exception when the device is idling.

metin-arm:
@marcbonnici , @douglas-raillard-arm : can you please review again? Thanks.

devlib/target.py
```diff
@@ -2213,13 +2213,20 @@ def wait_for_device(self, timeout=30):
     def reboot_bootloader(self, timeout=30):
         self.conn.reboot_bootloader()
 
+    @asyn.asyncf
+    async def is_screen_locked(self):
+        screen_state = await self.execute.asyn('dumpsys window')
+        return True if 'mDreamingLockscreen=true' in screen_state else False
```
douglas-raillard-arm, line 2219: `s/True if <cond> else False/<cond>/`

metin-arm: Neat! Done. Thanks.

marcbonnici (approved):

PR #665, opened 2024-01-17
Bunch of fixes & improvements

metin-arm:
#### Dropped the changes in `tests/test_target.py` for now. Should be OK to merge.

* ec006a0 - test_target: Add copyright statement
* be6dc8c - ChromeOsTarget: Fix building SSH connection parameter list
* f54bef0 - utils/ssh: Load host keys only if strict_host_check is true
* 34da9f9 - target: Customize as_root parameter of *write_value()
* 95f4a28 - devlib: Remove unused imports
* 121921f - devlib/init: Resolve pylint issues
* 586e340 - doc/connection: Fix typo Telenet

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -505,7 +505,7 @@ class Target(object):
             self.hard_reset()  # pylint: disable=no-member
         else:
             if not self.is_connected:
-                message = 'Cannot reboot target becuase it is disconnected. ' +\
+                message = 'Cannot reboot target because it is disconnected. ' +\
                           'Either connect() first, or specify hard=True ' +\
                           '(in which case, a hard_reset module must be installed)'
                 raise TargetTransientError(message)
@@ -2970,9 +2970,11 @@ class ChromeOsTarget(LinuxTarget):
                            'total_transfer_timeout', 'poll_transfers',
                            'start_transfer_poll_delay']
         self.ssh_connection_settings = {}
-        for setting in ssh_conn_params:
-            if connection_settings.get(setting, None):
-                self.ssh_connection_settings[setting] = connection_settings[setting]
+        self.ssh_connection_settings.update(
+            (key, value)
+            for key, value in connection_settings.items()
+            if key in ssh_conn_params
+        )
 
         super(ChromeOsTarget, self).__init__(connection_settings=self.ssh_connection_settings,
                                              platform=platform,
@@ -2991,9 +2993,11 @@ class ChromeOsTarget(LinuxTarget):
         # Pull out adb connection settings
         adb_conn_params = ['device', 'adb_server', 'timeout']
         self.android_connection_settings = {}
-        for setting in adb_conn_params:
-            if connection_settings.get(setting, None):
-                self.android_connection_settings[setting] = connection_settings[setting]
+        self.android_connection_settings.update(
+            (key, value)
+            for key, value in connection_settings.items()
+            if key in adb_conn_params
+        )
 
         # If adb device is not explicitly specified use same as ssh host
         if not connection_settings.get('device', None):
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1015,19 +1015,19 @@ class Target(object):
         return await self.read_value.asyn(path, kind=boolean)
 
     @asyn.asynccontextmanager
-    async def revertable_write_value(self, path, value, verify=True):
+    async def revertable_write_value(self, path, value, verify=True, as_root=True):
         orig_value = self.read_value(path)
         try:
-            await self.write_value.asyn(path, value, verify)
+            await self.write_value.asyn(path, value, verify=verify, as_root=as_root)
             yield
         finally:
-            await self.write_value.asyn(path, orig_value, verify)
+            await self.write_value.asyn(path, orig_value, verify=verify, as_root=as_root)
 
     def batch_revertable_write_value(self, kwargs_list):
         return batch_contextmanager(self.revertable_write_value, kwargs_list)
 
     @asyn.asyncf
-    async def write_value(self, path, value, verify=True):
+    async def write_value(self, path, value, verify=True, as_root=True):
         self.async_manager.track_access(
             asyn.PathAccess(namespace='target', path=path, mode='w')
         )
@@ -1059,7 +1059,7 @@ fi
         cmd = cmd.format(busybox=quote(self.busybox), path=quote(path), value=quote(value))
 
         try:
-            await self.execute.asyn(cmd, check_exit_code=True, as_root=True)
+            await self.execute.asyn(cmd, check_exit_code=True, as_root=as_root)
         except TargetCalledProcessError as e:
             if e.returncode == 10:
                 raise TargetStableError('Could not write "{value}" to {path}: {e.output}'.format(
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1,4 +1,4 @@
-#    Copyright 2018 ARM Limited
+#    Copyright 2024 ARM Limited
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -25,7 +25,6 @@ import time
 import logging
 import posixpath
 import subprocess
-import sys
 import tarfile
 import tempfile
 import threading
@@ -35,7 +34,6 @@ import copy
 import inspect
 import itertools
 from collections import namedtuple, defaultdict
-from contextlib import contextmanager
 from past.builtins import long
 from past.types import basestring
 from numbers import Number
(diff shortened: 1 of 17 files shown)


devlib/target.py
```diff
@@ -1018,19 +1015,19 @@ async def read_bool(self, path):
...
         finally:
-            await self.write_value.asyn(path, orig_value, verify)
+            await self.write_value.asyn(path, orig_value, verify, as_root)
 
     def batch_revertable_write_value(self, kwargs_list):
         return batch_contextmanager(self.revertable_write_value, kwargs_list)
 
     @asyn.asyncf
-    async def write_value(self, path, value, verify=True):
+    async def write_value(self, path, value, verify=True, as_root=True):
```
douglas-raillard-arm, line 1030: Same comment as for `revertable_write_value()`

devlib/target.py
```diff
@@ -2995,7 +2992,7 @@ def __init__(self,
         adb_conn_params = ['device', 'adb_server', 'timeout']
         self.android_connection_settings = {}
         for setting in adb_conn_params:
-            if connection_settings.get(setting, None):
+            if setting in connection_settings:
```
douglas-raillard-arm, line 2995: this is not equivalent. If `setting` is available in `connection_settings` with a falsy value, the `setting in connection_settings` will still enter the branch.

OTH if we decide that we want falsy values to be propagated, we can do it with a comprehension:
```
self.android_connection_settings.update(
  (key, value)
  for key, value in connection_settings.items()
  if key in adb_conn_params
)
```

devlib/target.py
```diff
@@ -1018,19 +1015,19 @@ async def read_bool(self, path):
...
-    async def revertable_write_value(self, path, value, verify=True):
+    async def revertable_write_value(self, path, value, verify=True, as_root=True):
         orig_value = self.read_value(path)
         try:
-            await self.write_value.asyn(path, value, verify)
+            await self.write_value.asyn(path, value, verify, as_root)
             yield
         finally:
-            await self.write_value.asyn(path, orig_value, verify)
+            await self.write_value.asyn(path, orig_value, verify, as_root)
```
douglas-raillard-arm, line 1024: At this point we might want to pass the parameters using named syntax:
```
self.write_value.asyn(path, orig_value=orig_value, verify=verify, as_root=as_root)
```
Otherwise we run a relatively high risk of mixing parameters, and this can easily backfire especially for booleans (e.g. passing the value of `orig_value` to `verify` is likely to work, as a non-empty string and non-zero int will be considered as `True` (`if  x` is equivalent to `if bool(x)`))

douglas-raillard-arm (commented):
LGTM, just a couple of nits

devlib/target.py
```diff
@@ -2991,9 +2993,11 @@ def __init__(self,
         # Pull out adb connection settings
         adb_conn_params = ['device', 'adb_server', 'timeout']
         self.android_connection_settings = {}
-        for setting in adb_conn_params:
-            if connection_settings.get(setting, None):
-                self.android_connection_settings[setting] = connection_settings[setting]
+        self.android_connection_settings.update(
```
douglas-raillard-arm, line 2996: So we established we want to include the values that are falsy in `android_connection_settings` ? That sounds reasonable but we will need to go through the list of keys and check that for each of them a falsy value is valid and not signalling anything special. For example, we might have a higher layer that sets a bunch of None that would be expected to be filtered at this stage, or maybe not.

tests/test_target.py
```diff
@@ -14,35 +15,69 @@
...
+        method_name = self.__class__.test_read_multiline_values.__qualname__
 
-    def test_read_multiline_values(self):
         data = {
             'test1': '1',
             'test2': '2\n\n',
             'test3': '3\n\n4\n\n',
         }
 
-        tempdir = tempfile.mkdtemp(prefix='devlib-test-')
```
douglas-raillard-arm, line 18: I'd recommend using tempfile.TemporaryDirectory as a context manager for proper resource disposal. It also allows specifying a base path so we can make it create the folder in the target working dir as well if needed

tests/test_target.py
```diff
@@ -14,35 +15,69 @@
...
+                               f'devlib-test-{get_random_string(12)}')
+        print(f'{method_name}: creating {dirname}...')
+        target.makedirs(dirname)
+
         for key, value in data.items():
-            path = os.path.join(tempdir, key)
-            with open(path, 'w') as wfh:
-                wfh.write(value)
+            path = os.path.join(dirname, key)
+            str_value = value.replace("\n", "\\n")
```
douglas-raillard-arm, line 45: that's fishy, why do we need that ? If you want to have the string displayed as a Python literal, you can use the `repr()` of the string, which can be done with `{value!r}` in the f-string.

tests/test_target.py
```diff
@@ -14,35 +15,69 @@
...
 
-    def test_read_multiline_values(self):
         data = {
             'test1': '1',
             'test2': '2\n\n',
             'test3': '3\n\n4\n\n',
         }
 
-        tempdir = tempfile.mkdtemp(prefix='devlib-test-')
+        dirname = os.path.join(target.working_directory,
```
douglas-raillard-arm, line 38: looks like `target` is used but never defined

tests/test_target.py
```diff
@@ -14,35 +15,69 @@
...
 
-        t = LocalLinuxTarget(connection_settings={'unrooted': True})
-        raw_result = t.read_tree_values_flat(tempdir)
+        print(f'{method_name}: reading values from target...')
+        raw_result = target.read_tree_values_flat(dirname)
         result = {os.path.basename(k): v for k, v in raw_result.items()}
 
-        shutil.rmtree(tempdir)
+        print(f'{method_name}: removing {dirname}...')
+        target.remove(dirname)
```
douglas-raillard-arm, line 54: this should be done with tempfile.TemporaryDirectory used as a context manager. Both this new and initial code will fail to cleanup if anything raises an exception.

tests/test_target.py
```diff
@@ -14,35 +15,69 @@
...
 
         self.assertEqual({k: v.strip()
                           for k, v in data.items()},
                          result)
+
+
+if __name__ == '__main__':
```
douglas-raillard-arm, line 61: I don't know if it's a good idea to leave that. These tests are supposed to be used with a test runner like `pytest`, so whatever this bit does should be done in a unittest fixture/case.

devlib/target.py
douglas-raillard-arm: Ok sounds good thanks

devlib/target.py
douglas-raillard-arm: Question answered above and the result is yes we want to propagate falsy value, so the comprehension is appropriate

tests/test_target.py
metin-arm: Can `tempfile` create files/folder on remote machine via SSH? I thought it cannot and decided to use `Target.makedirs()`.

tests/test_target.py
metin-arm: I just wanted to log what is written as `value`. `{value!r}` is really cool. Thanks.

tests/test_target.py
metin-arm: Got it. I'll leave `test_target.py` as is and will create a new test case in another merge request.
Thanks.

tests/test_target.py
metin-arm: Will it be ok if I run these commands in try/except block?

tests/test_target.py
metin-arm: `target` is defined in the new line #28.

tests/test_target.py
douglas-raillard-arm: > Can tempfile create files/folder on remote machine via SSH?

It can't but tempfile was being used before so the test was never about remotely creating a folder is it ? Or was it using a local target and you generalized to any kind ? I'll have to look again at the test modification if you re-upload it (I think you dropped the commit or that part of the change).

Also it might be a good idea to introduce a NamedTemporaryFile/TemporaryDirectory equivalent to devlib, it should be a good simple exercise to do and can be done behind a single method, e.g.:
```
class Target:
    ...
    def make_temp(self, type, prefix=None):
         ...
``` 
with `type` being `"file"` or `"folder"`. The mktemp command of busybox can be used to safely create a unique file/folder and `contextlib.contextmanager` can be used to make a context manager.

tests/test_target.py
metin-arm: I dropped [07dd146](https://github.com/ARM-software/devlib/commit/07dd14616dc8d1ac8d2658d156baf1183924e53e) for now. So, `test/test_target.py` only has copyright change.
But, `target` was defined as input arg of `test_read_multiline_values()`.

metin-arm:
Hey @marcbonnici, can you please review this merge request? Thanks in advance.

metin-arm:
Hey @marcbonnici, can you please review this merge request? Thanks in advance.

marcbonnici (approved):
The current state LGTM, perhaps some of your initial test changes could be a separate PR?

@douglas-raillard-arm thanks for the reviews, are you happy with this as it is now?

metin-arm:
> The current state LGTM, perhaps some of your initial test changes could be a separate PR?

I only have a copyright change in tests as of now.
That said, will file another PR for further changes in tests.
Thanks.

douglas-raillard-arm:
@marcbonnici LGTM

PR #667, opened 2024-01-23
Integrate buildroot and add support for QEMU Linux target

metin-arm:
**Commits:**
* 918c19c - utils/ssh: Try to free up resources during client creation
* 3b7fb5b - target: Introduce make_temp() for creating temp file/folder on target
* 1253903 - tests/test_target: Test more targets
* 82a722d - tools/buildroot: Add support for generating Linux target system images
* a4b3547 - (HEAD -> staging, origin/staging) target: Implement target runner classes


tools/buildroot/generate-kernel-initrd.sh
```diff
@@ -0,0 +1,123 @@
+#! /bin/bash
```
douglas-raillard-arm, line 1: It would be best to convert that to Python. The only reason it's a shell script is historical and due to me not pushing back enough back in the days :)

You can use `argparse` from the stdlib for the CLI, but there are some others like `click` which is quite popular and possibly require less boilerplate

devlib/target.py
```diff
@@ -2892,6 +2895,99 @@ def _resolve_paths(self):
         if self.executables_directory is None:
             self.executables_directory = '/tmp/devlib-target/bin'
 
+class QEMULinuxTarget(LinuxTarget):
```
douglas-raillard-arm, line 2898: I don't think it's really an issue but I was wondering if we really want to make that QEMU runner a subclass of LinuxTarget. If we do that, we can easily end up with a combinatorial explosion of the number of subclasses with just a few "modifier" classes like that, and the naming nightmare that goes with it.

Maybe a better approach to the problem is to make that QEMU thingy a separate class (e.g. let's say a TargetRunner class with a QEMUTargetRunner subclass), and then just pass the target class to use as a parameter. The created target can be accessed as a `.target` attribute. This way we also avoid any possible clash between methods defined here and current + future methods of Target classes.

devlib/target.py
```diff
@@ -2892,6 +2895,99 @@ def _resolve_paths(self):
...
+    '''
+    Class for launching Linux target on QEMU.
+    '''
+
+    # pylint: disable=too-many-locals,too-many-arguments
+    def __init__(self,
+                 kernel_image,
```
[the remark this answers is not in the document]
metin-arm: Right. Will use `**kwargs`. Thanks.

devlib/target.py
```diff
@@ -2892,6 +2895,99 @@ def _resolve_paths(self):
...
+
+        qemu_executable = f'qemu-system-{arch}'
+        qemu_path = shutil.which(qemu_executable)
+        if qemu_path is None:
+            raise FileNotFoundError(f'Cannot find {qemu_executable} executable!')
+
+        if not os.path.exists(kernel_image):
+            raise FileNotFoundError(f'{kernel_image} does not exist!')
+
+        qemu_cmd = f'sudo {qemu_path} -kernel {kernel_image} -append "{cmdline}" ' \
```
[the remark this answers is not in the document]
metin-arm: What does `if (id -nG "$USER" | grep -qw "kvm"); then echo "$USER is in kvm group"; else echo "$USER is not in kvm group"; fi` print for you? I guess you are in `kvm` group and able to access `/dev/kvm`.
I'm working on another task such as Android emulator requires the user to be in `kvm` group.
Maybe I should remove this `sudo` and mention this requirement in `doc/target.rst`?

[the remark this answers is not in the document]
metin-arm: Good catch! Will fix it.

devlib/target.py
```diff
@@ -2892,6 +2895,99 @@ def _resolve_paths(self):
...
+                   f'-netdev user,id=net0,hostfwd=tcp::{connection_settings["port"]}-:22 ' \
+                   '-device virtio-net-pci,netdev=net0 --nographic'
+
+        if initrd_image:
+            if not os.path.exists(initrd_image):
+                raise FileNotFoundError(f'{initrd_image} does not exist!')
+            qemu_cmd = f'{qemu_cmd} -initrd {initrd_image}'
+
+        if arch.startswith('x86'):
+            qemu_cmd = f'{qemu_cmd} {"--enable-kvm" if enable_kvm else ""}'
```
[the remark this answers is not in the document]
metin-arm: I'll pass `enable-kvm` parameter only if `kvm-ok` can reports `KVM acceleration can be used`.

devlib/target.py
```diff
@@ -2892,6 +2895,99 @@ def _resolve_paths(self):
...
+        if connect:
+            super().connect(check_boot_completed=False)
+
+    def wait_boot_complete(self, timeout=10):
+        start_time = time.time()
+        boot_completed = False
+        poll_obj = select.poll()
+        poll_obj.register(self.qemu_process.stdout, select.POLLIN)
+
+        while not boot_completed and timeout >= (time.time() - start_time):
```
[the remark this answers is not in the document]
metin-arm: I'm unsure if it's necessary since `select.poll()` already waits for IO event.

devlib/target.py
```diff
@@ -2892,6 +2895,99 @@ def _resolve_paths(self):
...
+            poll_result = poll_obj.poll(0)
+            if poll_result:
+                line = self.qemu_process.stdout.readline().rstrip(b'\r\n')
+                line = line.decode(sys.stdout.encoding or 'utf-8', "replace") if line else ''
+                self.logger.debug(line)
+                if 'Welcome to Buildroot' in line:
```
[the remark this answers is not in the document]
metin-arm: I think it indicates SSH server is ready:
```Bash
ssh-keygen: generating new host keys: RSA ECDSA ED25519 
Starting sshd: OK

Welcome to Buildroot
```
But it's mainly for waiting boot completion (so the method name is `wait_boot_complete()`).

I did not like hard-coding the banner, too, but port scanning looks overkill. Maybe it should be parametric. If a user customizes the default image or has an existing image with a different banner, they can specify that.

devlib/target.py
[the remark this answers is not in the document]
metin-arm: `shlex.quote()` converts CPU list to `'-cpu cortex-a72 -cpu cortex-a53'` and QEMU is unhappy with those `'`.

tests/spin_targets.py
```diff
@@ -0,0 +1,103 @@
...
+
+    ll_target = LocalLinuxTarget(connection_settings={'unrooted': True},
+                                 working_directory='/tmp/devlib-target')
+    print(f'{ll_target.__class__.__name__}: {ll_target.os}/{ll_target.hostname}')
+    TestReadTreeValues().test_read_multiline_values(ll_target)
+    print(f'{ll_target.__class__.__name__}: removing {ll_target.working_directory}...')
+    ll_target.remove(ll_target.working_directory)
+
+    q_target = QEMULinuxTarget(kernel_image='/home/user/devlib/tools/buildroot/buildroot-v2023.11.1/output/images/Image',
+                               connection_settings={'host': '127.0.0.1',
```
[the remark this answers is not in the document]
metin-arm: devlib launches QEMU, but the kernel / rootfs image may be built externally.
The only interesting bit of `connection_settings` is `'strict_host_check': False`. Rest of the parameters can be inferred from default settings.

tools/buildroot/generate-kernel-initrd.sh
metin-arm: This is really simple shell script. If we ignore empty lines, copyright, global variables, and parameter parsing stuff, then it's in fact less than 10 lines which do trivial jobs (literally git clone/checkout, mkdir, cp, make). And I think a shell script is a good fit for this kind of basic work.

tests/spin_targets.py
```diff
@@ -0,0 +1,103 @@
...
+        finally:
+            print(f'{method_name}: removing {dirname}...')
+            target.remove(dirname)
+
+        self.assertEqual({k: v.strip()
+                          for k, v in data.items()},
+                         result)
+
+
+if __name__ == '__main__':
```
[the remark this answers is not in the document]
metin-arm: Sure. Will remove `__main__`. Thanks.

setrofim:
This should probably be implemented as a `Platform` rather than a `Target`, as this handles "hardware" rather than OS-level functionality; And that way, the same code can be used for Android  targets running on QEMU as well.

devlib/target.py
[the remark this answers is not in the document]
douglas-raillard-arm: > I guess you are in kvm group and able to access /dev/kvm.

Indeed. I tried to have a peek at how this is typically configured (e.g. on Debian) and it's quite complicated: https://unix.stackexchange.com/questions/599651/whats-the-purpose-of-kvm-libvirt-and-libvirt-qemu-groups-in-linux
Basically with systemd (well, logind) you also get the permission added to you if you logged in on a console even if you are not part of the kvm group. Considered all that, maybe we should remove the sudo and let the user enable kvm, which seems to be automatically done in a typical desktop system.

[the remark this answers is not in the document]
douglas-raillard-arm: How did you use it ? You need to apply it to each parameter individually, otherwise it will add escaping so that the entire string is a single argument. That's partly why it's not recommended to use `shell=True` in `subprocess`.  It's typically a much better idea to simply pass an iterable, which won't require any escaping at all.

devlib/target.py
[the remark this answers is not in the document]
douglas-raillard-arm: I don't think it's overkill, it's not even more code:
https://stackoverflow.com/questions/19196105/how-to-check-if-a-network-port-is-open
The banner will change if we try to use e.g. Debian, and it can probably include variable things like the date.

devlib/target.py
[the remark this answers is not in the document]
douglas-raillard-arm: indeed, apparently it blocks forever unless a timeout is provided.

douglas-raillard-arm:
@setrofim I'm not sure platform would be appropriate either, we want an object that starts its life before the `Target` and survives after. Platform seems to be more for managing platform-specific interaction with a device (such as rebooting) more than spawning the device itself.

setrofim:
> we want an object that starts its life before the Target and survives after

A `Platform` is passed a a parameter on `Target` creation, so it is already assumed to exist before the `Target` (only if one is not provided, the `Target` will then create a "default" one). And the `Target` doesn't do anything to "terminate" a `Platform`, so there is no reason why the latter can't persist after the former stopped existing.

> Platform seems to be more for managing platform-specific interaction with a device (such as rebooting) more than spawning the device itself.

Only because a device is not normally "spawned". However the interactions can include the initial target setup prior to execution (e.g. flashing images -- see VExpress platform). Support for Gem5 simulation (conceptually similar to QEMU) is implemented as a `Platform` and that includes optionally starting the simulation.

tests/spin_targets.py
```diff
@@ -0,0 +1,108 @@
...
+        """ Test Target.read_tree_values_flat() """
+
+        def run_test(target):
+            data = {
+                'test1': '1',
+                'test2': '2\n\n',
+                'test3': '3\n\n4\n\n',
+            }
+
+            print(f'{method_name}: target={target.__class__.__name__} os={target.os} ' \
```
douglas-raillard-arm, line 43: generally it's better to avoid breaking a string literal, as this prevents from using grep easily when looking for the source of a given message

tests/spin_targets.py
```diff
@@ -0,0 +1,108 @@
...
+                  f'hostname={target.hostname}')
+
+            dirname = os.path.join(target.working_directory,
+                                   f'devlib-test-{get_random_string(12)}')
+            print(f'{method_name}: creating {dirname}...')
+            target.makedirs(dirname)
```
douglas-raillard-arm, line 49: as discussed before, this should go in a context manager as a Target method

tests/spin_targets.py
```diff
@@ -0,0 +1,108 @@
...
+                    path = os.path.join(dirname, key)
+                    print(f'{method_name}: writing {value!r} to {path}...')
+                    target.write_value(path, value, verify=False,
+                                       as_root=target.conn.connected_as_root)
+
+                print(f'{method_name}: reading values from target...')
+                raw_result = target.read_tree_values_flat(dirname)
+                result = {os.path.basename(k): v for k, v in raw_result.items()}
+            except Exception as ex:
+                raise TargetStableError('Error while processing the tree!') from ex
```
douglas-raillard-arm, line 62: Not sure if this adds much, the original exception will have all the details we need to debug.

tests/spin_targets.py
```diff
@@ -0,0 +1,108 @@
...
+
+            self.assertEqual({k: v.strip()
+                              for k, v in data.items()},
+                             result)
+
+            print(f'{method_name}: removing {target.working_directory}...')
+            target.remove(target.working_directory)
+
+        def create_targets():
+            targets = []
```
douglas-raillard-arm, line 75: looks like this whole function could just be:
```
return [
    AndroidTarget(connection_settings={'device': 'emulator-5554'},
                                     working_directory='/data/local/tmp/devlib-target'),
...
]
```

instead of loads of append() calls

tests/spin_targets.py
```diff
@@ -0,0 +1,108 @@
...
+                                                                'strict_host_check': False},
+                                           target_params={'working_directory': '/tmp/devlib-target'})
+            targets.append(qemu_runner.target)
+
+            return targets
+
+        method_name = self.__class__.test_read_multiline_values.__qualname__
+
+        targets = create_targets()
+        for target in targets:
```
douglas-raillard-arm, line 106: it's probably good enough for now but when we write more tests, we probably don't want to create a new set of target every time, so we will want to use the unittest infrastructure to keep those alive. Or maybe not if it stinks too much, let's see later

devlib/target.py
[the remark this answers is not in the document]
metin-arm: Agreed. Removed `sudo` already.

devlib/target.py
```diff
@@ -2893,6 +2898,127 @@ def _resolve_paths(self):
...
+                   f'-netdev user,id=net0,hostfwd=tcp::{connection_settings["port"]}-:22 ' \
+                   '-device virtio-net-pci,netdev=net0 --nographic'
+
+        if qemu_params_full["initrd_image"]:
+            if not os.path.exists(qemu_params_full["initrd_image"]):
+                raise FileNotFoundError(f'{qemu_params_full["initrd_image"]} does not exist!')
+            qemu_cmd = f'{qemu_cmd} -initrd {quote(qemu_params_full["initrd_image"])}'
+
+        if qemu_params_full["arch"] == machine():
+            qemu_cmd = f'{qemu_cmd} {"--enable-kvm" if qemu_params_full["enable_kvm"] else ""}'
```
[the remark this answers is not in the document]
metin-arm: They are the indicator for param in formatted string.
Another version would be something like this:
```Python
if qemu_params_full["enable_kvm"]:
                qemu_cmd = f'{qemu_cmd} --enable-kvm'
```
Looks more readable I think :)

doc/target.rst
```diff
@@ -780,6 +780,102 @@ Android Target
    support the majority of newer devices.
 
 
+Target Runner
```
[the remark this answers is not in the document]
metin-arm: Just to confirm: you mean not updating `doc/target.rst` for new classes, instead put all docs into source code?

devlib/target.py
```diff
@@ -2893,6 +2898,127 @@ def _resolve_paths(self):
...
+        try:
+            self.runner_process = get_subprocess(self.runner_cmd, shell=True)
+        except Exception as ex:
+            raise HostError(f'Error while running "{self.runner_cmd}": {ex}') from ex
+
+        atexit.register(self.terminate_runner)
+
+        self.wait_boot_complete()
+
+        self.target.connect(check_boot_completed=False)
```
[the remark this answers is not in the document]
metin-arm: Yes, default is `True`. Nothing fails since previous line (`self.wait_boot_completed()`) already waits for boot completion.
Leaving `check_boot_completed` as `True` just causes an extra function call.
I'll revisit this after addressing https://github.com/ARM-software/devlib/pull/667#discussion_r1467785241.

devlib/target.py
```diff
@@ -2893,6 +2928,249 @@ def _resolve_paths(self):
...
+        * ``cmdline``: Kernel command line parameter. It only specifies console device in
+            default (i.e., ``console=ttyAMA0``) which is valid for Arm architectures.
+            May be changed to ``ttyS0`` for x86 platforms.
+
+        * ``enable_kvm``: Specifies if KVM will be used as accelerator in QEMU or not.
+            Enabled by default if host architecture matches with target's for improving
+            QEMU performance.
+    :type qemu_params: Dict
+
+    :param connection_settings: he dictionary which stores connection settings
```
marcbonnici, line 3081: nit: Typo `he dictionary` -> `Dict to store..`

devlib/target.py
```diff
@@ -2893,6 +2928,249 @@ def _resolve_paths(self):
...
+            'initrd_image': '',
+            'mem_size': 512,
+            'num_cores': 2,
+            'num_threads': 2,
+            'cmdline': 'console=ttyAMA0',
+            'enable_kvm': True,
+        }
+
+        # Update default QEMU parameters with :param:`qemu_params`.
+        qemu_default_args.update(
```
marcbonnici, line 3125: nit: Would making `qemu_default_args` to just be named `qemu_args` be a bit clearer in the following code that it is taking in account the user provided information as well?

devlib/target.py
```diff
@@ -1071,6 +1072,38 @@ async def write_value(self, path, value, verify=True, as_root=True):
             else:
                 raise
 
+    @contextmanager
+    def make_temp(self, is_directory=True, directory='', prefix='devlib-test'):
```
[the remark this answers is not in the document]
metin-arm: But it's working directory?

tests/test_target.py
```diff
@@ -37,23 +36,57 @@ def build_targets():
 
     targets = []
 
+    if target_configs.get('AndroidTarget') is not None:
+        print('> Android targets:')
```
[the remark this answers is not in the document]
metin-arm: They are dummy prints. Not sure if it's worth. Please let me know if you have a strong opinion to use logger.

tests/test_target.py
```diff
@@ -62,15 +95,26 @@ def test_read_multiline_values(target):
...
 
-    tempdir = tempfile.mkdtemp(prefix='devlib-test-')
-    for key, value in data.items():
-        path = os.path.join(tempdir, key)
-        with open(path, 'w', encoding='utf-8') as wfh:
-            wfh.write(value)
+    print(f'target={target.__class__.__name__} os={target.os} hostname={target.hostname}')
+
+    with target.make_temp() as tempdir:
+        print(f'created {tempdir}.')
```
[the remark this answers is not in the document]
metin-arm: Fixed.

tests/target_configs.yaml
```diff
@@ -1,5 +1,30 @@
...
+
 LocalLinuxTarget:
   entry-0:
     connection_settings:
       unrooted: True
 
+QEMUTargetRunner:
+  entry-0:
+    qemu_params:
+      kernel_image: '/devlib/tools/buildroot/buildroot-v2023.11.1-aarch64/output/images/Image'
```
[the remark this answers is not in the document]
metin-arm: OK.

devlib/target.py
metin-arm: Done.

devlib/target.py
```diff
@@ -2893,6 +2928,249 @@ def _resolve_paths(self):
             self.executables_directory = '/tmp/devlib-target/bin'
 
 
+class TargetRunner:
```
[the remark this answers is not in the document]
metin-arm: Yep, makes sense.

tools/buildroot/generate-kernel-initrd.sh
```diff
@@ -0,0 +1,118 @@
...
+	echo "		-a: set arch (default is aarch64, x86_64 is also supported)"
+	echo "		-p: purge buildroot to force a fresh build"
+	echo "		-h: print this help message"
+}
+
+function set_arch
+{
+	if [[ "${1}" == "aarch64" || "${1}" == "x86_64" ]]; then
+		ARCH=${1}
+		KERNEL_IMAGE_NAME="bzImage"
```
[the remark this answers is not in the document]
metin-arm: Good catch. Thanks.

devlib/target.py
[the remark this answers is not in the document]
marcbonnici: In my mind the function is designed to create a new temp directory/file under the path provided from the user. The default location is just set to the working directory.

Having said this, after checking the documentation for `tempfile.mkstemp` I see they use `dir` as parameter name so not a strong opinion on this.

tests/target_configs.yaml
```diff
@@ -1,5 +1,30 @@
+AndroidTarget:
```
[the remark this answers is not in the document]
marcbonnici: Sorry, I didn't mean it would help in automating android emulators.

I was thinking that for a user who just wants to check out devlib and run the tests to validate some changes they made, it could be useful to provide a "default" config that just works out of the box with no additional setup required. (I'm thinking going forward if we have a greater set of tests to run). 

The "example" version would then provide details of all the available options including the android emulators that requrie manually starting to provide a more comprehensive testing setup.

tests/test_target.py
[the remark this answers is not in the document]
marcbonnici: That's a fair point, if they are not shown by default this should be ok.

marcbonnici (approved):

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
...
+        self.connection_settings = {
+            'host': '127.0.0.1',
+            'port': 8022,
+            'username': 'root',
+            'password': 'root',
+            'strict_host_check': False,
+        }
+
+        if connection_settings is not None:
+            self.connection_settings = self.connection_settings | connection_settings
```
douglas-raillard-arm, line 211: Union of dict is brought by [PEP 584](https://peps.python.org/pep-0584/), which was implemented in Python 3.8. For now, devlib supports down to 3.7. The backward-compatible equivalent would be:
```
        if connection_settings is not None:
            self.connection_settings = {**self.connection_settings, **connection_settings}
``` 
and you can further avoid a separate path with:
```
self.connection_settings = {**self.connection_settings, **(connection_settings or {})}
```

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
...
+
+        qemu_args = {
+            'kernel_image': '',
+            'arch': 'aarch64',
+            'cpu_type': 'cortex-a72',
+            'initrd_image': '',
```
douglas-raillard-arm, line 217: It seems the consuming code checks to see if the value is fals-y. This could be avoided by simply not having the key by default. That typically results in cleaner code, as "having a value" means "the value must be used", and "not having a value" will result in a KeyError that can be explicitly handled. Otherwise, you end up with in-band signaling (some values are "good to use" and others are "not good to use"), on top of always having the risk of a KeyError because it's just a dict after all.

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
...
+            'mem_size': 512,
+            'num_cores': 2,
+            'num_threads': 2,
+            'cmdline': 'console=ttyAMA0',
+            'enable_kvm': True,
+        }
+
+        qemu_args = qemu_args | qemu_settings
```
douglas-raillard-arm, line 225: ditto on PEP 584

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
...
+        qemu_executable = f'qemu-system-{qemu_args["arch"]}'
+        qemu_path = which(qemu_executable)
+        if qemu_path is None:
+            raise FileNotFoundError(f'Cannot find {qemu_executable} executable!')
+
+        if not os.path.exists(qemu_args["kernel_image"]):
+            raise FileNotFoundError(f'{qemu_args["kernel_image"]} does not exist!')
+
+        # pylint: disable=consider-using-f-string
+        qemu_cmd = '''\
```
douglas-raillard-arm, line 236: the `\` does not seem to really serve any purpose does it ?

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
...
+
+        qemu_executable = f'qemu-system-{qemu_args["arch"]}'
+        qemu_path = which(qemu_executable)
+        if qemu_path is None:
+            raise FileNotFoundError(f'Cannot find {qemu_executable} executable!')
+
+        if not os.path.exists(qemu_args["kernel_image"]):
+            raise FileNotFoundError(f'{qemu_args["kernel_image"]} does not exist!')
+
+        # pylint: disable=consider-using-f-string
```
douglas-raillard-arm, line 235: I'd agree with pylint here, that would make the code more readable. If that was not possible for whatever reason, I'd still use the keyword arguments of format() to provide named placeholders, rather than positional ones. Otherwise it's very easy to shift everything by one location and break the code, making it needlessly harder to modify in the future

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
...
+            self.connection_settings["port"],
+        )
+
+        if qemu_args["initrd_image"]:
+            if not os.path.exists(qemu_args["initrd_image"]):
+                raise FileNotFoundError(f'{qemu_args["initrd_image"]} does not exist!')
+
+            qemu_cmd += f' -initrd {qemu_args["initrd_image"]}'
+
+        if qemu_args["arch"] == machine():
```
douglas-raillard-arm, line 255: Are we sure that whatever Python returns in platform.machine() actually agrees with qemu arch names ? This is typically a can of worm, where half the world uses `arm64`, the other half `aarch64` and I'm not even talking about 32bits arm. x86 is usually ok, but even there you can end up with `x86-64` vs `x86_64`

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
...
+            if qemu_args["enable_kvm"]:
+                qemu_cmd += ' --enable-kvm'
+        else:
+            qemu_cmd += f' -machine virt -cpu {qemu_args["cpu_type"]}'
+
+        self.target = make_target(connect=False,
+                                  conn_cls=SshConnection,
+                                  connection_settings=self.connection_settings)
+
+        super().__init__(runner_cmd=qemu_cmd,
```
douglas-raillard-arm, line 265: All user input that makes it to qemu_cmd must be escaped using shlex.quote(). I know this is cumbersome, but the only other way to deal with that problem is to ask for the command as a list like `subprocess` does by default. `devlib` does not for various reasons (although we could easily add such an API, maybe we should ...)

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
...
+
+            qemu_cmd += f' -initrd {qemu_args["initrd_image"]}'
+
+        if qemu_args["arch"] == machine():
+            if qemu_args["enable_kvm"]:
+                qemu_cmd += ' --enable-kvm'
+        else:
+            qemu_cmd += f' -machine virt -cpu {qemu_args["cpu_type"]}'
+
+        self.target = make_target(connect=False,
```
douglas-raillard-arm, line 261: assigning to self.target is the responsibility of the base class's `__init__`.

devlib/utils/ssh.py
```diff
@@ -367,25 +367,32 @@ def __init__(self,
...
-            self.execute(
-                # sudo -n is broken on some versions on MacOSX, revisit that if
-                # someone ever cares
-                'sudo -n true || echo NEED_PASSWORD',
-                as_root=False,
-                check_exit_code=False,
+        self.client = None
+        try:
+            self.client = self._make_client()
+            atexit.register(self.close)
```
douglas-raillard-arm, line 373: unrelated to this PR but we should probably fix this: if we keep creating connection objects (and we do because of the multithreading support), we end up:
* accumulating atexit handlers without ever removing them
* leaking the connection object, as a reference to it will stay alive in the method's `__self__`

let me know if you want to get your hands dirty in low level Python, otherwise I'll do it

devlib/utils/ssh.py
```diff
@@ -367,25 +367,32 @@ def __init__(self,
...
+                    # sudo -n is broken on some versions on MacOSX, revisit that if
+                    # someone ever cares
+                    'sudo -n true || echo NEED_PASSWORD',
+                    as_root=False,
+                    check_exit_code=False,
+                )
             )
-        )
+
+        except BaseException:
```
douglas-raillard-arm, line 392: This is one of the cases where `BaseException` is appropriate: we want to cleanup the situation as far as we can in case we get interrupted, even if the user asked for the interruption. Since we re-raise the exception, we are not going to accidentally mask it. The only bad thing I could see happen is this code being called from a loop that retries in case of failures. If `__init__` failed because the user used ctrl-c, leading to a broken state in the client, leading to client.close() to raise an exception. Then this exception would propagate, and the retry logic would swallow it and retry. The fix would be ensuring the initial exception is always preserved:
```
        except BaseException as e:
            try:
                if self.client is not None:
                    self.client.close()
            except BaseException:
                raise e
```

tests/test_target.py
```diff
@@ -37,23 +36,57 @@ def build_targets():
 
     targets = []
 
+    if target_configs.get('AndroidTarget') is not None:
+        print('> Android targets:')
+        for entry in target_configs['AndroidTarget'].values():
+            pp(entry)
+            a_target = AndroidTarget(
+                connection_settings=entry['connection_settings'],
+                conn_cls=lambda **kwargs: AdbConnection(adb_as_root=True, **kwargs),
```
douglas-raillard-arm, line 45: Not an issue per say here, but FYI `functools.partial()` does the same as that lambda, somewhat better in minor ways

tests/test_target.py
douglas-raillard-arm, line 40: I'd still use a logger there. If you don't, any code that collects the log from logging but not stdout will miss this part. `pytest` has some integration with logging as it can set the log level, which can also be useful. If we really want to be lazy here, you can still use `logging.info` and that won't be much more to type than `print`

tests/test_target.py
```diff
@@ -37,23 +36,57 @@ def build_targets():
...
             pp(entry)
             ll_target = LocalLinuxTarget(connection_settings=entry['connection_settings'])
-            targets.append(ll_target)
+            targets.append((ll_target, None))
+
+    if target_configs.get('QEMUTargetRunner') is not None:
+        print('> QEMU target runners:')
+        for entry in target_configs['QEMUTargetRunner'].values():
+            pp(entry)
+            qemu_settings = entry.get('qemu_settings') and entry['qemu_settings']
```
douglas-raillard-arm, line 67: the `and ...` part is useless, `dict.get()` already supports providing a default value if the key does not exist (that's the only reason this function exists), and that default defaults to `None`. So you'll get the exact same result by dropping the `and ...`

tests/test_target.py
```diff
@@ -37,23 +36,57 @@ def build_targets():
...
+            connection_settings = entry.get(
```
douglas-raillard-arm, line 68: ditto

tests/test_target.py
```diff
@@ -37,23 +36,57 @@ def build_targets():
...
+                'connection_settings') and entry['connection_settings']
+
+            qemu_runner = QEMUTargetRunner(
+                qemu_settings=qemu_settings,
+                connection_settings=connection_settings,
+            )
+            targets.append((qemu_runner.target, qemu_runner))
```
douglas-raillard-arm, line 75: is there a point in returning tuples rather than just the runner, considered you can get the target by using `runner.target` ?

tests/test_target.py
```diff
@@ -37,23 +36,57 @@ def build_targets():
...
+            l_target = LinuxTarget(connection_settings=entry['connection_settings'])
+            targets.append((l_target, None))
+
     if target_configs.get('LocalLinuxTarget') is not None:
         print('> LocalLinux targets:')
         for entry in target_configs['LocalLinuxTarget'].values():
             pp(entry)
             ll_target = LocalLinuxTarget(connection_settings=entry['connection_settings'])
-            targets.append(ll_target)
+            targets.append((ll_target, None))
```
douglas-raillard-arm, line 61: Returning `None` is a bad idea: https://github.com/ARM-software/devlib/pull/669/files#r1541156876
I do note that you avoided your own trap, so hats off :) , but I can guarantee you'll fall in it at one point or another without type checking (I've never seen anyone not fall in it, me included)

The cleanest approach here is to simply create a no-op target runner that implements the API and does nothing apart from exposing a `.target` attribute. This way, the rest of the code does not need to care about special cases, and if part of the API is impossible to implement, that will just have revealed a design problem that needs to be addressed.

tests/test_target.py
```diff
@@ -62,15 +95,26 @@ def test_read_multiline_values(target):
...
+        raw_result = target.read_tree_values_flat(tempdir)
+        result = {os.path.basename(k): v for k, v in raw_result.items()}
 
-    raw_result = target.read_tree_values_flat(tempdir)
-    result = {os.path.basename(k): v for k, v in raw_result.items()}
+    print(f'Removing {target.working_directory}...')
+    target.remove(target.working_directory)
 
-    shutil.rmtree(tempdir)
+    if target_runner is not None:
```
douglas-raillard-arm, line 116: the `None` test can be dropped once the case disappears.

tests/test_target.py
```diff
@@ -62,15 +95,26 @@ def test_read_multiline_values(target):
...
+        print('Terminating target runner...')
+        target_runner.terminate()
```
douglas-raillard-arm, line 118: `target_runner` is supposed to be used as a context manager. There might be some legitimate uses for having a separate `terminate()` method so I'm not saying we should necessarily remove it, but that specific use-site is not one.

devlib/target.py
```diff
@@ -1071,6 +1072,38 @@ async def write_value(self, path, value, verify=True, as_root=True):
             else:
                 raise
 
+    @contextmanager
+    def make_temp(self, is_directory=True, directory='', prefix='devlib-test'):
+        """
+        Creates temporary file/folder on target and deletes it once it's done.
+
+        :param is_directory: Specifies if temporary object is a directory, defaults to True.
+        :type is_directory: bool, optional
```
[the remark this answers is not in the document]
metin-arm: I do see `, optional` on https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html. Is that wrong doc for Sphinx?
How do you verify docstring?

devlib/target.py
```diff
@@ -1071,6 +1072,38 @@ async def write_value(self, path, value, verify=True, as_root=True):
...
+
+        :param directory: Temp object will be created under this directory,
```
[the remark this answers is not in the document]
metin-arm: `tempfile.mkstemp` also uses `dir`. Hence, would keep it as is.

devlib/target.py
```diff
@@ -1071,6 +1072,38 @@ async def write_value(self, path, value, verify=True, as_root=True):
...
+        :type prefix: str, optional
+
+        :yield: Full path of temp object.
+        :rtype: str
+        """
+
+        directory = directory or self.working_directory
+        temp_obj = None
+        try:
+            cmd = f'mktemp -p {directory} {prefix}-XXXXXX'
```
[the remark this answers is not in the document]
metin-arm: Done.

devlib/target.py
```diff
@@ -1071,6 +1072,38 @@ async def write_value(self, path, value, verify=True, as_root=True):
...
+        Creates temporary file/folder on target and deletes it once it's done.
+
+        :param is_directory: Specifies if temporary object is a directory, defaults to True.
+        :type is_directory: bool, optional
+
+        :param directory: Temp object will be created under this directory,
+            defaults to :attr:`Target.working_directory`.
+        :type directory: str, optional
+
+        :param prefix: Prefix of temp object's name, defaults to 'devlib-test'.
```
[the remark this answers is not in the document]
metin-arm: Done.

devlib/target.py
```diff
@@ -1071,6 +1072,38 @@ async def write_value(self, path, value, verify=True, as_root=True):
...
+    @contextmanager
+    def make_temp(self, is_directory=True, directory='', prefix='devlib-test'):
+        """
+        Creates temporary file/folder on target and deletes it once it's done.
+
+        :param is_directory: Specifies if temporary object is a directory, defaults to True.
+        :type is_directory: bool, optional
+
+        :param directory: Temp object will be created under this directory,
+            defaults to :attr:`Target.working_directory`.
```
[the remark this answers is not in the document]
metin-arm: Done.

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
...
+
+        start_time = time.time()
+        elapsed = 0
+        while self.boot_timeout >= elapsed:
+            try:
+                self.target.connect(timeout=self.boot_timeout - elapsed)
+                self.logger.info('Target is ready.')
+                return
+            # pylint: disable=broad-except
+            except BaseException as ex:
```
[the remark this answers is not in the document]
metin-arm: Replaced it with `Exception`.

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
...
+            self.runner_process.stderr.close()
+
+            if self.runner_process.poll() is None:
+                self.logger.debug('Terminating target runner...')
+                os.killpg(self.runner_process.pid, signal.SIGTERM)
+                # Wait 3 seconds before killing the runner.
+                self.runner_process.wait(timeout=3)
+        except subprocess.TimeoutExpired:
+            self.logger.info('Killing target runner...')
+            os.killpg(self.runner_process.pid, signal.SIGKILL)
```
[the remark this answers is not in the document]
metin-arm: We first try to terminate the process gracefully, and then forcefully kill it if the process does not exit in 3 seconds.

Replaced the logic in `terminate()` with `Popen.kill()` for now. Did you mean removing `wait_boot_complete()` as well and use `Popen` as context manager?

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
...
+    :Variable positional arguments: Forwarded to :class:`TargetRunner`.
+
+    :raises FileNotFoundError: if QEMU executable, kernel or initrd image cannot be found.
+    """
+
+    def __init__(self,
+                 qemu_settings,
+                 connection_settings=None,
+                 # pylint: disable=unnecessary-lambda
+                 make_target=lambda **kwargs: LinuxTarget(**kwargs),
```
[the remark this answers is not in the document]
metin-arm: Done.

devlib/target_runner.py
metin-arm: Done.

devlib/target_runner.py
metin-arm: Done.

devlib/target_runner.py
metin-arm: Prevents prepending leading spaces in following line.

tests/test_target.py
metin-arm: My bad. Forgot `dict.get()` here.

tests/test_target.py
metin-arm: Done.

tests/test_target.py
metin-arm: Will keep it as is.

devlib/target_runner.py
metin-arm: Agreed. Removed empty values.

devlib/target_runner.py
metin-arm: Changed it to f-string.

devlib/target_runner.py
metin-arm: I thought this is addressed by `self.runner_process = get_subprocess(list(runner_cmd.split()))` in `TargetRunner` class?

devlib/target_runner.py
metin-arm: Done.

devlib/utils/ssh.py
metin-arm: Will leave up to you :)

devlib/utils/ssh.py
metin-arm: Included this fix with pylint enhancements.

tests/test_target.py
metin-arm: Done.

devlib/target_runner.py
metin-arm: Relaxed the check by `machine().startswith('x86')`.

devlib/target_runner.py
```diff
@@ -0,0 +1,267 @@
+#    Copyright 2024 ARM Limited
```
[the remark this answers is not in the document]
metin-arm: Done.

tests/test_target.py
metin-arm: You cannot if we want to test ChromeOS with QEMU, i.e., `targets.append((c_target, qemu_runner))`.

tests/test_target.py
metin-arm: Yes, but `TargetRunner` is created in `build_targets()` and terminated at the end of `test_read_multiline_values()`.

tests/test_target.py
metin-arm: Done.

tests/test_target.py
metin-arm: Done - kind of.
Let's continue discussion on the PR that I'm gonna file for addressing these comments.

metin-arm:
@douglas-raillard-arm, https://github.com/ARM-software/devlib/pull/677 is filed to address your comments except open questions here.

devlib/target.py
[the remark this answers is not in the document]
douglas-raillard-arm: That seems to be the right doc, but I think that either the doc is wrong or it's actually the doc on how to contribute to Sphinx itself, which may use something else than plain autodoc. That syntax is part of the [numpy style](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard), which requires the napoleon Sphinx extension. Apparently that napoleon extension pre-processes the docstring before autodoc sees it.

>  How do you verify docstring? 

You build the doc and check for warnings (`lisa-doc-build` for LISA). I tried by modifying a docstring in LISA. Since devlib's doc is not configured with autodoc, building the doc won't do anything as nothing will touch those docstrings.

Either way, if the function accepts a type T or `None` (which is the standard way of making a parameter "optional" in the common understanding of the term[1]), then it's actually more accurate to state `T or None`. 

[1] You can make a parameter actually optional without any default value with the variable keyword argument syntax, but that's definitely not something you'd do on a daily basis.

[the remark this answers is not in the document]
douglas-raillard-arm: > You build the doc and check for warnings
Another way would be to setup a test sphinx project, which is very easy too. Or port devlib to use autodoc. It's not particularly difficult, just an invasive PR (all existing function documentations need to be moved back as docstrings). If you were to do that, it would probably take ~1week (or maybe less) and I think it would be reasonable to as @marcbonnici  to freeze merging any other stuff so you don't end up with conflicts while working on it.

devlib/target.py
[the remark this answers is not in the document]
douglas-raillard-arm: Apparently I can't close comments (maybe because the PR is already merged), I'll just add an "ack" comment.

Ack

tests/test_target.py
douglas-raillard-arm: The point where it is created is not a problem, test_read_multiline_values() could be implemented as:
```python
def test_read_multiline_values(..., target_runner):
    with target_runner as target_runner:
       do_stuff()
```
That also makes me realize that currently, we will spin a fresh set of target runners for every single test, so we will need to fix that. I'm not sure of how it's best done with pytest, but you can probably find some inspiration online on how to deal with database connections (which are typically used as a context manager).

devlib/target.py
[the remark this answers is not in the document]
metin-arm: I can see `Resolve conversation` button under comments. Let me try resolving this conversation :)

devlib/target_runner.py
douglas-raillard-arm: AFAIK having the space or not won't change anything. If it does, it's probably a bug in devlib. You can also just use `.strip()` on the final string if you really needed to trim it. And if you want to preserve normal code indentation level in a multiline string without wanting the indentation inside the string itself, you can use `textwrap.dedent`

devlib/target_runner.py
douglas-raillard-arm: Is it correct though ? I know that for now arm64 hosts are not _really_ supported in LISA, but devlib might have a different standing there, and we also don't want to prevent it from working the day the rest (e.g. adb) is actually ready.

devlib/target_runner.py
douglas-raillard-arm: > get_subprocess(list(runner_cmd.split()))

This is broken. Either the `runner_cmd` is already a list and nothing needs to be split, or it's not and it needs to be split with `shlex.split()`. Just using `str.split()` will wreck any escaping that was applied. So considered the situation, we should be manipulating only lists here, and `qemu_cmd` should be a list as well. We _could_ use strings and then use `shlex.split()` but that's pointless and will just require extra work to create the string (shlex.quote on every single user input)

devlib/utils/ssh.py
douglas-raillard-arm: This is a bug that needs to be fixed separately. When `__init__` is called, the object is already created, as `__init__` is not a constructor per say, it only initializes an already-constructed object. That means that even if `__init__` ends up raising, `__del__` will still be called. So `__del__` needs to be prepared to deal with an object where `__init__` never ran (or e.g. raised on the first line before setting any attribute).

I'm not quite sure why that problem is masked by your initial code though, but I suppose it doesn't really matter

douglas-raillard-arm: EDIT: actually there is an issue with my code, it should be:
```python
        except BaseException as e:
            try:
                if self.client is not None:
                    self.client.close()
            finally:
                raise e
```
The reason you saw that issue is because the exception was discarded and never re-raised in the self.client.close() path. This means that the metaprogramming `__del__` relies on told it that the object finished initializing cleanly, but in reality  it did not and was missing an attribute.

devlib/target.py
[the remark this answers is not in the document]
metin-arm: I will not go for updating entire devlib repo now :)
Would you be OK if I just remove `, optional` part from docstring? Otherwise, it does not seem trivial to verify docstrings' Sphinx compatibility in devlib.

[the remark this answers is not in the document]
douglas-raillard-arm: Yes just go for replacing `, optional` by `or None`

devlib/utils/ssh.py
metin-arm: Applied this patch.

devlib/target_runner.py
metin-arm: Having leading spaces is not a functional issue. Just a cosmetic one.
Will use `textwrap.dedent`.

devlib/target_runner.py
metin-arm: This is the new version of the code in #677:
```Python
        if qemu_args['arch'].startswith('x86') and machine().startswith('x86'):
            if qemu_args["enable_kvm"]:
                qemu_cmd += ' --enable-kvm'
        else:
            qemu_cmd += f' -machine virt -cpu {qemu_args["cpu_type"]}'
```
I want to enable KVM only for x86 and specify `-machine virt -cpu ...` only if the target machine/guest is Arm.
Will this hurt devlib's current arm64 support?

devlib/target_runner.py
metin-arm: Converted `qemu_cmd` to a list of string/parameters.

tests/test_target.py
metin-arm: Implemented separate initialize / tear down routines and made other improvements on #677 . However, I still don't use context manager. Because targets would be terminated after the first test case (I know there is only 1 test case for now). Let's continue reviewing there if you don't have any other open comments here.

devlib/target_runner.py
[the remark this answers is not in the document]
douglas-raillard-arm: You could call `Popen.__exit__` in terminate. That won't forcefully kill it though, so if it needs a signal to exit, you can simplify all that into:
```
p = self.runner_process
p.kill(...)
p.__exit__(None, None, None)
```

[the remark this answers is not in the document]
metin-arm: Done on #677 .

devlib/target_runner.py
douglas-raillard-arm: > I want to enable KVM only for x86

Is KVM not working on aarch64 ?

> specify -machine virt -cpu ... only if the target machine/guest is Arm

What does `-machine virt` do that we don't want done on x86 ?

tests/test_target.py
douglas-raillard-arm: As commented below/offline, this should be achievable with an implementation of a runner that allows sharing a target with another runner and only disposing of the Target once both are finished with it.

devlib/target_runner.py
metin-arm: > Is KVM not working on aarch64 ?

Works for sure, but there are some restrictions: https://stackoverflow.com/a/71220201.

> What does -machine virt do that we don't want done on x86 ?

`-machine virt` is unsupported for x86_64. `qemu-system-x86_64 -machine help | grep -i virt` returns no match.

douglas-raillard-arm: I had a look at those restrictions and they don't seem to be too problematic. No-one is running an arm 32bits host these days for any CI system AFAIK, and `-cpu host -machine gic-version=host` does not sound too bad considered the speed benefits (unless this has a consequence I can't imagine).

 >-machine virt is unsupported for x86_64
 
Fair enough, so that's really attached to x86 target itself and not assuming the host is x86.

Make sure that in both cases, you add a comment stating the reason why this code takes those decisions so it can be revisited easily in the future (e.g. one day x86_64 might support virt).

metin-arm: `-cpu host` and `gic-version` options are not supported for `qemu-system-aarch64` on an `x86_64` host.
Also https://wiki.ubuntu.com/ARM64/QEMU says that it's not possible to use KVM accelerator for a different architecture (e.g., x86_64 host, aarch64 guest) which makes sense to me.

metin-arm: Made a small improvement regarding the host/guest checks. If we don't have any open comments left here, let's continue reviewing on #677 please.

PR #668, opened 2024-01-25
tools/android: Add support for creating Android virtual devices

metin-arm:
tools/android: Add support for creating Android virtual devices
Introduce ``tools/android/install_base.sh`` [1] script to install
Android command line tools including necessary platforms and
system-images for Linux and create Android Virtual Devices (AVD) for
Pixel 6 with Android v12 & v14 as well as an Android virtual *desktop*
device (v13) for ChromeOS tests.

[1] Forked from https://github.com/ARM-software/lisa/blob/main/install_base.sh


tools/android/install_base.sh
```diff
@@ -0,0 +1,295 @@
...
+    call_android_sdk avdmanager "$@"
+}
+
+# Needs install_android_sdk_manager first
+install_android_tools() {
+    yes | call_android_sdkmanager --verbose --channel=0 --install "platform-tools"
+    yes | call_android_sdkmanager --verbose --channel=0 --install "platforms;android-31"
+    yes | call_android_sdkmanager --verbose --channel=0 --install "platforms;android-34"
+    yes | call_android_sdkmanager --verbose --channel=0 --install "system-images;android-31;google_apis;${HOST_ARCH}"
+    yes | call_android_sdkmanager --verbose --channel=0 --install "system-images;android-34;google_apis;${HOST_ARCH}"
```
metin-arm, line 111: I use `HOST_ARCH` here, but I believe Android SDK supports only `arm64-v8a` Arm variant. Maybe I should use `arm64-v8a` if the host is not an Intel machine.

marcbonnici (approved):

tools/android/install_base.sh
```diff
@@ -0,0 +1,314 @@
+#!/usr/bin/env bash
```
douglas-raillard-arm, line 1: This file is essentially a copy/paste of the one from LISA. That creates a maintenance problem as this needs to evolve along how the android SDK manager is distributed, and that thing _does_ evolve. On top of that I can see that even from day 0 of that PR, you made improvement to the script such as building a URL from a version number, without making the equivalent PR in LISA to keep them in sync:
```
    local url="https://dl.google.com/android/repository/commandlinetools-linux-${CMDLINE_VERSION}_latest.zip" 
```

So that stuff needs to be deduplicated with what we have in LISA. I think it would be reasonable to call the devlib script from lisa's install_base.sh and dedup it this way unless they really need to share some internal state.

However, once this happens, we need to be able to be reactive when things change and break. @marcbonnici would you be happy to grant me merge permission in the devlib repo ? I'd only use it for minor stuff that require short reaction time like fixing a broken SDK install procedure (e.g. a broken URL), or typos and such. Normal PRs would keep following the current review process.

tools/android/install_base.sh
```diff
@@ -0,0 +1,314 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Forked from https://github.com/ARM-software/lisa/blob/main/install_base.sh
+#
+
+# shellcheck disable=SC2317
```
douglas-raillard-arm, line 22: If that's the only violation found by shellcheck I'm glad :) (although I don't know how deep shellcheck actually looks)

tools/android/install_base.sh
douglas-raillard-arm: EDIT: there is a whole range of helper functions that we definitely don't want to duplicate in here, such as test_os_release(). Maybe we should just source the devlib's script from LISA, and ensure that what we source "does nothing", i.e. just defines functions that we then call. devlib can have it's own entry point to call the functions, and the shared script would be a "library script".

tools/android/install_base.sh
```diff
@@ -0,0 +1,314 @@
...
+    local value=${2}
+
+    if [[ "$(read_os_release "${field_name}")" == "${value}" ]]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+function set_host_arch
```
douglas-raillard-arm, line 45: generic name for something not generic, this should be renamed with a more specific name.

tools/android/install_base.sh
```diff
@@ -0,0 +1,314 @@
...
+{
+	# Google ABI type for Arm platforms
+	HOST_ARCH="arm64-v8a"
```
douglas-raillard-arm, line 48: I don't think we should call that `HOST_ARCH`. I'd expect at the very least such variable to match the devlib definition of abis, which would use `arm64` in this case (see `devlib.utils.misc.ABI_MAP`). Maybe `ANDROID_SDK_HOST_ARCH` or something like that would be more appropriate.

tools/android/install_base.sh
```diff
@@ -0,0 +1,314 @@
...
+
+	local machine
+	machine=$(uname -m)
+	if [[ "${machine}" == "x86"* ]]; then
+		HOST_ARCH=${machine}
```
douglas-raillard-arm, line 53: I don't think we need any global mutable state here, with the problem that would ensue (someone not calling the function). The function can just `print "%s" "$arch"`  the arch value and code that needs it can call the function as `$(get_android_sdk_host_arch)`, like in basically any language.

tools/android/install_base.sh
metin-arm: It sounds pretty descriptive to me. Do you have a recommendation?

tools/android/install_base.sh
metin-arm: Yes, that was the only shellcheck violation at least when I ran it for the last time.

tools/android/install_base.sh
metin-arm: >without making the equivalent PR in LISA to keep them in sync

Because I knew we would eliminate the one in LISA at some point.

https://github.com/ARM-software/devlib/pull/678 addresses all your comments except this one. Will work on removing the duplicate later (for a separate ticket).

PR #670, opened 2024-02-05
Read target connection settings from a YAML file

metin-arm:
**Commits:**
1. [utils/android: Add debug log about connection settings](https://github.com/ARM-software/devlib/pull/670/commits/2169e61)
1. [utils/misc: Move load_struct_from_yaml() from WA to devlib](https://github.com/ARM-software/devlib/pull/670/commits/fb8887f)
1. [tests/test_target: Read target connection settings from a YAML file](https://github.com/ARM-software/devlib/pull/670/commits/6217c77)

---

### PRs https://github.com/ARM-software/devlib/pull/667 and https://github.com/ARM-software/workload-automation/pull/1248  can be merged after this one.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -294,6 +294,10 @@ class AdbConnection(ConnectionBase):
             total_transfer_timeout=total_transfer_timeout,
             transfer_poll_period=transfer_poll_period,
         )
+
+        self.logger.debug('server=%s port=%s device=%s as_root=%s',
+                          adb_server, adb_port, device, adb_as_root)
+
         self.timeout = timeout if timeout is not None else self.default_timeout
         if device is None:
             device = adb_get_device(timeout=timeout, adb_server=adb_server, adb_port=adb_port)
@@ -545,9 +549,9 @@ def adb_connect(device, timeout=None, attempts=MAX_ATTEMPTS, adb_server=None, ad
             break
         time.sleep(10)
     else:  # did not connect to the device
-        message = 'Could not connect to {}'.format(device or 'a device')
+        message = f'Could not connect to {device or "a device"} at {adb_server}:{adb_port}'
         if output:
-            message += '; got: "{}"'.format(output)
+            message += f'; got: {output}'
         raise HostError(message)
 
 
--- a/devlib/utils/misc.py
+++ b/devlib/utils/misc.py
@@ -23,6 +23,7 @@ from functools import partial, reduce, wraps
 from itertools import groupby
 from operator import itemgetter
 from weakref import WeakSet
+from ruamel.yaml import YAML
 
 import ctypes
 import logging
@@ -588,6 +589,30 @@ class LoadSyntaxError(Exception):
     def __str__(self):
         message = 'Syntax Error in {}, line {}:\n\t{}'
         return message.format(self.filepath, self.lineno, self.message)
+
+
+def load_struct_from_yaml(filepath):
+    """
+    Parses a config structure from a YAML file.
+    The structure should be composed of basic Python types.
+
+    :param filepath: Input file which contains YAML data.
+    :type filepath: str
+
+    :raises LoadSyntaxError: if there is a syntax error in YAML data.
+
+    :return: A dictionary which contains parsed YAML data
+    :rtype: Dict
+    """
+
+    try:
+        yaml = YAML(typ='safe', pure=True)
+        with open(filepath, 'r', encoding='utf-8') as file_handler:
+            return yaml.load(file_handler)
+    except yaml.YAMLError as ex:
+        message = ex.message if hasattr(ex, 'message') else ''
+        lineno = ex.problem_mark.line if hasattr(ex, 'problem_mark') else None
+        raise LoadSyntaxError(message, filepath=filepath, lineno=lineno) from ex
 
 
 RAND_MOD_NAME_LEN = 30
--- a/tests/test_target.py
+++ b/tests/test_target.py
@@ -14,35 +14,63 @@
 # limitations under the License.
 #
 
+"""Module for testing targets."""
+
 import os
 import shutil
 import tempfile
-from unittest import TestCase
+from pprint import pp
+import pytest
 
 from devlib import LocalLinuxTarget
+from devlib.utils.misc import load_struct_from_yaml
 
 
-class TestReadTreeValues(TestCase):
+def build_targets():
+    """Read targets from a YAML formatted config file"""
 
-    def test_read_multiline_values(self):
-        data = {
-            'test1': '1',
-            'test2': '2\n\n',
-            'test3': '3\n\n4\n\n',
-        }
+    config_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'target_configs.yaml')
 
-        tempdir = tempfile.mkdtemp(prefix='devlib-test-')
-        for key, value in data.items():
-            path = os.path.join(tempdir, key)
-            with open(path, 'w') as wfh:
-                wfh.write(value)
+    target_configs = load_struct_from_yaml(config_file)
+    if target_configs is None:
+        raise ValueError(f'{config_file} looks empty!')
 
-        t = LocalLinuxTarget(connection_settings={'unrooted': True})
-        raw_result = t.read_tree_values_flat(tempdir)
-        result = {os.path.basename(k): v for k, v in raw_result.items()}
+    targets = []
 
-        shutil.rmtree(tempdir)
+    if target_configs.get('LocalLinuxTarget') is not None:
+        print('> LocalLinux targets:')
+        for entry in target_configs['LocalLinuxTarget'].values():
+            pp(entry)
+            ll_target = LocalLinuxTarget(connection_settings=entry['connection_settings'])
+            targets.append(ll_target)
 
-        self.assertEqual({k: v.strip()
-                          for k, v in data.items()},
-                         result)
+    return targets
+
+
+@pytest.mark.parametrize("target", build_targets())
+def test_read_multiline_values(target):
+    """
+    Test Target.read_tree_values_flat()
+
+    :param target: Type of target per :class:`Target` based classes.
+    :type target: Target
+    """
+
+    data = {
+        'test1': '1',
+        'test2': '2\n\n',
+        'test3': '3\n\n4\n\n',
+    }
+
+    tempdir = tempfile.mkdtemp(prefix='devlib-test-')
+    for key, value in data.items():
+        path = os.path.join(tempdir, key)
+        with open(path, 'w', encoding='utf-8') as wfh:
+            wfh.write(value)
+
+    raw_result = target.read_tree_values_flat(tempdir)
+    result = {os.path.basename(k): v for k, v in raw_result.items()}
+
+    shutil.rmtree(tempdir)
+
+    assert {k: v.strip() for k, v in data.items()} == result
--- a/setup.py
+++ b/setup.py
@@ -105,6 +105,7 @@ params = dict(
         'lxml', # More robust xml parsing
         'nest_asyncio', # Allows running nested asyncio loops
         'future', # for the "past" Python package
+        'ruamel.yaml >= 0.15.72', # YAML formatted config parsing
     ],
     extras_require={
         'daq': ['daqpower>=2'],
--- a/setup.py
+++ b/setup.py
@@ -102,6 +102,7 @@ params = dict(
         'wrapt',  # Basic for construction of decorator functions
         'numpy',
         'pandas',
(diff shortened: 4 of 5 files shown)


devlib/utils/misc.py
```diff
@@ -37,6 +37,7 @@
 import types
 import warnings
 import wrapt
+import yaml
```
douglas-raillard-arm, line 40: I can't remember the specifics but there is a fundamental design issue with that library. I used ruamel.yaml in Lisa instead. That might have to do with a mutable singleton used in the lib for something central that could conflict between different users in the same process.

douglas-raillard-arm: I remember now, it's this pattern that is bonkers:

```
import yaml
yaml.add_constructor()
```

So any other dependency can decide to change the behavior of pyyaml and everyone in the same process will get the change of behavior, devlib included. That may come from the underlying C library that this wraps but this is the best recipe for unsolvable issues.

Ruamel.yaml added a way to get some sort of context object and custom constructors are only registered against it. Every user has their own instance and no-one is stepping on each other's toes.

metin-arm: Thanks for the hint. Move from `yaml` to `ruamel`.

devlib/utils/misc.py
```diff
@@ -590,6 +591,31 @@ def __str__(self):
         return message.format(self.filepath, self.lineno, self.message)
 
 
+def load_struct_from_yaml(filepath):
```
douglas-raillard-arm, line 594: For now we don't want the testing infrastructure to be a public API, so this needs a non-public name starting with a single underscore (https://peps.python.org/pep-0008/)

devlib/utils/misc.py
```diff
@@ -590,6 +591,31 @@ def __str__(self):
...
+    The structure should be composed of basic Python types.
+
+    Args:
+        filepath (str): Input file which contains YAML data.
+
+    Raises:
+        LoadSyntaxError: if there is a syntax error in YAML data.
+
+    Returns:
+        dict: which contains parsed YAML data.
```
douglas-raillard-arm, line 606: I would recommend using the autodoc style as this is the most native one with Sphinx (you can see what we use in LISA)

devlib/utils/misc.py
```diff
@@ -590,6 +591,31 @@ def __str__(self):
...
+    '''
+
+    try:
+        yaml = YAML(typ='safe', pure=True)
+        with open(filepath, 'r', encoding='utf-8') as file_handler:
+            return yaml.load(file_handler)
+    except yaml.YAMLError as ex:
+        message = ex.message if hasattr(ex, 'message') else ''
+        lineno = ex.problem_mark.line if hasattr(ex, 'problem_mark') else None
+        raise LoadSyntaxError(message, filepath=filepath, lineno=lineno) from ex
```
douglas-raillard-arm, line 616: I'm not sure what LoadSyntaxError was initially for. Currently it seems like dead code in devlib. Unless it really adds something, I'd recommend just letting the YAMLError bubble up.

tests/test_target.py
```diff
@@ -14,17 +14,30 @@
 # limitations under the License.
 #
 
+''' Module for testing targets. '''
```
douglas-raillard-arm, line 17: Docstrings should generally be trimmed, no space at the beginning or the end (unless it's needed for indentation after a newline)

tests/test_target.py
```diff
@@ -14,17 +14,30 @@
...
 from unittest import TestCase
 
 from devlib import LocalLinuxTarget
+from devlib.utils.misc import load_struct_from_yaml
 
 
 class TestReadTreeValues(TestCase):
+    ''' Class testing Target.read_tree_values_flat() '''
+
+    TARGETS_CONFIG_FILE = 'tests/target_configs.yaml'
```
douglas-raillard-arm, line 31: Does pytest guarantees the current folder when running tests ? If not this could be made relative to `__file__`'s folder for example.

tests/test_target.py
```diff
@@ -14,17 +14,30 @@
...
 
     def test_read_multiline_values(self):
+        ''' Test Target.read_tree_values_flat() '''
```
douglas-raillard-arm, line 34: `s/^ //` and `s/ $//`

tests/test_target.py
```diff
@@ -34,11 +47,13 @@ def test_read_multiline_values(self):
         tempdir = tempfile.mkdtemp(prefix='devlib-test-')
```
douglas-raillard-arm, line 47: should use context manager classes available in `tempfile`

devlib/utils/android.py
```diff
@@ -294,6 +294,10 @@ def __init__(
             total_transfer_timeout=total_transfer_timeout,
             transfer_poll_period=transfer_poll_period,
         )
+
+        self.logger.debug('AdbConnection(): server=%s port=%s device=%s as_root=%s',
```
douglas-raillard-arm, line 298: s/AdbConnection(): //

The name of the class should be used when the logger is made.

douglas-raillard-arm, line 298: There is a slight technical benefit of using this way of interpolating (i.e. avoiding interpolation unless that debug level is enabled) but it's negligible and mostly just increases the risk of mixup. You can use an f-string instead to avoid that, and even let the compiler avoid redundancy with e.g.  `{adb_server=}`.

douglas-raillard-arm (commented):
That's not a deep review on the added code, just superficial comments

tests/test_target.py
metin-arm: I run tests under `/devlib` dir. Will use `__file__`'s folder to ensure that.

devlib/utils/misc.py
metin-arm: `load_struct_from_yaml()` is not for testing purposes only.
It's just moved from WA to devlib: https://github.com/ARM-software/workload-automation/pull/1248

devlib/utils/misc.py
metin-arm: `LoadSyntaxError()` is referenced in `load_struct_from_python()`, too.
I don't know the history, but apparently it makes the syntax error location more explicit, IMHO.

tests/test_target.py
metin-arm: I don't have a change here in this PR.
https://github.com/ARM-software/devlib/pull/667/commits/79a91b7b5f9c35c9329c508de28cb34e4e9365c2 uses context manager anyway.

devlib/utils/misc.py
marcbonnici: We use this in WA to try and provide a more friendly error message to clearly indicate the location of the syntax error. 
Since this code is being moved, I'd vote for keeping the same error class.

tests/test_target.py
```diff
@@ -14,17 +14,31 @@
...
 from unittest import TestCase
 
 from devlib import LocalLinuxTarget
+from devlib.utils.misc import load_struct_from_yaml
 
 
 class TestReadTreeValues(TestCase):
+    """Class testing Target.read_tree_values_flat()"""
+
+    TARGETS_CONFIG_FILE = os.path.join(os.path.dirname(os.path.realpath(__file__)),
```
marcbonnici, line 31: If the intention is to be able to perform these tests on different device types based on the config, do we want the parsing / initialization to be performed within each test case? 
Should these actions be performed in some setup method instead that can be reused?

metin-arm: The final product is something like this: https://github.com/ARM-software/devlib/pull/667/commits/a7d0b9d80bc47c22e88a398d5a0bcb6c0c332b2b
Yes, there are different target types and there can be multiple entries for a target.
I read config file first and see which targets need to be tested. So, I could keep target classes as is.

marcbonnici: Hmm.. I see. Apologies in advanced if I'm missing the obvious but in the final product do we end up with this code just for a single `test_read_multiline_values` unittest? 
Does this mean this will need to be duplicated when we add another test or test Class?

Also does this require setting up a new connection to the target for each test? I'm wondering what the overhead/performance would be on this?

metin-arm: There is only one `test_read_multiline_values()` function. It parses the YAML file and executes the test specified in `run_test()` for each target in the YAML file. YAML file has necessary connection information and QEMU params for each target. Depending on the target type, running `run_test()` can take ~1 (LocalLinux target) or ~15 secs (for QEMU target).

Final version of `TestReadTreeValues()` looks like this:
```Python
class TestReadTreeValues(TestCase):
    def test_read_multiline_values(self):
        def run_test(target):
            ...

        target_configs = load_struct_from_yaml(self.TARGETS_CONFIG_FILE)
        ...

        if target_configs.get('AndroidTarget') is not None:
            ...
                target = AndroidTarget(
                    connection_settings=entry['connection_settings'],
                    conn_cls=lambda **kwargs: AdbConnection(adb_as_root=True, **kwargs),
                )
                run_test(target)

        if target_configs.get('LinuxTarget') is not None:
            ...
                target = LinuxTarget(connection_settings=entry['connection_settings'])
                run_test(target)

        if target_configs.get('LocalLinuxTarget') is not None:
            ...
                target = LocalLinuxTarget(connection_settings=entry['connection_settings'])
                run_test(target)

        if target_configs.get('QEMUTargetRunner') is not None:
           ...
                with QEMUTargetRunner(
                        qemu_params=qemu_params,
                        connection_settings=connection_settings,
                ) as qemu_target:
                    run_test(qemu_target.target)
```
And this is a sample YAML file:
```Yaml
AndroidTarget:
  entry-0:
    connection_settings:
      device: 'emulator-5554'
  entry-1:
    connection_settings:
      device: 'emulator-5556'

LinuxTarget:
  entry-0:
    connection_settings:
      host: 'example.com'
      username: 'username'
      password: 'password'

LocalLinuxTarget:
  entry-0:
    connection_settings:
      unrooted: True

QEMUTargetRunner:
  entry-0:
    qemu_params:
      kernel_image: '/devlib/tools/buildroot/buildroot-v2023.11.1-aarch64/output/images/Image'

  entry-1:
    qemu_params:
      kernel_image: '/devlib/tools/buildroot/buildroot-v2023.11.1-x86_64/output/images/bzImage'
      arch: 'x86_64'
      cmdline: 'console=ttyS0'
```

marcbonnici: I see, thanks for the information.
 
Although we have only have a single test case `test_read_multiline_values` at the moment, if we have this nice infrastructure to be able to run these tests and have more meaningful output, I'm thinking if we were expecting to add more test cases going forward, how would that fit into the current implementation? For example if we wanted to add a `test_read_escape_values` test case that checks certain characters are escaped correctly.

The parsing / creation of the various targets looks like it should be mostly generic and not tied to this single test case, therefore I'm wondering if there is a common location this would be better suited to be shared?

devlib/utils/misc.py
douglas-raillard-arm: Sounds good, I did not look beyond devlib codebase

tests/test_target.py
metin-arm: That totally makes sense to me.

How about the new version? If it looks OK, I'm gonna update other PRs.
Moving to `pytest` is the simplest solution I could find to prevent duplicate target config reading, passing target parameter to the test function (as well as future ones).

Thanks.

marcbonnici (approved):

tests/test_target.py
```diff
@@ -14,35 +14,63 @@
...
+        'test2': '2\n\n',
+        'test3': '3\n\n4\n\n',
+    }
 
-        tempdir = tempfile.mkdtemp(prefix='devlib-test-')
-        for key, value in data.items():
-            path = os.path.join(tempdir, key)
-            with open(path, 'w') as wfh:
-                wfh.write(value)
+    tempdir = tempfile.mkdtemp(prefix='devlib-test-')
```
douglas-raillard-arm, line 65: This should use tempfile.TemporaryDirectory, as a context manager.

metin-arm: Latest version of this code:
```Python
    with target.make_temp() as tempdir:
        ...
        raw_result = target.read_tree_values_flat(tempdir)
        result = {os.path.basename(k): v for k, v in raw_result.items()}
```
And this is `Target()::make_temp()` implementation (295f1269ed55c6a1c1b2a1c3b1e081f419931ed2):
```Python
    @contextmanager
    def make_temp(self, is_directory=True, directory='', prefix='devlib-test'):
        ...
        try:
            cmd = f'mktemp -p {directory} {prefix}-XXXXXX'
            if is_directory:
                cmd += ' -d'

            temp_obj = self.execute(cmd).strip()
            yield temp_obj
        finally:
            if temp_obj is not None:
                self.remove(temp_obj)
```

PR #673, opened 2024-02-23
Add support for testing ChromeOS targets

metin-arm:
**Commits:**
* 24a084e - target: Address pylint issues in ChromeOsTarget class
* 1bd22c4 - target: tests: Add support for testing ChromeOS targets

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -2939,7 +2939,7 @@ def _get_part_name(section):
     variant = section.get('CPU variant', '0x0')
     name = get_cpu_name(*list(map(integer, [implementer, part, variant])))
     if name is None:
-        name = '{}/{}/{}'.format(implementer, part, variant)
+        name = f'{implementer}/{part}/{variant}'
     return name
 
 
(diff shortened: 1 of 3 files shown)


devlib/target.py
```diff
@@ -3009,17 +3012,17 @@ def __init__(self,
...
+                         platform=platform,
+                         working_directory=working_directory,
+                         executables_directory=executables_directory,
+                         connect=False,
+                         modules=modules,
+                         load_default_modules=load_default_modules,
+                         shell_prompt=shell_prompt,
+                         conn_cls=SshConnection,
+                         is_container=is_container,
+                         max_async=max_async)
```
douglas-raillard-arm, line 3025: might go the extra mile and just remove the link between length of the function name and indentation:
```
foobarextremelylongandalsowewillchangeittomorrowandbreakindentation(
    iamaparameterandidontcareaboutmyfunctionnamelength=42,
    ...,
)
```

devlib/target.py
```diff
@@ -3041,7 +3044,7 @@ def __init__(self,
                                                working_directory=android_working_directory,
                                                executables_directory=android_executables_directory,
                                                connect=False,
-                                               modules=[], # Only use modules with linux target
+                                               modules=[],  # Only use modules with linux target
```
douglas-raillard-arm, line 3047: That's unrelated to this PR so maybe don't fix it here, but that's a nice antipattern in there: mutable default values. The default value is created when the function is created. Then that specific instance will be used for all function calls where no value is specified:

```
def foo(x, mylist=[]):
    mylist.append(x)
    return mylist

foo(1)
foo(2)
mylist = foo(3)

assert mylist == [1, 2, 3]
```
So the moral here is: never use mutable values as defaults. `None` is not mutable (or it is mutable but has no state to actually mutate, so both views are actually valid), so it makes a nice default. If an empty iterable default was _really_ wanted, `tuple()` would do it.

metin-arm: Removed `modules` param since it defaults to `None` already.

PR #674, opened 2024-03-19
Add option not to validate PMU counters.

ola1olssonarm:
The validation call can take a long for targets where PLL:s have been clocked down, such as FPGAs.


marcbonnici (approved):

ola1olssonarm:
Thank you Marc. Just tell me if there is anything I should do now.

PR #683, opened 2024-04-26
Replace nest_asyncio with greenlets

### LINKED ISSUE #682: Remove nest_asyncio dependency
The author of nest_asyncio has passed away and the project is now abandoned: https://github.com/erdewit/nest_asyncio

Unfortunately, nest_asyncio in its current state does not support alternative event loops like uvloop, used by a number of project, e.g. the gunicorn WSGI server. According to this issue, the Python stdlib [still does not have any plan on supporting nested event loop](https://github.com/python/cpython/issues/66435), but someone seems to have found a way to run an async function in synchronous world without an event loop at all: https://stackoverflow.com/questions/78170266/python-decorator-for-async-and-sync-function-without-code-duplication/78170267#78170267

I'll have a look tomorrow at that option and try to make a PR to remove the dependency

douglas-raillard-arm:
Considering https://github.com/ARM-software/devlib/issues/682, this PR updates devlib to use `greenlets` in cases where `nest_asyncio` cannot be used (e.g. using `uvloop` event loop).

This alternative path works by using `greenlet` to provide a way for nested coroutines separated from the top-level coroutine by blocking calls to yield their action. 

TODO:

- [x] Handle exceptions raised in `__await__`, so that they are propagated to the correct coroutine rather than the top-level one
- [x] Write unit test to check everything works on various event loops (e.g. uvloop)
- [x] Add some rational in asyn.py to explain the moving pieces

Fixes #682

### DIFF
--- a/devlib/module/cgroups.py
+++ b/devlib/module/cgroups.py
@@ -19,13 +19,12 @@ from collections import namedtuple
 from shlex import quote
 import itertools
 import warnings
-import asyncio
 
 from devlib.module import Module
 from devlib.exception import TargetStableError
 from devlib.utils.misc import list_to_ranges, isiterable
 from devlib.utils.types import boolean
-from devlib.utils.asyn import asyncf
+from devlib.utils.asyn import asyncf, run
 
 
 class Controller(object):
@@ -410,7 +409,7 @@ class CgroupsModule(Module):
                              controller.mount_point)
             self.controllers[ss.name] = controller
 
-        asyncio.run(
+        run(
             target.async_manager.map_concurrently(
                 register_controller,
                 subsys,
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -103,7 +103,8 @@ def call_conn(f):
 
     @functools.wraps(f)
     def wrapper(self, *args, **kwargs):
-        reentered = self.conn.is_in_use
+        conn = self.conn
+        reentered = conn.is_in_use
         disconnect = False
         try:
             # If the connection was already in use we need to use a different
@@ -113,8 +114,9 @@ def call_conn(f):
             if reentered:
                 # Shallow copy so we can use another connection instance
                 _self = copy.copy(self)
-                _self.conn = _self.get_connection()
-                assert self.conn is not _self.conn
+                new_conn = _self.get_connection()
+                assert conn is not new_conn
+                _self.conn = new_conn
                 disconnect = True
             else:
                 _self = self
@@ -122,6 +124,13 @@ def call_conn(f):
         finally:
             if disconnect:
                 _self.disconnect()
+            elif not reentered:
+                # Return the connection to the pool, so if we end up exiting
+                # the thread the connection can then be reused by another
+                # thread.
+                del self.conn
+                with self._lock:
+                    self._unused_conns.add(conn)
 
     return wrapper
 
@@ -284,7 +293,8 @@ class Target(object):
     @tls_property
     def _conn(self):
         try:
-            return self._unused_conns.pop()
+            with self._lock:
+                return self._unused_conns.pop()
         except KeyError:
             return self.get_connection()
 
@@ -311,6 +321,8 @@ class Target(object):
                  is_container=False,
                  max_async=50,
                  ):
+
+        self._lock = threading.RLock()
         self._async_pool = None
         self._async_pool_size = None
         self._unused_conns = set()
@@ -435,6 +447,7 @@ class Target(object):
         ignored.update((
             '_async_pool',
             '_unused_conns',
+            '_lock',
         ))
         return {
             k: v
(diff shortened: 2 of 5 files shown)


douglas-raillard-arm:
PR updated with:
1. implementation directly based on greenlet rather than greenback. This avoids some ctypes-based hacks that are not necessary. That implementation is compatible with any asyncio event loop implementation.
2. Still use `nest_asyncio` when possible, as the greenlet version might execute the coroutine on a separate thread in some circumstances, whereas `nest_asyncio` never needs to.

douglas-raillard-arm:
PR updated with exception handling in coroutine and exception injection with `.throw()`. 
The state of this PR will stay as [Draft] until I add some unit tests to ensure correct operations (and fix all the items in the TODO list in this PR cover letter) as discussed with Vincent Coubard.

douglas-raillard-arm:
PR updated with:
* unit tests added for `devlib.utils.asyn.run()`, simulating various event loop setups including non-stdlib event loops (`uvloop`). A fairly comprehensive set of nested run() calls is checked, as well as exception and value injection (`coro.send()` and `coro.throw()`)
* `nest_asyncio` properly removed, as it relies on `asyncio.get_event_loop()` will not create an event loop automatically if one is not setup when called from a non-main thread. `asyncio.get_event_loop()` is deprecated and cannot be relied upon to create an event loop, so `nest_asyncio` usage is basically broken. Since we want devlib to be importable in the non-main thread (no reason for it to explode), this is not acceptable and we can just fully switch to our new implementation

douglas-raillard-arm:
PR updated with extra comments. I now consider it ready so I'll remove the `[Draft]` status

douglas-raillard-arm:
Found an issue, please do not merge as-is

douglas-raillard-arm:
PR updated with:
* extra unit tests that reproduced the issue
* `_AsyncPolymorphicCM` now spins an event loop if necessary in `__enter__` and closes it in `__exit__`, so that both the `__aenter__()` and `__aexit__()` coroutines are executed on the same event loop.

douglas-raillard-arm:
@marcbonnici, Vincent Coubard, Branislav Rankov, this should be ready for the last review/testing round. I consider it to be ready.

douglas-raillard-arm:
PR rebased. As of today we started dogfooding that PR in LISA's vendored devlib tree, so it should get some more real-world exposure in the coming weeks.

douglas-raillard-arm:
PR updated with extra tests to check the blocking API works when invoked from `asyncio.to_thread()`. Everything does work.

douglas-raillard-arm:
PR updated with a fix to use `run()` on the return values of `anext()` for async generators, with the matching unit test.

douglas-raillard-arm:
PR rebased

douglas-raillard-arm:
Update with a check to only call  loop.shutdown_default_executor() if it exists, since it was added in Python 3.9

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.shutdown_default_executor

douglas-raillard-arm:
An issue was found, I'm currently working on fixing it, please do not merge until it's fixed.

douglas-raillard-arm:
PR updated with:
* Some refactoring of private functions involved in `run()` implementation.
* Ensure that we really do use the same loop for both `__aenter__` and `__aexit__` for async context managers.
* Ensure that each `asyncio.Task` gets its own top-level instrumented coroutine. This is critical as a Task would not be able to yield on behalf of a coroutine in another task. This would deadlock, since the event loop would be polling task B while a nested coroutine inside task B would be trying to yield through task A. Task A would be ignored by the event loop until task B yields, leading to a deadlock.

douglas-raillard-arm:
One thing I realized and might want to change is that if an event loop is already running (e.g. in a jupyterlab notebook), we will dispatch the coroutine on a loop setup in a separate thread. It's all good except we have a single such thread.

This means that code making parallel devlib invocation of devlib in threads with a pre-setup event loop will end up being serialized. It shouldn't be very hard to change, I'll see if I can do that tomorrow

douglas-raillard-arm:
PR updated with:
* significant simplification in the way the different cases are handled. There is now a `_CoroRunner` subclass for each case.
* simplification of how async context managers are handled, thanks to `_CoroRunner`. All we now need is to get a runner that will be usable to execute multiple coroutines on the same event loop, even if that event loop is sitting in a separate thread.
* Added `contextvars` test and fixed the code. `contextvars` are now propagated down through `devlib.utils.asyn.run()` and any update done inside the coroutine will be propagated back in the caller of `run()`, so it is completely transparent.
* The threaded coroutine runner used as last resort when an event loop is already running in the current thread can now use an arbitrary number of threads. This prevents serialization of otherwise parallel code and ensures no deadlock can occur.

Considered it is a substantial change, I'd be more comfortable dogfooding it in LISA for a little while before we consider merging it.

douglas-raillard-arm:
1.5 month later there has been no reported issue, so I think it's good to go

marcbonnici (approved):

Issue #684, opened 2024-04-29
utils.ssh.SshConnectionBase issues caused by garbage collection


ep1cman:
I am hitting a rather odd issue where at the end of my scripts I am seeing a flood of the following messages:
```
Exception ignored in: <function ConnectionBase.__del__ at 0x1062e1a20>
Traceback (most recent call last):
  File "/Users/sebgos01/Library/Caches/pypoetry/virtualenvs/appalib-JvmmYHe4-py3.10/lib/python3.10/site-packages/devlib/utils/misc.py", line 961, in wrapper
  File "/Users/sebgos01/Library/Caches/pypoetry/virtualenvs/appalib-JvmmYHe4-py3.10/lib/python3.10/site-packages/devlib/connection.py", line 121, in __del__
  File "/Users/sebgos01/Library/Caches/pypoetry/virtualenvs/appalib-JvmmYHe4-py3.10/lib/python3.10/site-packages/devlib/utils/misc.py", line 961, in wrapper
  File "/Users/sebgos01/Library/Caches/pypoetry/virtualenvs/appalib-JvmmYHe4-py3.10/lib/python3.10/site-packages/devlib/connection.py", line 111, in close
  File "/Users/sebgos01/Library/Caches/pypoetry/virtualenvs/appalib-JvmmYHe4-py3.10/lib/python3.10/site-packages/devlib/utils/misc.py", line 961, in wrapper
  File "/Users/sebgos01/Library/Caches/pypoetry/virtualenvs/appalib-JvmmYHe4-py3.10/lib/python3.10/site-packages/devlib/utils/ssh.py", line 725, in _close
AttributeError: 'NoneType' object has no attribute 'debug'
```

Both `logger` and `_handle_paramiko_exceptions` no longer seem to exist.

What I think is happening is that during interpreter shut down these have already been garbage collected, and when it comes time to garbage collect `ConnectionBase`, `__del__` is called but ends up being unable to run.

marcbonnici:
@douglas-raillard-arm Have you seen this behavior on your setup?

douglas-raillard-arm:
I haven't seen it at that specific location but in general yes, that makes sense. That's partly why relying on __del__ is a bad idea and context managers should be used. 

What we should do is :
1. Provide the context manager 
2. Maybe silently ignore all exceptions in del. It's not ideal as it might hide other bugs, but not much other choice. We could still log something at debug level

douglas-raillard-arm:
[...]

PR #685, opened 2024-05-07
Fix __del__ excep

douglas-raillard-arm:
Fixes https://github.com/ARM-software/devlib/issues/684

This PR is untested for now.


douglas-raillard-arm:
PR updated to also disconnect upon process exit, before the modules are torn down. This should prevent getting exceptions logged in `__del__` in common cases, for anything that is cleaned-up in `Target.disconnect()` (which includes all current connection objects)

marcbonnici (approved):

PR #687, opened 2024-05-22
Validate cgroups_run_into has taken effect

ep1cman:
On some systems this seems to have no effect, leaving the executed shell in the root cgroup. Before, this function would still execute and the end user would think the desired process was run in the cgroup when infact it had not.


marcbonnici (approved):
Thanks for adding this, as Douglas mentioned it might be worth migrating to the updated module if you can, however this LGTM to help other users of this implementation.

PR #701, opened 2024-09-23
ftrace: Preserve kprobe events during trace-cmd reset

metin-arm:
FtraceCollector.reset() executes 'trace-cmd reset ..' command which
clears all kprobes. This breaks tracing existing kprobe events (if any).
Thus, save kprobe events before trace-cmd reset and restore them after
the reset operation.

For the context, I want to trace an ordinary function in kernel (e.g.,
"echo 'p do_sys_open' > /sys/kernel/tracing/kprobe_events"). However,
FtraceCollector.reset() destroys kprobes, too. Preserving existing
kprobes allows me to use FtraceCollector class as is.

### DIFF
--- a/devlib/collector/ftrace.py
+++ b/devlib/collector/ftrace.py
@@ -107,7 +107,8 @@ class FtraceCollector(CollectorBase):
         self.function_profile_file    = self.target.path.join(self.tracing_path, 'function_profile_enabled')
         self.marker_file              = self.target.path.join(self.tracing_path, 'trace_marker')
         self.ftrace_filter_file       = self.target.path.join(self.tracing_path, 'set_ftrace_filter')
-        self.available_tracers_file  = self.target.path.join(self.tracing_path, 'available_tracers')
+        self.available_tracers_file   = self.target.path.join(self.tracing_path, 'available_tracers')
+        self.kprobe_events            = self.target.path.join(self.tracing_path, 'kprobe_events')
 
         self.host_binary = which('trace-cmd')
         self.kernelshark = which('kernelshark')
@@ -240,6 +241,9 @@ class FtraceCollector(CollectorBase):
         return self.target.read_value(self.available_functions_file).splitlines()
 
     def reset(self):
+        # Save kprobe events
+        kprobe_events = self.target.read_value(self.kprobe_events)
+
         self.target.execute('{} reset -B devlib'.format(self.target_binary),
                             as_root=True, timeout=TIMEOUT)
 
@@ -256,6 +260,10 @@ class FtraceCollector(CollectorBase):
 
         if self.functions:
             self.target.write_value(self.function_profile_file, 0, verify=False)
+
+        # Restore kprobe events
+        self.target.write_value(self.kprobe_events, kprobe_events)
+
         self._reset_needed = False
 
     @asyncf


devlib/collector/ftrace.py
```diff
@@ -97,7 +97,7 @@ def __init__(self, target,
         self.function_string = None
         self.trace_clock = trace_clock
         self.saved_cmdlines_nr = saved_cmdlines_nr
-        self._reset_needed = True
```
douglas-raillard-arm, line 100: This is part of the internal state of the object. Binding it to a tunable will break some flows where tracing is interrupted and restarted.

devlib/collector/ftrace.py
```diff
@@ -61,7 +61,6 @@ def __init__(self, target,
                  trace_children_functions=False,
                  buffer_size=None,
                  top_buffer_size=None,
-                 buffer_size_step=1000,
```
douglas-raillard-arm, line 64: This can't be removed without a warning and a deprecation period as this is part of the public API.

douglas-raillard-arm (changes requested):
I don't think exposing `reset_needed` is a good idea. There are reasons why ftrace needs to be reset before tracing as it could have been left in an arbitrary state by a previous user. Currently, all we have to define is the behavior of each parameter when applied on a reset state. If we expose that, we will have to provide clear documentation for all the other options and how they interact with a non-reset state. This is extra tricky as trace-cmd itself will set some things but not some others unless asked for. 

When it comes to the actual use case, this really should be supported in trace-cmd itself (i.e. add an option not to reset kprobe events, or even possibly change the default TBH unless those kprobes events are always-on have some other weird behavior). In the interim, we can emulate that behavior by saving and restoring the kprobe events on each side of the reset.

devlib/collector/ftrace.py
metin-arm: OK. Dropped that much for now.

douglas-raillard-arm: Yeah some cleanup is overdue, I think this is sort of cleanup is on the roadmap of the next maintainers

devlib/collector/ftrace.py
metin-arm: Implemented save/restore logic for kprobes as you recommended.
Thank you!

metin-arm:
Filed a [question](https://bugzilla.kernel.org/show_bug.cgi?id=219302) to check if we can introduce a new flag in trace-cmd to exclude resetting existing kprobe events.

devlib/collector/ftrace.py
```diff
@@ -240,6 +241,9 @@ def available_functions(self):
         return self.target.read_value(self.available_functions_file).splitlines()
 
     def reset(self):
+        # Save kprobe events
+        kprobe_events = self.target.read_value(self.kprobe_events)
```
douglas-raillard-arm, line 245: Out of curiosity, do you have a sample of what reading the file provides with a probe set ? The kernel doc does not seem to describe that, it only talks about the file behavior when written to (I may have missed it).

douglas-raillard-arm (commented):
LGTM

devlib/collector/ftrace.py
metin-arm: Sorry if I misunderstood, but is this what you asked for?
```Bash
$ echo 'p do_sys_open' > /sys/kernel/tracing/kprobe_events                                                                                                                                                                                                                       
$ echo 'r do_sys_open' >> /sys/kernel/tracing/kprobe_events                                                                                                                                                                                                                      
$ cat /sys/kernel/tracing/kprobe_events                                                                                                                                                                                                                                          
p:kprobes/p_do_sys_open_0 do_sys_open
r18:kprobes/r_do_sys_open_0 do_sys_open
```

douglas-raillard-arm: Makes sense, thanks

marcbonnici (approved):

PR #707, opened 2024-11-20
Fix ssh folder pull

douglas-raillard-arm:
Fix SSH pull for folders over SFTP and asyncify the generic part of the push/pull implementation so it can run concurrently for multiple files when globbing is used.


douglas-raillard-arm:

marcbonnici (approved):

PR #709, opened 2025-01-30
collector/dmesg: Fix dmesg variant detection

douglas-raillard-arm:
Check for all the CLI options we are going to use when deciding whether to use the system's dmesg or the one we ship via busybox.


douglas-raillard-arm:

marcbonnici (approved):

PR #710, opened 2025-01-30
target: Remove duplicated disconnection logic

douglas-raillard-arm:
The logic in Target.disconnect() appears to have been duplicated by error. While _probably_ harmless, this is at least confusing, and since this happens outside of the lock, this may actually be a real problem.


douglas-raillard-arm:

marcbonnici (approved):

PR #712, opened 2025-02-03
Fix cpufreq event

douglas-raillard-arm:


douglas-raillard-arm:

devlib/collector/ftrace.py
```diff
@@ -195,7 +194,11 @@ def event_is_in_list(event, events):
             elif self.tracer == 'function_graph':
                 self.function_string = _build_graph_functions(selected_functions, trace_children_functions)
 
-        self.event_string = _build_trace_events(selected_events)
+        self._selected_events = selected_events
+
+    @property
+    def event_string(self):
```
arjpur01, line 200: why change event_string to property? what was the issue when it is a variable?

arjpur01 (approved):

devlib/collector/ftrace.py
douglas-raillard-arm: It's derived from `self._selected_events`, leaving it as an attribute both clutters the state of the object with redundant information and also opens the door for mismatching content. Having it as a property makes the broken state impossible to represent.

arjpur01 (approved):

marcbonnici (approved):

PR #713, opened 2025-02-04
utils/android: Restore adb root state when disconnecting

douglas-raillard-arm:
The current behavior is to issue "adb unroot" if the device needed to be rooted upon connection. This breaks use of nested Targets, which LISA requires as some target interaction needs to happen in a subprocess.

Fix that by restoring the same adb root state that there was when creating the connection, rather than blindly unrooting the device upon disconnection.


douglas-raillard-arm:

marcbonnici (approved):

PR #720, opened 2025-03-19
Fix make_temp: revert fallback to working_directory

simant01:
This pull request fixes an issue introduced by a recent commit that changed the temporary file creation behavior in make_temp

Previously, make_temp used self.working_directory and a default prefix of 'devlib-test', which provided the necessary permissions for temporary file cleanup. The change to using self.tmp_directory caused permission errors during cleanup and consequently breakages. 

This PR reverts the change by:
- Setting the default prefix to 'devlib-test'
- Reverting the fallback directory from self.tmp_directory back to self.working_directory

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -1215,7 +1215,7 @@ fi
                 raise
 
     @asyn.asynccontextmanager
-    async def make_temp(self, is_directory=True, directory=None, prefix=None):
+    async def make_temp(self, is_directory=True, directory=None, prefix='devlib-test'):
         """
         Creates temporary file/folder on target and deletes it once it's done.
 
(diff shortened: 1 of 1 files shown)


douglas-raillard-arm:
Instead of reverting that change, as `tmp_directory`'s only purpose is to be the root of temporary locations, could you give a go at fixing the logic that computes a sane default for it ?
In particular:
https://github.com/ARM-software/devlib/blob/b9f2384a8b0bb81413a99bfbbb36949807991a08/devlib/target.py#L2633

douglas-raillard-arm:
Also if you can share details on your setup, the specific paths involved and the specific error you encounter that would be useful. AFAICT `tmp_directory` can only be derived from the following sources:
1. User-given. I assume it's not what is happening
2. From `mktemp -d` call. If that returns a folder that is not writeable (which I assume is the problem ?) your platform is pretty broken and you may need to set TMPDIR manually in some shell configuration file.
3. `/data/local/tmp`, which AFAIR is an ok location on android (or used to be)
4. From `working_directory`, which I assume is not the case considering your report.

PR #722, opened 2025-05-20
utils/android: Fix AdbConnection.execute(check_exit_code=False) default

douglas-raillard-arm:
Switch to have check_exit_code=True just like any other connection.

The current behavior will not raise any exception if the command returns a non-zero exit code. This leads to failed attempt at parsing the output, which is now an error message rather than the expected data. Worse, the caller may never realize the command failed. This is especially bad as that behavior will only manifest itself when things go wrong, which is not the majority of the time, leading to code that seems to work ok, but does not handle failure properly (like a shell script). Lastly, since this is at odds with all the other connection types, generic code will typically assume check_exit_code=True by default and end up being buggy when used in conjunction of the AdbConnection.

### DIFF
--- a/devlib/utils/android.py
+++ b/devlib/utils/android.py
@@ -359,7 +359,7 @@ class AdbConnection(ConnectionBase):
                 popen.communicate()
 
     # pylint: disable=unused-argument
-    def execute(self, command, timeout=None, check_exit_code=False,
+    def execute(self, command, timeout=None, check_exit_code=True,
                 as_root=False, strip_colors=True, will_succeed=False):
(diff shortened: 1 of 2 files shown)


marcbonnici (commented):
Would it be possible to update the documentation [1] to reflect this change? 
[1] https://github.com/ARM-software/devlib/blob/f71b6a7d96fda77112d1a29c01fcb1230e7eb56c/doc/connection.rst?plain=1#L44

douglas-raillard-arm:
@marcbonnici PR updated with doc fix, thanks for catching that

PR #729, opened 2025-08-15
Add cross-platform LocalTarget with macOS and Linux support

### LINKED ISSUE #728: Error when running on macOS: /proc/cpuinfo not found
Hi,

I tried running my macOS machine and encountered the following error during the tests:

`cat: /proc/cpuinfo: No such file or directory`

This happens because `/proc/cpuinfo` is not available on macOS.
Just reporting in case this behavior is unexpected or could be handled more gracefully.

denini08:
This PR introduces the new `LocalTarget` class, which unifies local target handling across macOS and Linux. Fix #728.

Key changes include:

- Automatic platform detection (mac, linux)
- Native command usage on macOS (e.g., sysctl, screencapture) and Linux (e.g., /proc/cpuinfo, scrot)
- Fallback to busybox commands for unsupported platforms
- Updated tests and configuration to use LocalTarget instead of platform-specific classes
- Maintained backward compatibility (LocalMacTarget remains as alias)

Benefits:

- Single class works on both macOS and Linux without manual selection
- Platform-specific logic centralized for easier maintenance
- Simplifies test configuration and improves code clarity

Let me know your thoughts. Happy to help.

### DIFF
--- a/devlib/target.py
+++ b/devlib/target.py
@@ -3138,16 +3138,13 @@ class LocalLinuxTarget(LinuxTarget):
             self.working_directory = '/tmp/devlib-target'
 
 
-class LocalTarget(Target):
+class LocalTarget(LocalLinuxTarget):
     """
     Auto-detecting local target that works on both Linux and macOS.
     
-    This class automatically detects the underlying operating system and provides
-    appropriate implementations for each platform, using native commands instead
-    of busybox when needed.
+    This class inherits from LocalLinuxTarget to reuse Linux implementations
+    and overrides only the methods that need macOS-specific behavior.
     """
-    
-    path = posixpath
 
     def __init__(self,
                  connection_settings=None,
@@ -3204,9 +3201,9 @@ class LocalTarget(Target):
     @memoized
     def abi(self):
         """Get the system architecture"""
-        try:
-            arch = self.execute('uname -m').strip()
-            if self._is_macos():
+        if self._is_macos():
+            try:
+                arch = self.execute('uname -m').strip()
                 # Map macOS architecture names to common names
                 arch_map = {
                     'arm64': 'arm64',
@@ -3214,11 +3211,11 @@ class LocalTarget(Target):
                     'i386': 'x86',
                 }
                 return arch_map.get(arch, arch)
-            else:
-                # Linux architecture mapping
-                return ABI_MAP.get(arch, arch)
-        except TargetStableError:
-            return 'unknown'
+            except TargetStableError:
+                return 'unknown'
+        else:
+            # Use parent implementation for Linux systems
+            return super(LocalTarget, self).abi
 
     @property
     @memoized
@@ -3237,19 +3234,9 @@ class LocalTarget(Target):
                 }
             except TargetStableError:
                 return {}
-        elif self._is_linux():
-            try:
-                os_version = {}
-                if self.file_exists('/etc/os-release'):
-                    output = self.execute('cat /etc/os-release')
-                    for line in output.split('\n'):
-                        if '=' in line and not line.startswith('#'):
-                            key, value = line.split('=', 1)
-                            os_version[key] = value.strip('"')
-                return os_version
-            except TargetStableError:
-                return {}
-        return {}
+        else:
+            # Use parent implementation for Linux and other systems
+            return super(LocalTarget, self).os_version
 
     @property
     @memoized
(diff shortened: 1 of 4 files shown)


marcbonnici (commented):
Hi,

Thank you very much for the pull request. I had some initial high level design questions with regards to the new `Target`.

Thanks

devlib/target.py
```diff
@@ -3138,6 +3138,447 @@ def _resolve_paths(self):
             self.working_directory = '/tmp/devlib-target'
 
 
+class LocalTarget(Target):
```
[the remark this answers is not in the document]
denini08: I made a commit now trying to resolve all your feedbacks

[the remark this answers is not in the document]
denini08: also, feel free to make commits if you find something that can be fixed (I don't have much knowledge of the code base so I might have made a mistake, I'm just trying to help)

douglas-raillard-arm:
This sort of thing is a common failure mode for OO approaches:
>        if self._is_macos():

AFAICT this is implementing that aspect:

> Platform-specific logic centralized for easier maintenance

This is _really_ not in the benefits category. If all target classes were merged into a single class like this one with a bunch of `if self._is_foobar()`, the code would be unmaintainable: each method would be several times larger than it is today and it would be hard to assess when adding a new target whether a spot was missed or not. On top of that, there likely would be some broken code sharing between target that would lead to problems, e.g. mac code falling back on Linux code because it looks the same, and Linux code being updated one day to fix something in the linux world, accidentally breaking the mac world.

This could be replaced by having a class that instantiate another class chosen dynamically and redirects all method attribute lookup to that internal one using e.g. `__getattribute__`. 

Once this is done, a `LocalMacTarget` can be created, mirroring the existing `LocalLinuxTarget`.

This way we keep the fundamental design property of devlib that each target class is specialized in a single OS, with the exception of a "dispatch" target that just redirects to an actual implementation.

Note that this design is what we implemented in LISA, where there is a single user-facing `Target` class that instantiate one of the devlib classes under the hood. Similarly, devlib currently instantiate connection classes dynamically, so that we don't have a class for each `OS x connection`.

If there is any code that makes sense to actually share, it can be split out in separate private standalone functions called by both Linux and Mac targets.

devlib/target.py
```diff
@@ -3138,6 +3138,416 @@ def _resolve_paths(self):
...
+                return {
+                    'name': name,
+                    'version': version,
+                    'build': build,
+                }
+            except TargetStableError:
+                return {}
+        else:
+            # Use parent implementation for Linux and other systems
+            return super(LocalTarget, self).os_version
```
douglas-raillard-arm, line 3239: `super(LocalTarget, self)` is just `super()` in this context. Arguments are only needed when:
1. The parser does not provide the interpreter enough context (e.g. in a nested function).
2. When trying to achieve something unusual (not following the normal MRO)
3. When trying to be compatible with Python 2 (official eol in 2020, devlib has dropped support for it a long time ago)

devlib/target.py
```diff
@@ -3138,6 +3138,416 @@ def _resolve_paths(self):
...
+                f'model name\t: {brand_string}',
+            ]
+            if cpu_family:
+                section.append(f'cpu family\t: {cpu_family}')
+            if cpu_model:
+                section.append(f'model\t\t: {cpu_model}')
+            
+            cpuinfo_text.append('\n'.join(section))
+        
+        return Cpuinfo('\n\n'.join(cpuinfo_text))
```
douglas-raillard-arm, line 3309: A new class that allows passing the various components explicitly is needed here. This would be a base class of the current `Cpuinfo`, and the latter could simply parse the text format in its `__init__`. That base class would be used directly here, or we can make a subclass that does essentially nothing to keep the base private.

The only issue with that scheme is that `isinstance(x, Cpuinfo)` would now be inaccurate, but I don't think anyone relies on that as there is only a single Cpuinfo class currently.

devlib/target.py
```diff
@@ -3138,6 +3138,416 @@ def _resolve_paths(self):
...
+        """Write value to file using platform-appropriate command"""
+        if self._is_macos():
+            # macOS has native printf
+            self.async_manager.track_access(
+                asyn.PathAccess(namespace='target', path=path, mode='w')
+            )
+            value = str(value)
+
+            if verify:
+                # Use the same verification logic as the parent class but with native commands
```
douglas-raillard-arm, line 3356: > but with native commands

What is the advantage for that ? The main reason we use busybox everywhere in devlib is so that we know what behaviour to expect. While a mac target has to deal with only one OS rather than multiple for linux-based code, there may still be changes from one version to another as we have painfully learnt with android over the years.

douglas-raillard-arm (changes requested):
Some generic comments that would likely stay relevant regardless of future refactoring. It's more things to keep in mind while reworking the code than something to address on the current PR version.

